Issue #16, opened 2018-12-13
Set up Swagger


smadbe:
Integrate Swagger to the API so that an API doc can be run (document it rather than automate the doc generation part)

smadbe:
Fixed by #149

PR #19, opened 2018-12-18
WIP: generating Swagger docs

slomek:
Opening to have an initial discussion

Fixes https://github.com/France-ioi/AlgoreaBackend/issues/16.


docs/docs.go
```diff
@@ -0,0 +1,107 @@
...
+// This file was generated by swaggo/swag at
+// 2018-12-17 22:19:48.312668 +0100 CET m=+0.051467374
+
+package docs
+
+import (
+	"bytes"
+
+	"github.com/alecthomas/template"
+	"github.com/swaggo/swag"
```
smadbe: isn't go-swagger more appropriate? seems to be more popular

docs/swagger/swagger.json
```diff
@@ -0,0 +1,63 @@
+{
```
smadbe: if generated, should probably not be committed !?

docs/swagger/swagger.yaml
```diff
@@ -0,0 +1,43 @@
...
+    name: API Support
+    url: http://www.swagger.io/support
+  description: This is a sample server Petstore server.
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+  termsOfService: http://swagger.io/terms/
+  title: Swagger Example API
+  version: "1.0"
+paths:
```
smadbe: currently repeating stuff from the service file. not DRY.

docs/docs.go
slomek: I'll take a look on go-swagger as well, this is the first thing I found. Quick look suggests that they are very similar, but I'll check how do they differ.

docs/swagger/swagger.json
slomek: Well, depends on how you want to look at it, eg. if you want to require everyone to have the whole toolset to run the server - yes. If you want to limit the dev-dependencies to minimum you might want to have this committed. Eg. if you get some testers/FE people to run the server locally, this might just be in the repo. I usually commit such stuff even if it's generated, but I'm open to deleting it as well 😉

docs/swagger/swagger.yaml
slomek: This file is auto-generated.

smadbe: So should it really be committed? A risk to have inconsistent files if not regenerated when quickly changing a service.

smadbe:
What the status of that? 
What about "swaggo/swag" vs "go-swagger"?

docs/swagger/swagger.json
ribice: I think it should be omitted and generated by CI/CD.

Issue #17, opened 2018-12-13
Hierarchy check on get-items-list


smadbe:
see specs

smadbe:
Fixed by #31

PR #21, opened 2018-12-20
validate the hierarchy

vyloy:
validate the hierarchy of items

Fixes https://github.com/France-ioi/AlgoreaBackend/issues/17.

### DIFF
--- a/app/api/items/get-list.go
+++ b/app/api/items/get-list.go
@@ -29,8 +29,11 @@ func (srv *Service) getList(w http.ResponseWriter, r *http.Request) s.APIError {
 		return s.ErrForbidden(errors.New("Insufficient access on given item ids"))
 	}
 
-	// Todo: validate the hierarchy
-	// srv.Store.Items.IsValidHierarchy(...)
+	if valid, err := srv.Store.Items().IsValidHierarchy(ids); err != nil {
+		return s.ErrUnexpected(err)
+	} else if !valid {
+		return s.ErrForbidden(errors.New("Insufficient access on given item ids"))
+	}
 
 	// Build response
 	// Fetch the requested items
--- a/app/api/items/get-list_test.go
+++ b/app/api/items/get-list_test.go
@@ -1,0 +1,108 @@
+package items
+
+import (
+	"encoding/json"
+	"github.com/France-ioi/AlgoreaBackend/app/config"
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+	"github.com/jinzhu/gorm"
+	assert_lib "github.com/stretchr/testify/assert"
+	"log"
+	"os"
+	"os/exec"
+	"testing"
+)
+
+func TestIsValidHierarchy(t *testing.T) {
+	assert := assert_lib.New(t)
+
+	config.Path = "testdata/default.yaml"
+	conf, err := config.Load()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	db, err := database.Open(conf.Database)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	db.SetLogger(gorm.Logger{log.New(os.Stdout, "\r\n", 0)})
+	db.LogMode(true)
+
+	base := service.Base{Store: database.NewDataStore(db), Config: conf}
+	srv := &Service{Base: base}
+
+	testCases := []struct {
+		desc   string
+		items  []string
+		args   []int64
+		expect bool
+	}{
+		{
+			desc: "normal",
+			items: []string{
+				`{"id":21, "type":"Root", "strings":[{"language_id":1, "title":"21"}], "parents": [{"id":21, "order":0}]}`,
+				`{"id":22, "type":"Category", "strings":[{"language_id":1, "title":"22"}], "parents": [{"id":21, "order":0}]}`,
+				`{"id":23, "type":"Chapter", "strings":[{"language_id":1, "title":"23"}], "parents": [{"id":22, "order":1}]}`,
+			},
+			args:   []int64{21, 22, 23},
+			expect: true,
+		},
+		{
+			desc: "one item missing",
+			items: []string{
+				`{"id":21, "type":"Root", "strings":[{"language_id":1, "title":"21"}], "parents": [{"id":21, "order":0}]}`,
+				`{"id":22, "type":"Category", "strings":[{"language_id":1, "title":"22"}], "parents": [{"id":21, "order":0}]}`,
+				// `{"id":23, "type":"Chapter", "strings":[{"language_id":1, "title":"23"}], "parents": [{"id":22, "order":1}]}`,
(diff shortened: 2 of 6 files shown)


app/api/items/get-list_test.go
```diff
@@ -0,0 +1,108 @@
...
+				t.Fatal(err)
+			}
+
+			assert.Equal(tCase.expect, isValidHierarchy)
+		})
+	}
+}
+
+func restoreDB(t *testing.T) {
+	cmd := exec.Command("/bin/bash", "testdata/restoredb.sh")
```
smadbe: I don't like the idea of calling bash script from the go file. (especially as there are lib for that in the project)

app/api/items/testdata/default.yaml
```diff
@@ -0,0 +1,17 @@
+server:
```
smadbe: Wow, we are not gonna replicate the default config file in every directory that we test, this is not gonna work, right?

app/database/itemStore.go
```diff
@@ -38,9 +38,31 @@ func (s *ItemStore) Insert(data *Item) error {
...
 
+// All creates a composable query without filtering
+func (s *ItemStore) All() *DB {
+	return &DB{s.db.Table(s.tableName())}
+}
+
 // IsValidHierarchy gets an ordered set of item ids and returns whether they forms a valid item hierarchy path from a root
 func (s *ItemStore) IsValidHierarchy(ids []int64) (bool, error) {
-	return false, nil
+	l := len(ids) - 1
```
smadbe: in my opinion, using a variable for that makes the code more complex to understand. it not even easy to express what `l` represents

app/database/itemStore.go
```diff
@@ -38,9 +38,31 @@ func (s *ItemStore) Insert(data *Item) error {
...
+	if l < 1 {
+		return false, fmt.Errorf("not enough ids %v", ids)
```
smadbe: meaning `len(ids) < 2` ? there is no problem of having a hierarchy of 1 element... i.e., here asking for the breadcrumb for the root only should be valid

smadbe: side remark: When you do testing (and especially unit testing and you add a case such as this one, it MUST be included in your tests!

app/database/itemStore.go
```diff
@@ -38,9 +38,31 @@ func (s *ItemStore) Insert(data *Item) error {
...
+	}
+
+	items := s.Items().All().Where("items.ID IN (?)", ids[1:]).Or("items.ID = ? AND items.sType = ?", ids[0], "Root")
+	all := s.ItemItems().All()
+	for i := 0; i < l; i++ {
+		all.DB = all.Or("items_items.idItemParent = ? AND items_items.idItemChild = ?", ids[i], ids[i+1])
```
smadbe: you are not supposed to modify the inner db of a request, you should see that as immutable

app/database/itemStore.go
```diff
@@ -38,9 +38,31 @@ func (s *ItemStore) Insert(data *Item) error {
...
 // IsValidHierarchy gets an ordered set of item ids and returns whether they forms a valid item hierarchy path from a root
 func (s *ItemStore) IsValidHierarchy(ids []int64) (bool, error) {
-	return false, nil
+	l := len(ids) - 1
+	if l < 1 {
+		return false, fmt.Errorf("not enough ids %v", ids)
+	}
+
+	items := s.Items().All().Where("items.ID IN (?)", ids[1:]).Or("items.ID = ? AND items.sType = ?", ids[0], "Root")
+	all := s.ItemItems().All()
```
smadbe: `all` is probably not the most appropriate name here, make it difficult to understand in the subquery

app/database/itemStore.go
```diff
@@ -38,9 +38,31 @@ func (s *ItemStore) Insert(data *Item) error {
...
+
 // IsValidHierarchy gets an ordered set of item ids and returns whether they forms a valid item hierarchy path from a root
 func (s *ItemStore) IsValidHierarchy(ids []int64) (bool, error) {
-	return false, nil
+	l := len(ids) - 1
+	if l < 1 {
+		return false, fmt.Errorf("not enough ids %v", ids)
+	}
+
+	items := s.Items().All().Where("items.ID IN (?)", ids[1:]).Or("items.ID = ? AND items.sType = ?", ids[0], "Root")
```
smadbe: Using the `items` table is only needed to check the root (trivial request). The rest of the filter (with the expensive `IN (...)` is useless as its `items_items` which is checking the relations. 
Actually I think I see what you tried to do, by doing everything in a single query instead of two, but in this case, this the very inefficient I think. Checking the root in one first request is so trivial. And even with a single request, doing it with a basic join (without any condition on `items`) and then checking the root in code would be better (in my opinion).

app/api/items/testdata/restoredb.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
vyloy: Yeah, I know about it.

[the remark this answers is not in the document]
vyloy: But it depends on mysql, which not installed on my pc.
And I dont want to change the config of docker-compose.
The password is in the projects files on github already.
I dont think it will be used for production env...
So I use the mysql command in the docker instead of install anything.

app/api/items/get-list_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"os/exec"
+	"testing"
+)
+
+func TestIsValidHierarchy(t *testing.T) {
+	assert := assert_lib.New(t)
+
+	config.Path = "testdata/default.yaml"
+	conf, err := config.Load()
+	if err != nil {
```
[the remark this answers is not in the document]
vyloy: Errors which are not about the logic of tests, like read files and db operations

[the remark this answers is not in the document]
vyloy: Fatal directly are more clear instead of use other lib to fatal, lol.
and can share code between real code and test code.

app/api/items/testdata/restoredb.sh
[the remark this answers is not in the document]
smadbe: nothing in what was in place required to have mysql installed

app/api/items/get-list_test.go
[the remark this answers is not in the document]
smadbe: don't use `assertlib` then, the use is for it is to be able to fail directly as well but with the short syntax it offers.

app/api/items/get-list.go
```diff
@@ -29,8 +29,11 @@ func (srv *Service) getList(w http.ResponseWriter, r *http.Request) s.APIError {
 		return s.ErrForbidden(errors.New("Insufficient access on given item ids"))
 	}
 
-	// Todo: validate the hierarchy
-	// srv.Store.Items.IsValidHierarchy(...)
+	if valid, err := srv.Store.Items().IsValidHierarchy(ids); err != nil {
+		return s.ErrUnexpected(err)
+	} else if !valid {
```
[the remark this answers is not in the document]
smadbe: for readability (and variable scoping), probably it would be better to have the assignment on a specific line anyway

PR #22, opened 2018-12-22
Remove aliases for imports

slomek:
Aliasing is not necessary unless we have multiple imports that share their name.

### DIFF
--- a/app/api/groups/get-all.go
+++ b/app/api/groups/get-all.go
@@ -5,10 +5,10 @@ import (
 
 	"github.com/go-chi/render"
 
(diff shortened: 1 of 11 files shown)


slomek:
I get your point, so I reverted the changes regarding assert library 👍

smadbe:
good for me. are you able to merge?

slomek:
Nope, I get _Merging is blocked_ and _The base branch restricts merging to authorized users._

smadbe:
better now?

slomek:
Yup 🙂

PR #25, opened 2019-01-07
Verify permission on item creation

smadbe:


app/database/itemStore.go
```diff
@@ -38,6 +38,30 @@ func (s *ItemStore) Insert(data *Item) error {
...
+
+	var dbRes = []struct {
+		ItemID        int64 `sql:"column:idItem"`
+		ManagerAccess bool  `sql:"column:bManagerAccess"`
+		OwnerAccess   bool  `sql:"column:bOwnerAccess"`
+	}{}
+
+	db := s.GroupItems().MatchingUserAncestors(user).
+		Select("idItem, bManagerAccess, bOwnerAccess").
+		Where("groups_items.idItem = ?", itemID).
```
slomek:
```suggestion
		Where("idItem = ?", itemID).
```

smadbe: it depends a bit what's in "MatchingUserAncestors"  but apparently that's ok here

PR #28, opened 2019-01-09
add the LanguageID to the user profile context using a join so that it can be used with table using the language id

smadbe:
to be used to get the user default language for items

### DIFF
--- a/app/database/userStore.go
+++ b/app/database/userStore.go
@@ -5,14 +5,17 @@ type UserStore struct {
 	*DataStore
 }
 
-// GetByID populates `dest` with the user identified by userID
-func (s *UserStore) GetByID(userID int64, dest interface{}) error {
-	return s.ByID(userID).Scan(dest).Error
+// GetProfileByID populates `dest` with the user identified by userID
+func (s *UserStore) GetProfileByID(userID int64, dest interface{}) error {
+	return s.ByID(userID).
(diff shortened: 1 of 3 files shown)


app/database/userStore.go
```diff
@@ -5,14 +5,17 @@ type UserStore struct {
 	*DataStore
 }
 
-// GetByID populates `dest` with the user identified by userID
-func (s *UserStore) GetByID(userID int64, dest interface{}) error {
-	return s.ByID(userID).Scan(dest).Error
+// GetProfileByID populates `dest` with the user identified by userID
+func (s *UserStore) GetProfileByID(userID int64, dest interface{}) error {
```
slomek: This function should return two things: user and an error, so that you limit the risk of panicking when called with an incorrect type. Scanning SQL results into a type should be done as close to the database as possible.

slomek: Also, following Go conventions, this function should be called `ProfileByID(..)`

smadbe: This function does not know anything about the attribute being requested. One part of the application may require only the username, and another one everything.  I'm not gonna write 10 GetUserSomething functions, one with every combination of value... ah and each requiring its own struct. 
Actually, this lib is supposed to be just a light wrapper around gorm for reused code, it should not hide to the using service all the db. Some part of the SQL (through the builder) is done in the services.
Maybe we should discuss all this on audio chat.

PR #29, opened 2019-01-11
Finalize `add-item` service

smadbe:
To be reviewed commit per commit to make things easier/clearer


app/database/db.go
```diff
@@ -115,3 +106,24 @@ func (db *DB) insert(tableName string, data interface{}) error {
 	db = &DB{db.Exec(query, values...)}
 	return db.Error
 }
+
+func decodeValue(value interface{}) (unWrappedValue interface{}, isNull bool, skip bool) {
```
slomek: Why do you need to do this instead of implementing `Scan` functions on the type?

app/types/bool.go
```diff
@@ -0,0 +1,75 @@
...
+	// NullableBool must be set and can be null
+	NullableBool struct{ Bool }
+	// OptionalBool can be not set. If set, cannot be null.
+	OptionalBool struct{ Bool }
+	// OptNullBool can be not set or null
+	OptNullBool struct{ Bool }
+)
+
+// NewBool creates a Bool which is not-null and set with the given value
+func NewBool(s bool) *Bool {
```
slomek: Building custom types and wrapping gorm with yet another, custom layer of abstraction seem like a very bad idea. I think you should operate on `bool` instead of using this.

app/types/datetime.go
```diff
@@ -0,0 +1,76 @@
...
+	// NullableDatetime must be set and can be null
+	NullableDatetime struct{ Datetime }
+	// OptionalDatetime can be not set. If set, cannot be null.
+	OptionalDatetime struct{ Datetime }
+	// OptNullDatetime can be not set or null
+	OptNullDatetime struct{ Datetime }
+)
+
+// NewDatetime creates a Datetime which is not-null and set with the given value
+func NewDatetime(t time.Time) *Datetime {
```
slomek: Ditto, SQL clients are capable of operating on `time.Time` values, there should be no need of creating such types.

slomek: https://github.com/go-sql-driver/mysql/blob/749ddf1598b47e3cd909414bda735fe790ef3d30/utils.go#L123

app/types/bool.go
slomek: https://github.com/golang/go/blob/master/src/database/sql/sql.go#L266

testhelpers/steps.go
```diff
@@ -233,6 +235,14 @@ func (ctx *TestContext) IAmUserWithID(id int64) error { // nolint
 	return nil
 }
 
+func (ctx *TestContext) TimeNow(timeStr string) error { // nolint
+	testTime, err := time.Parse(time.RFC3339Nano, timeStr)
+	if err == nil {
+		monkey.Patch(time.Now, func() time.Time { return testTime })
```
slomek: Instead of patching the time, you should extract your logic so that the time is an input parameter and unit-test that. When it comes to the end-to-end tests, you should not check the time, because using solutions like this you are risking that you are altering the behaviour of the logic under test.

smadbe: It is a discussion on StackOverflow about abused dependency injection pattern (abused DI pattern)... and with time in Go, you need very "constructor" to start passing the time from the app to every component and service. If you have a smart way, I'm interested in seeing it

app/types/datetime.go
smadbe: this is not related with that, this is for the same reason we have the types/int64, to be able to distinguish between not set, set to default and set to null, which all arrive as the same value in Go.

app/types/bool.go
smadbe: that's not become of that, we already discussed it. see comment below

app/database/db.go
smadbe: To manage the set and null property. we should figure out by reading the function. Actually, the code was already there, it was just moved to a function as a linter was complaining.

slomek: But why do you need that instead of managing those properties in `Scan()`?

app/types/datetime.go
slomek: You have a built-in type for that as well: https://github.com/golang/go/blob/master/src/database/sql/sql.go#L214

app/database/db.go
smadbe: Ah. I think there was something which not working with Scan in this case. But should check if it is still relevant.

app/types/datetime.go
smadbe: Still mixing a null value given and "not set".

smadbe: Actually, this is an interesting debate but a bit out of scope of this PR this was introduced earlier. I see this wrapped type is not convenient in Go, but I still find ridiculous to drop a sound feature because of a language stupidity (personnal opinion 😁).

testhelpers/steps.go
slomek: As I wrote, my suggestion would be to instead of calling a function that uses `time.Now()` inside, you provide a `time.Time` value as an input argument. Not sure why you need to check the whole row in the end-to-end tests. Those should be only to make sure the high-level, business functionality works, not the low-level implementation details such as database structure or data stored is a particular row.

app/database/db.go
```diff
@@ -158,3 +149,24 @@ func (conn *db) insert(tableName string, data interface{}) error {
 	query := fmt.Sprintf("INSERT INTO `%s` (%s) VALUES (%s)", tableName, strings.Join(attributes, ", "), strings.Join(valueMarks, ", ")) // nolint: gosec
 	return conn.Exec(query, values...).Error
 }
+
+func decodeValue(value interface{}) (unWrappedValue interface{}, isNull bool, skip bool) {
+	isNull = false
+	skip = false
+	unWrappedValue = value
+	switch value.(type) {
```
golangcibot: assigning the result of this type assertion to a variable (switch value := value.(type)) could eliminate the following type assertions:
	app/database/db.go:159:10
	app/database/db.go:162:10
	app/database/db.go:165:10
	app/database/db.go:168:10 (from `gosimple`)

PR #31, opened 2019-01-28
Implement the items hierarchy validation

zenovich:


zenovich:
This one depends on #33

app/database/itemStore.go
```diff
@@ -64,7 +74,19 @@ func (s *ItemStore) HasManagerAccess(user AuthUser, itemID int64) (found bool, a
 
 // IsValidHierarchy gets an ordered set of item ids and returns whether they forms a valid item hierarchy path from a root
 func (s *ItemStore) IsValidHierarchy(ids []int64) (bool, error) {
-	return false, nil
+	if len(ids) == 0 {
+		return false, nil
+	}
+
+	if valid, err := s.checkIfItemIsRoot(ids[0]); !valid || err != nil {
+		return false, err
```
smadbe: shouldn't you return `valid, err` instead? here you are turning something valid but with err (don't know for what reason) into something not valid

app/database/itemStore.go
```diff
@@ -119,3 +141,42 @@ func checkAccessForID(id int64, last bool, accDets []itemAccessDetails) error {
 	// no row matching this item_id
 	return fmt.Errorf("not visible item_id %d", id)
 }
+
+func (s *ItemStore) checkIfItemIsRoot(id int64) (bool, error) {
```
smadbe: I would call it `isItemRoot`instead. Tell me if you do not agree.

app/database/itemStore.go
```diff
@@ -119,3 +141,42 @@ func checkAccessForID(id int64, last bool, accDets []itemAccessDetails) error {
...
+	if err := s.ByID(id).Count(&count).Error(); err != nil {
+		return false, err
+	}
+	if count == 0 {
+		return false, nil
+	}
+	return true, nil
+}
+
+func (s *ItemStore) checkHierarchicalChain(ids []int64) (bool, error) {
```
smadbe: same for the naming remark

app/database/itemStore.go
```diff
@@ -119,3 +141,42 @@ func checkAccessForID(id int64, last bool, accDets []itemAccessDetails) error {
 	// no row matching this item_id
 	return fmt.Errorf("not visible item_id %d", id)
 }
+
+func (s *ItemStore) checkIfItemIsRoot(id int64) (bool, error) {
+	count := 0
+	if err := s.ByID(id).Count(&count).Error(); err != nil {
```
smadbe: I think you forgot to check that the type is root.

app/database/db.go
```diff
@@ -98,6 +105,10 @@ func (conn *db) Scan(dest interface{}) DB {
 	return &db{conn.DB.Scan(dest)}
 }
 
+func (conn *db) Count(dest interface{}) DB {
+	return &db{conn.DB.Count(dest)}
```
smadbe: would it be easier to use if returning `(int, DB)` instead? (open question)

app/database/itemStore.go
```diff
@@ -119,3 +141,42 @@ func checkAccessForID(id int64, last bool, accDets []itemAccessDetails) error {
...
+	}
+
+	db := s.ItemItems().All()
+	previousID := ids[0]
+	for index, id := range ids {
+		if index == 0 {
+			continue
+		}
+
+		db = db.Or("idItemParent=? AND idItemChild=? AND iChildOrder=1", previousID, id)
```
smadbe: Dor `iChildOrder`, it is used to set the order of the child among all the parents children. (it is not the depth). Even if it is not the first child, it is still a child.

features/get-all-items-breadcrumbs.feature
```diff
@@ -84,30 +108,45 @@ Scenario: Corrupt breadcrumb hierarchy (one parent-child link missing), but user
...
+  Then the response code should be 400
   And the response body should be, in JSON:
     """
-    [
-    { "item_id": 21, "language_id": 1, "title": "Graph: Methods" },
-    { "item_id": 22, "language_id": 1, "title": "DFS" },
-    { "item_id": 23, "language_id": 1, "title": "Reduce Graph" },
-    { "item_id": 21, "language_id": 2, "title": "Graphe: Methodes" }
-    ]
+    {"success":false,"message":"Bad Request","error_text":"The IDs chain is corrupt"}
```
smadbe: Side note: The same it has been done for `add-item`, I would prefer moving the error handling case to a different file (`get-all-items-breadcrumbs.robustness.feature`). This way it is easier to reflect on all the happy paths (all scenarios) on one side, and on all possible errors handled on the other.  
As there are other error cases already in this file, we can do this once it has been merged.  
The same way, the error handling should be using the step "the response error message should contain ..." instead of exact error matching.

zenovich:
> I haven't had the time to review your tests yet, but from I saw, it is great work. You understood exactly what was expected and how the current code is organized. 👍 👍
Thanks! I'm very pleased.

app/database/itemStore.go
zenovich: > shouldn't you return `valid, err` instead? here you are turning something valid but with err (don't know for what reason) into something not valid

Actually, we never return true together with error, but I'll fix this anyway.

app/database/db.go
zenovich: > would it be easier to use if returning `(int, DB)` instead? (open question)

I think it would not :( The method writes the data into the given argument similar to Scan(). I prefer keeping it similar to Gorm's method.

app/api/items/get-all-items-breadcrumbs.robustness.feature
```diff
@@ -0,0 +1,129 @@
+@wip
```
smadbe: @wip (work in progress) should just be there for debugging

PR #36, opened 2019-01-29
add missing tests ("types" pkg)

smadbe:
Testing the `AllAttributes` implem, quite trivial.

### DIFF
--- a/app/types/bool_test.go
+++ b/app/types/bool_test.go
@@ -21,10 +21,15 @@ func (v *SampleBoolInput) validate() error {
 
 func TestNewBool(t *testing.T) {
 	assert := assertlib.New(t)
-	n := NewBool(true)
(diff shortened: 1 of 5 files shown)


app/types/bool_test.go
```diff
@@ -21,10 +21,15 @@ func (v *SampleBoolInput) validate() error {
 
 func TestNewBool(t *testing.T) {
 	assert := assertlib.New(t)
-	n := NewBool(true)
-	assert.Equal(true, n.Value)
+	var value = true
+	n := NewBool(value)
+	val, null, set := n.AllAttributes()
+	assert.Equal(value, n.Value)
+	assert.Equal(value, val)
```
zenovich: The 'value' is alway true, so we don't need the variable here (my editor complains about this, that's why I changed these lines yesterday)

PR #37, opened 2019-01-29
Fix error strings and other CodeReviewComments-related issues

zenovich:
+ provide information about a field's name when validation fails


zenovich:
> 👍
> I would have prefered a different commit for each "topic" (error strings, defer's, assertlib, ...) but that's great anyway

thanks :)

PR #38, opened 2019-01-29
100% test coverage for "service" package

smadbe:
100% test coverage for "service" package

### DIFF
--- a/app/service/responses_test.go
+++ b/app/service/responses_test.go
@@ -1,0 +1,36 @@
+package service
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-chi/render"
+	assertlib "github.com/stretchr/testify/assert"
+)
+
+func httpResponseForResponse(renderer render.Renderer) *httptest.ResponseRecorder {
+	var fn AppHandler = func(respW http.ResponseWriter, req *http.Request) APIError {
+		_ = render.Render(respW, req, renderer)
+		return NoError
+	}
+	handler := http.HandlerFunc(fn.ServeHTTP)
+
+	req, _ := http.NewRequest("GET", "/dummy", nil)
+	recorder := httptest.NewRecorder()
+
+	handler.ServeHTTP(recorder, req)
+	return recorder
+}
+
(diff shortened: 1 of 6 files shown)


app/service/responses_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-chi/render"
+	assertlib "github.com/stretchr/testify/assert"
+)
+
+func httpResponseForResponse(renderer render.Renderer) *httptest.ResponseRecorder {
+	var fn AppHandler = func(respW http.ResponseWriter, req *http.Request) APIError {
+		render.Render(respW, req, renderer)
```
golangcibot: Error return value of `render.Render` is not checked (from `errcheck`)

app/service/utils.go
```diff
@@ -9,9 +9,12 @@ import (
 
 // QueryParamToInt64Slice extracts from the query parameter of the request a list of integer separated by commas (',')
 func QueryParamToInt64Slice(req *http.Request, paramName string) ([]int64, error) {
-
-	idsStr := strings.Split(req.URL.Query().Get(paramName), ",")
-	var ids []int64
+	ids := []int64{}
```
zenovich: I would rather prefer `var ids []int64` (see here https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

zenovich: I would rather prefer var ids []int64 (see here https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

smadbe: it is not the same. with `var ids []int64` I get a nil while I want to start with an empty array

zenovich: Nil is not a problem https://play.golang.org/p/nUbbKcKhxaG

zenovich: https://play.golang.org/p/LRoPTFtNurN

smadbe: But for my function, it is. My function is returning an empty list (as tested and as used) when there is no IDs. If I use nil at this line, it might return nil in some cases with is not expected (in tests, by caller)

PR #41, opened 2019-02-06
Remove a duplicate from breadcrumbs tests

zenovich:


smadbe:
For this kind of PRs, it would be useful to give a short explanation of what it is, so that it is easier to review. 🙂

zenovich:
> For this kind of PRs, it would be useful to give a short explanation of what it is, so that it is easier to review. 🙂

I just removed a copy of the 'Full access on all breadcrumb' spec. The diff is so big just because of formatting

zenovich:
> Thx. Easier to review with the explanation 👍

Sorry about that, I didn't expect such an ambigous diff.

PR #27, opened 2019-01-08
WIP: loading tree struture of the item

slomek:

PR #42, opened 2019-02-11
Implement the itemView service

zenovich:
(includes #27)


app/database/itemStore.go
```diff
@@ -106,21 +108,214 @@ func (s *ItemStore) GetRawNavigationData(rootID int64, userID, userLanguageID in
...
+			"union_table.sFullScreen, " +
+			"union_table.bShowSource, " +
+			"union_table.iValidationMin, " +
+			"union_table.bShowUserInfos, " +
+			"union_table.sContestPhase, " +
+			"union_table.sUrl, " +
+			"union_table.bUsesAPI, " +
+			"union_table.bHintsAllowed, " +
+			"accessRights.fullAccess, accessRights.partialAccess, accessRights.grayedAccess, accessRights.accessSolutions "+
+			"FROM " +
```
smadbe: could you align the `FROM` with the initial `SELECT` so that it is easier to find the different part?
In the same way, for the next line, I would do something like:
```
FROM (
  SELECT 
    items.ID. ...
    ...
  FROM ...
)
```
etc with indent from the subqueries

app/database/itemStore.go
```diff
@@ -106,21 +108,214 @@ func (s *ItemStore) GetRawNavigationData(rootID int64, userID, userLanguageID in
...
+			"items.idDefaultLanguage, " +
+			" iChildOrder, sCategory, bAlwaysVisible, bAccessRestricted FROM items " +
+			" JOIN items_items ON items.ID=idItemChild AND idItemParent=?) union_table " +
+			"LEFT JOIN users_items ON users_items.idItem=union_table.ID AND users_items.idUser=? " +
+			"LEFT JOIN items_strings dstrings FORCE INDEX (idItem) " +
+			" ON dstrings.idItem=union_table.ID AND dstrings.idLanguage=union_table.idDefaultLanguage " +
+			"LEFT JOIN items_strings ustrings ON ustrings.idItem=union_table.ID AND ustrings.idLanguage=? " +
+			"JOIN ? accessRights on accessRights.idItem=union_table.ID AND (fullAccess>0 OR partialAccess>0 OR grayedAccess>0) " +
+			"ORDER BY iChildOrder",
+		rootID, rootID, userID, userLanguageID, s.accessRightsSubQuery(user)).Scan(&result).Error(); err != nil {
```
smadbe: Global comment: at least the benefit of using the SQL builder is that the subqueries are easier to identify (if the variable has a proper name + comment if needed), and query parameters (?) are not 50lines further.

app/database/itemStore.go
```diff
@@ -106,21 +108,214 @@ func (s *ItemStore) GetRawNavigationData(rootID int64, userID, userLanguageID in
...
+		return nil, err
+	}
+	return &result, nil
+}
+
 func (s *ItemStore) accessRightsSubQuery(user AuthUser) interface{} {
 	return s.GroupItems().MatchingUserAncestors(user).
-		Select("idItem, MAX(bCachedFullAccess) AS fullAccess, MAX(bCachedPartialAccess) AS partialAccess, MAX(bCachedGrayedAccess) AS grayedAccess").
+		Select(
```
smadbe: This is typically very re-usable by other services.  
Comment: (personal opinion) I would find better to return DB instead of returning Subqueries (unknown interface{}). IMO, it would make more sense that the function caller decide what to do with the query, as a subquery (which is auto-adding the parenthesis around the query, the query with parenthesis, or whatever)

app/database/itemStore.go
```diff
@@ -106,21 +108,214 @@ func (s *ItemStore) GetRawNavigationData(rootID int64, userID, userLanguageID in
...
+			"users_items.nbSubmissionsAttempts AS nbSubmissionsAttempts, " +
+			"users_items.bValidated AS bValidated, " +
+			"users_items.bFinished AS bFinished, " +
+			"users_items.bKeyObtained AS bKeyObtained, " +
+			"users_items.nbHintsCached AS nbHintsCached, " +
+			"users_items.sStartDate AS sStartDate, " +
+			"users_items.sValidationDate AS sValidationDate, " +
+			"users_items.sFinishDate AS sFinishDate, " +
+			"users_items.sContestStartDate AS sContestStartDate, " +
+			"IF(union_table.sType <> 'Chapter', users_items.sState, NULL) AS sState, " +
```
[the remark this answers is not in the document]
zenovich: The problem here is that two of these attrubutes cannot be NULL: ````bHintsAllowed` tinyint(1) NOT NULL, `bUsesAPI` tinyint(1) NOT NULL DEFAULT '1',```

app/database/itemStore.go
```diff
@@ -106,21 +108,214 @@ func (s *ItemStore) GetRawNavigationData(rootID int64, userID, userLanguageID in
...
+	var result []RawItem
+
+	// This query can be simplified if we add a column for relation degrees into `items_ancestors`
+	if err := s.Raw(
+		"SELECT "+
+			"union_table.ID, "+
+			"union_table.sType, "+
+			"union_table.bDisplayDetailsInParent, "+
+			"union_table.sValidationType, "+
+			"COALESCE(union_table.idItemUnlocked, '')<>'' as hasUnlockedItems, "+
```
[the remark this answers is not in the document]
zenovich: There are two cases: NULL and an empty string. We want to consider both as FALSE.

app/database/itemStore.go
```diff
@@ -106,21 +108,214 @@ func (s *ItemStore) GetRawNavigationData(rootID int64, userID, userLanguageID in
...
+	*ItemAccessDetails
+}
+
+// GetRawItemData reads data needed by the getItem service from the DB and returns an array of RawItem's
+func (s *ItemStore) GetRawItemData(rootID, userID, userLanguageID int64, user AuthUser) (*[]RawItem, error) {
+	var result []RawItem
+
+	// This query can be simplified if we add a column for relation degrees into `items_ancestors`
+	if err := s.Raw(
+		"SELECT "+
```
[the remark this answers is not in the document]
zenovich: done

app/database/itemStore.go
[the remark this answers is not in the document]
zenovich: done for the fields that can be NULL in the DB

app/database/itemStore.go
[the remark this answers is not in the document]
zenovich: Added the comment

app/database/itemStore.go
zenovich: done

app/database/itemStore.go
zenovich: I agree. As I see this, we will refactor these huge SQLs later.

app/database/itemStore.go
zenovich: done

PR #47, opened 2019-02-19
Implement groupView service

zenovich:
(contains groupRecentActivity as well)

### DIFF
--- a/app/api/groups/get_group.feature
+++ b/app/api/groups/get_group.feature
@@ -1,0 +1,36 @@
+Feature: Get group by groupID (groupView)
+  Background:
+    Given the database has the following table 'users':
+      | ID | sLogin | tempUser | idGroupSelf | idGroupOwned | sFirstName  | sLastName | sDefaultLanguage |
+      | 1  | owner  | 0        | 21          | 22           | Jean-Michel | Blanquer  | fr               |
+    And the database has the following table 'groups':
+      | ID | sName      | iGrade | sDescription    | sDateCreated         | sType     | sRedirectPath                          | bOpened | bFreeAccess | sPassword  | sPasswordTimer | sPasswordEnd        | bOpenContest |
+      | 11 | Group A    | -3     | Group A is here | 2019-02-06T09:26:40Z | Class     | 182529188317717510/1672978871462145361 | true    | true        | ybqybxnlyo | 01:00:00       | 2017-10-13 05:39:48 | true         |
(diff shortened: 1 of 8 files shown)


app/api/groups/get_group.feature
```diff
@@ -0,0 +1,36 @@
...
+      | ID | idGroupAncestor | idGroupChild | bIsSelf | iVersion |
+      | 75 | 22              | 13           | 0       | 0        |
+
+  Scenario: The user is an owner of the group
+    Given I am the user with ID "1"
+    When I send a GET request to "/groups/13"
+    Then the response code should be 200
+    And the response body should be, in JSON:
+    """
+    {
```
smadbe: don't you think we should return the `ID` as well?

app/api/groups/get_group.go
```diff
@@ -0,0 +1,38 @@
...
+	groupID, err := service.ResolveURLQueryPathInt64Field(r, "group_id")
+	if err != nil {
+		return service.ErrInvalidRequest(err)
+	}
+
+	user := srv.GetUser(r)
+
+	query := srv.Store.GroupAncestors().OwnedByUserID(user.UserID).
+		Joins("JOIN groups ON idGroupChild=groups.ID").
+		Where("idGroupChild = ?", groupID).Select(
```
smadbe: for the pure logic point of view (so readability as well), I would have expected `groups` as root table. So have `srv.Store.Groups().OwnedBy(user).Where(groups.ID = ?, groupID)`  with the `OwnedBy(...)` doing the "matching," e.g. inner-joining with group ancestors and only caring for the "left" part

app/api/groups/get_group.go
```diff
@@ -0,0 +1,38 @@
...
+		`groups.sName, groups.iGrade, groups.sDescription, groups.sDateCreated, groups.sType,
+     groups.sRedirectPath, groups.bOpened, groups.bFreeAccess,
+     groups.sPassword, groups.sPasswordTimer, groups.sPasswordEnd, groups.bOpenContest`).Limit(1)
+
+	var result []map[string]interface{}
+	if err := query.ScanIntoSliceOfMaps(&result).Error(); err != nil {
+		fmt.Println(err)
```
smadbe: println?

zenovich:
> I stop here as it seems there is the other PR included in this one. It makes it very difficult to review. We should either wait for the other PR to be merged or this PR should be on `feat/groupView...<the_other_branch>`

Sorry for this. I thought it's possible since all the task is in the last commit.

app/api/groups/get_group.go
zenovich: fixed

app/api/groups/get_group.feature
zenovich: fixed

app/api/groups/get_group.go
zenovich: fixed

PR #46, opened 2019-02-18
Implement groupRecentActivity service

zenovich:

PR #51, opened 2019-02-21
Refactoring of #46 (groupRecentActivity). "First Blood Part II"

zenovich:


app/database/db.go
```diff
@@ -16,6 +16,11 @@ type DB struct {
 	*gorm.DB
 }
 
+// NewDB wraps *gorm.DB
+func NewDB(db *gorm.DB) *DB {
+	return &DB{db}
```
smadbe: Gorm shouldn't be known by the external world, so it should be public (->`newDB` or using struct directly ;-)). The package should change its engine without impact on the other package.

app/api/groups/recent_activity.go
```diff
@@ -37,13 +37,14 @@ func (srv *Service) getRecentActivity(w http.ResponseWriter, r *http.Request) se
        items.ID AS Item__ID, items.sType AS Item__sType,
 		   users.sLogin AS User__sLogin, users.sFirstName AS User__sFirstName, users.sLastName AS User__sLastName,
 			 IF(user_strings.idLanguage IS NULL, default_strings.sTitle, user_strings.sTitle) AS Item__String__sTitle`).
+		JoinsUserAndDefaultItemStrings(user).
```
smadbe: 👍 clear 🙂

app/database/db.go

PR #52, opened 2019-02-21
Fix multi test run

### LINKED ISSUE #50: Run BDD tests just once per `make test`
Currently, when run `make test` it executes BDD tests twice which makes testing slow. It would be much better to execute them just once.

smadbe:
Fix #50 
@zenovich Don't forget to include the `bdd_test.go` file in your package when you introduce a new package using bdd tests

### DIFF
--- a/app/api/answers/bdd_test.go
+++ b/app/api/answers/bdd_test.go
@@ -1,0 +1,18 @@
+package answers_test
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


zenovich:
> Fix #50
> @zenovich Don't forget to include the `bdd_test.go` file in your package when you introduce a new package using bdd tests

will do

PR #56, opened 2019-02-26
Move ByID() method into DataStore

zenovich:


zenovich:
added a test for that case

PR #60, opened 2019-02-28
"null" in bdd tests (TableAtIDShouldBe())

zenovich:
TableAtIDShouldBe():
1) do not cast columns to CHAR(50);
2) compare to "null" instead of "NULL" (same way as "true" and "false").

### DIFF
--- a/testhelpers/steps.go
+++ b/testhelpers/steps.go
@@ -32,12 +32,12 @@ type dbquery struct {
 // TestContext implements context for tests
 type TestContext struct {
 	// nolint
-	application                      *app.Application // do NOT call it directly, use `app()`
-	userID                           int64            // userID that will be used for the next requests
-	featureQueries                   []dbquery
-	lastResponse                     *http.Response
-	lastResponseBody                 string
-	inScenario                       bool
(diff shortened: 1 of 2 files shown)


testhelpers/steps.go
```diff
@@ -406,13 +406,25 @@ func (ctx *TestContext) TableAtIDShouldBe(tableName string, id int64, data *gher
...
-			sqlValue := rowValPtr[iCol].(*string)
-			if dataValue != *sqlValue {
-				return fmt.Errorf("not matching expected value at row %d, col %s, expected '%s', got: '%v'", iDataRow-1, colName, dataValue, *sqlValue)
+			sqlValue := rowValPtr[iCol].(**string)
+
+			if *sqlValue == nil {
+				sqlValue = &pNullValue
+			}
+
+			if dataValue == "true" && **sqlValue == "1" || dataValue == "false" && **sqlValue == "0" {
```
smadbe: I would prefer having parenthesis for this kind of expressions

zenovich: done

PR #61, opened 2019-03-01
Change return type of DB.inTransaction() & DataStore.InTransaction()

zenovich:
(error) -> (interface{}, error)


zenovich:
After discussing, we decided to drop this PR out

PR #65, opened 2019-03-01
Logging reorg ++

smadbe:
The global goal was to reorganize the logging so that it can be stored in a log file instead of outputting it to stderr. 
Many related had to be fixed to have something proper:
- cleaner way to log
- multiple formats for logging
- 100% test coverage for logging
- db logger working correctly
- saner config for logging
- use local config file to allow each dev to have its own config

To be reviewed commit per commit, should be quite clean.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,10 @@ jobs:
       ALGOREA_DATABASE.CONNECTION.USER: alg_ci_user
       ALGOREA_DATABASE.CONNECTION.PASSWD: dummy_password
       ALGOREA_DATABASE.CONNECTION.DBNAME: ci_db
+      ALGOREA_LOGGING.FORMAT: json
+      ALGOREA_LOGGING.OUTPUT: file
+      ALGOREA_LOGGING.LOGSQLQUERIES: 1
+      ALGOREA_LOGGING.LEVEL: debug
     steps:
       - attach_workspace:
           at: ./
@@ -57,6 +61,8 @@ jobs:
       - store_test_results: &TESTPATH
           path: test-results
       - store_artifacts: *TESTPATH
+      - store_artifacts:
+          path: log
       - run:
           when: on_fail
           name: On failure, run BDD tests with details
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,7 @@ jobs:
           at: ./
       - restore_cache: *CACHEKEYMOD
       - run: sudo apt-get install mysql-client # required for db-restore
+      - run: cp conf/config.sample.yaml conf/config.yaml # as env var are not loaded for entries not in config file (bug)
       - run:
           name: Wait for MySQL
           command: dockerize -wait tcp://127.0.0.1:3306 -timeout 30s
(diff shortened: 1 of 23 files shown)


zenovich:
It looks like the logger escapes color sequences in default output: ```=== RUN   TestBase_SetQueryLimit/with_limit
level=info msg="\x1b[35m(/Users/dzenovich/gowork/src/github.com/France-ioi/AlgoreaBackend/app/database/db.go:158)\x1b[0m \n\x1b[33m[2019-03-04 12:20:11]\x1b[0m  \x1b[36;1m[1.60ms]\x1b[0m  SELECT * FROM `users`   LIMIT 10  \n\x1b[36;31m[0 rows affected or returned ]\x1b[0m "
=== RUN   TestBase_SetQueryLimit/without_limit
level=info msg="\x1b[35m(/Users/dzenovich/gowork/src/github.com/France-ioi/AlgoreaBackend/app/database/db.go:158)\x1b[0m \n\x1b[33m[2019-03-04 12:20:11]\x1b[0m  \x1b[36;1m[6.63ms]\x1b[0m  SELECT * FROM `users`   LIMIT 500  \n\x1b[36;31m[0 rows affected or returned ]\x1b[0m "
=== RUN   TestBase_SetQueryLimit/wrong_limit
level=info msg="\x1b[35m(/Users/dzenovich/gowork/src/github.com/France-ioi/AlgoreaBackend/app/database/db.go:158)\x1b[0m \n\x1b[33m[2019-03-04 12:20:11]\x1b[0m  \x1b[36;1m[0.83ms]\x1b[0m  SELECT * FROM `users`   LIMIT 500  \n\x1b[36;31m[0 rows affected or returned ]\x1b[0m "
--- PASS: TestBase_SetQueryLimit (0.05s)```

zenovich:
Probably, it would be better to use a format like ```

smadbe:
> Probably, it would be better to use a format like
> 
> ```
> [2017-06-22 19:52:11 +0300] INFO : ...
> ```

For the console version of the log? For now, I haven't invested time in formatting, I just reused the existing formatter. These colored formats are supposed to be nice but I understand that it may be better to have something consistent. Anyway as you don't mind it is mixed with tests outputs (both `go-test` and gherkins), more or less unformat should not be a problem 😄. More seriously, it is a matter of writing another (simple) formatter. It may be done but it will be in another PR.

zenovich:
> > Probably, it would be better to use a format like
> > ```
> > [2017-06-22 19:52:11 +0300] INFO : ...
> > ```
> 
> For the console version of the log? For now, I haven't invested time in formatting, I just reused the existing formatter. These colored formats are supposed to be nice but I understand that it may be better to have something consistent. Anyway as you don't mind it is mixed with tests outputs (both `go-test` and gherkins), more or less unformat should not be a problem 😄. More seriously, it is a matter of writing another (simple) formatter. It may be done but it will be in another PR.

At least we need something like 
```
INFO: ...
```
because the current default format (level=info msg="...") contains non-informative words (level, msg) and escapes messages making them unreadable

smadbe:
Should be fixed now. I don't really understand why it didn't detect it was a TTY to display color but now it works. Alternative format may be used, but I think that at least this is better than before for both text-in-termal and json-in-file cases

PR #70, opened 2019-03-07
Extract service.GetResponseForTheRouteWithMockedDBAndUser() method (used in tests)

zenovich:


app/service/utils.go
```diff
@@ -163,6 +168,41 @@ func toSnakeCase(in string) string {
 	return string(out)
 }
 
+// GetResponseForRouteWithMockedDBAndUser executes a route for unit tests
```
smadbe: The functions dedicated to tests should be moved to a specific file (if we consider it as cross-package even, we should move in to "testhelpers" package). It bothers me to have monkey patching lib loaded in a file with the functions called in production. This lib is fragile and we should make sure it is not used by mistake.

app/service/utils.go
```diff
@@ -163,6 +168,41 @@ func toSnakeCase(in string) string {
...
+	defer func() { _ = db.Close() }() // nolint: gosec
+
+	setMockExpectationsFunc(mock)
+
+	base := Base{Store: database.NewDataStore(db), Config: nil}
+	router := chi.NewRouter()
+	setRouterFunc(router, &base)
+
+	monkey.Patch(auth.UserIDFromContext, func(context context.Context) int64 {
+		return userID
```
smadbe: And you do not remove the patch after it then 🤨 . So you do it when it is trivial, not in other case ? ;-) (okay okay, probably I do the same 😉)
Again here as it is not an external uncontrollable package (such as time or random), we should do it properly, it should be quite easy actually. Let me try.

app/auth/mocks_test.go
```diff
@@ -0,0 +1,37 @@
...
+	ts := httptest.NewServer(middleware(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		userID := UserIDFromContext(r.Context())
+		_, _ = w.Write([]byte(strconv.FormatInt(userID, 10)))
+	})))
+	defer ts.Close()
+
+	request, _ := http.NewRequest("GET", ts.URL, nil)
+	response, err := http.DefaultClient.Do(request)
+	assert.NoError(err)
+	respBody, err := ioutil.ReadAll(response.Body)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

smadbe:
I'm done with my part.

app/service/utils.go
zenovich: moved to app/servicetest

PR #71, opened 2019-03-07
Implement groupChangePassword

zenovich:


app/api/groups/change_password.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+retry:
+	newPassword, err := GenerateGroupPassword()
+	service.MustNotBeError(err)
+
+	// `CREATE UNIQUE INDEX sPassword ON groups(sPassword)` must be done
+	err = srv.Store.Groups().Where("ID = ?", groupID).Updates(map[string]interface{}{"sPassword": newPassword}).Error()
+	if err != nil && strings.Contains(err.Error(), "Duplicate entry") {
+		goto retry
```
smadbe: I would handle the case of broken `GenerateGroupPassword` which would always generate the same value and thus "loop" forever (which would be disastrous: kill the DB and block processes)

app/api/groups/change_password.feature
```diff
@@ -0,0 +1,31 @@
...
+    And the database has the following table 'groups_ancestors':
+      | ID | idGroupAncestor | idGroupChild | bIsSelf | iVersion |
+      | 75 | 22              | 13           | 0       | 0        |
+      | 76 | 13              | 11           | 0       | 0        |
+      | 77 | 22              | 11           | 0       | 0        |
+      | 78 | 21              | 21           | 1       | 0        |
+
+  Scenario: User is an admin of the group
+    Given I am the user with ID "1"
+    And the generated group password is "newpassword"
```
smadbe: I know you love to monkey-patch everything but for tests, as the seed is fixed at reset at test start, you should get a fixed predictable value that you can use directly in your tests.

testhelpers/featureContext.go
```diff
@@ -10,9 +10,13 @@ func FeatureContext(s *godog.Suite) {
 	s.BeforeScenario(ctx.SetupTestContext)
 
 	s.Step(`^the database has the following table \'([\w\-_]*)\':$`, ctx.DBHasTable)
+	s.Step(`^the table "([^"]*)" has a unique key "([^"]*)" on "([^"]*)"$`, ctx.TableHasUniqueKey)
```
smadbe: I wouldn't have polluted the current code to simulate what the future DB should be. I think this is typically the kind of stuff that will be difficult to understand in a few months.

testhelpers/steps.go
```diff
@@ -566,3 +592,41 @@ func (ctx *TestContext) TableAtIDShouldBe(tableName string, id int64, data *gher
...
+	passwords := multipleStringsRegexp.FindAllStringSubmatch(generatedPasswords, -1)
+	monkey.Patch(groups.GenerateGroupPassword, func() (string, error) {
+		currentIndex++
+		if currentIndex >= len(generatedPasswords) {
+			return "", errors.New("not enough generated passwords")
+		}
+		return passwords[0][currentIndex], nil
+	})
+	return nil
+}
```
smadbe: The next time we add something in this file, we should split it. It becomes too long and messy.

app/api/groups/change_password.go
zenovich: fixed

app/api/groups/change_password.feature
zenovich: It is impossible to set the seed of "crypto/rand" :(

testhelpers/featureContext.go
zenovich: We can remove this line as soon as the DB is upgraded. I think it would be better to keep the scenario although the DB is not ready yet, because this scenario covers the feature well.
Or do you want me to remove it anyway?

testhelpers/steps.go
zenovich: kk

app/api/groups/change_password.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Change the password of the given group
+
+  Background:
+    Given the database has the following table 'users':
+      | ID | sLogin | tempUser | idGroupSelf | idGroupOwned | sFirstName  | sLastName | sDefaultLanguage |
+      | 1  | owner  | 0        | 21          | 22           | Jean-Michel | Blanquer  | fr               |
+    And the database has the following table 'groups':
+      | ID | sName   | iGrade | sDescription    | sDateCreated         | sType     | sRedirectPath                          | bOpened | bFreeAccess | sPassword  | sPasswordTimer | sPasswordEnd         | bOpenContest |
+      | 11 | Group A | -3     | Group A is here | 2019-02-06T09:26:40Z | Class     | 182529188317717510/1672978871462145361 | true    | true        | ybqybxnlyo | 01:00:00       | 2017-10-13T05:39:48Z | true         |
+      | 13 | Group B | -2     | Group B is here | 2019-03-06T09:26:40Z | Class     | 182529188317717610/1672978871462145461 | true    | true        | ybabbxnlyo | 01:00:00       | 2017-10-14T05:39:48Z | true         |
```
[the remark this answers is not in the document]
zenovich: These passwords are different. I agree, the difference is very small.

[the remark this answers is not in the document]
zenovich: fixed

app/api/groups/change_password.go
```diff
@@ -23,16 +25,24 @@ func (srv *Service) changePassword(w http.ResponseWriter, r *http.Request) servi
...
+		if err != nil && strings.Contains(err.Error(), "Duplicate entry") {
+			continue
+		}
+		service.MustNotBeError(err)
 
-	// `CREATE UNIQUE INDEX sPassword ON groups(sPassword)` must be done
-	err = srv.Store.Groups().Where("ID = ?", groupID).Updates(map[string]interface{}{"sPassword": newPassword}).Error()
-	if err != nil && strings.Contains(err.Error(), "Duplicate entry") {
-		goto retry
+		break
```
smadbe: I know people who would have been mad with a `continue` or a `break`... so here having the if case actually continuing but the continuation actually breaking, except a specific case which panics ... 🤪

Would you mind replacing these 5 lines by this (equivalent but a bit saner in terms of "flow")?
```go
if err == nil {
  break
}
if !strings.Contains(err.Error(), "Duplicate entry") {
  service.MustNotBeError(err)  // or panic(err) if you prefer
}
```

testhelpers/featureContext.go
smadbe: I would not have put it, but ok you can leave it if you prefer.

app/api/groups/change_password.go
zenovich: discussed in Slack

PR #72, opened 2019-03-08
Improve service.ResolveURLQuery* methods to distinguish between missing and wrong values

zenovich:


smadbe:
Should `QueryParamToInt64Slice` in `utils.go` be consistent with the other fcts, so `Resolve...` ?

zenovich:
fixed

PR #73, opened 2019-03-08
Implement common sorting and paging for all services

zenovich:


smadbe:
Sorting as 0% test coverage.

zenovich:
> Sorting has 0% test coverage.

That's because we do not consider what recent_activity*.feature cover

app/service/sorting.go
```diff
@@ -0,0 +1,191 @@
...
+	sortStatements := strings.Split(sortingRules, ",")
+	usedFields = make([]string, 0, len(sortStatements)+1)
+	fieldsDirections = make(map[string]int, len(sortStatements)+1)
+	for _, sortStatement := range sortStatements {
+		fieldName, direction := getFieldNameAndDirectionFromSortStatement(sortStatement)
+		if fieldsDirections[fieldName] != 0 {
+			return nil, nil, fmt.Errorf("a field cannot be a sorting parameter more than once: %q", fieldName)
+		}
+		if _, ok := acceptedFields[fieldName]; !ok {
+			return nil, nil, fmt.Errorf("unknown field in sorting parameters: %q", fieldName)
```
smadbe: "unknown" -> "unallowed"

app/service/sorting.go
```diff
@@ -0,0 +1,191 @@
...
+type FieldSortingParams struct {
+	// ColumnName is a DB column name (may contain a table name as a prefix, e.g. "groups.ID")
+	ColumnName string
+	// FieldType is one of "int64", "bool", "string"
+	FieldType string
+}
+
+// ApplySorting applies ordering and paging according to given accepted fields and sorting rules
+// taking into the account the URL parameters 'from.*'
+func ApplySorting(r *http.Request, query *database.DB, acceptedFields map[string]*FieldSortingParams,
```
smadbe: Shouldn't the name reflect that there is paging done as well?

app/service/sorting.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	return fromValues, nil
+}
+
+func getFromValueForField(r *http.Request, fieldName string,
+	acceptedFields map[string]*FieldSortingParams) (interface{}, error) {
+	fromFieldName := "from." + fieldName
+	switch acceptedFields[fieldName].FieldType {
+	case "string":
+		return r.URL.Query()[fromFieldName][0], nil
```
smadbe: I think you are supposed to use `Get(...)` on `Values` (the result type from `Query()`). In particular, they check that the slice is not empty (I supposed it may happen).

(apply probably to other part of the code)

app/service/sorting_test.go
```diff
@@ -0,0 +1,171 @@
...
+		{name: "one of the from. fields is skipped",
+			args: args{
+				urlParameters: "?from.id=2",
+				acceptedFields: map[string]*FieldSortingParams{
+					"name": {ColumnName: "sName", FieldType: "string"},
+					"id":   {ColumnName: "ID", FieldType: "int64"},
+				},
+				defaultRules: "-name,id",
+			},
+			wantAPIError: ErrInvalidRequest(errors.New(`all 'from' parameters (from.name, from.id) or none of them must be present`))},
```
smadbe: Either the message and the logic are not really exact. (the message I think) You want all (or none) "from" parameters among the given sort attributes, right? If so, the message should say that.

smadbe: Could you add (or update this one) also a test where the expected parameters for "from" are not exactly the same as the fields accepted for sorting? (in short: add a field in "acceptedFields" which is not used)

app/service/sorting.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	return query
+}
+
+func parsePagingParameters(r *http.Request, usedFields []string,
+	acceptedFields map[string]*FieldSortingParams) ([]interface{}, error) {
+	fromValueSkipped := false
+	fromValueAccepted := false
+	fromValues := make([]interface{}, 0, len(usedFields))
+	for _, fieldName := range usedFields {
```
smadbe: I think a bit more comments all around this file would be useful to explain what you are trying to do. I think it will not be easy at all to re-enter into this code later, especially as ordering has an impact on paging.

app/service/sorting.go
```diff
@@ -0,0 +1,191 @@
...
+	ColumnName string
+	// FieldType is one of "int64", "bool", "string"
+	FieldType string
+}
+
+// ApplySorting applies ordering and paging according to given accepted fields and sorting rules
+// taking into the account the URL parameters 'from.*'
+func ApplySorting(r *http.Request, query *database.DB, acceptedFields map[string]*FieldSortingParams,
+	defaultRules string) (*database.DB, APIError) {
+	sortingRules, acceptedFields := prepareSoringRulesAndAcceptedFields(r, acceptedFields, defaultRules)
```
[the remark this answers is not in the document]
zenovich: fixed

app/service/sorting_test.go
```diff
@@ -0,0 +1,171 @@
...
+			wantAPIError: NoError},
+		{name: "sorting + paging",
+			args: args{
+				urlParameters: "?from.id=1&from.name=Joe&from.flag=1",
+				acceptedFields: map[string]*FieldSortingParams{
+					"name": {ColumnName: "sName", FieldType: "string"},
+					"id":   {ColumnName: "ID", FieldType: "int64"},
+					"flag": {ColumnName: "bFlag", FieldType: "bool"},
+				},
+				defaultRules: "-name,id,flag",
```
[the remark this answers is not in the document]
zenovich: It was ordering on flag. There just was a bug in tests. Now it is fixed

app/service/sorting.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func prepareSoringRulesAndAcceptedFields(r *http.Request, acceptedFields map[string]*FieldSortingParams,
+	defaultRules string) (string, map[string]*FieldSortingParams) {
+	newAcceptedFields := make(map[string]*FieldSortingParams, len(acceptedFields)+1)
+	for field, params := range acceptedFields {
+		newAcceptedFields[field] = params
+	}
+	if _, ok := newAcceptedFields["id"]; !ok {
+		newAcceptedFields["id"] = &FieldSortingParams{ColumnName: "ID", FieldType: "int64"}
```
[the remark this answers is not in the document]
zenovich: I think it would be more of DRY to allow services skip this field in declarations adding it to the end implicitly

app/service/sorting.go
zenovich: done

app/service/sorting.go
zenovich: fixed

app/service/sorting.go
zenovich: We check that in parsePagingParameters(): ```
		if len(r.URL.Query()[fromFieldName]) > 0 {
```

app/service/sorting_test.go
zenovich: > Either the message and the logic are not really exact. (the message I think) You want all (or none) "from" parameters among the given sort attributes, right? If so, the message should say that.

The sort attributes are not always given in URL since the 'sort' parameter is optional. So the sort attributes can be default ones of a service as well.

zenovich: > Could you add (or update this one) also a test where the expected parameters for "from" are not exactly the same as the fields accepted for sorting? (in short: add a field in "acceptedFields" which is not used)

done

app/service/sorting.go
zenovich: added more comments

app/service/sorting.go
smadbe: but they created this "Values" abstraction so that you do not do it in code by yourself, why not using it?

zenovich: This "Values" abstraction doesn't allow us to distinguish between absent values and empty values. In both cases it returns an empty string.

app/service/sorting_test.go
smadbe: > > Either the message and the logic are not really exact. (the message I think) You want all (or none) "from" parameters among the given sort attributes, right? If so, the message should say that.
> 
> The sort attributes are not always given in URL since the 'sort' parameter is optional. So the sort attributes can be default ones of a service as well.

My point was that each "from" parameters **must** be among the sorted fields (implicit or explicit). Do you validate that? I do not see an error message matching this.

In the same way, these "from" parameters **must** be fiven in the same order and the "sort" (implicit or explicit). Do you check that?

zenovich: We silently ignore all "from.*" parameters that are not among the sorted fields. It's consistent with the way we process usual get-parameters: all unknown parameters are just ignored.

"from.*" parameters are explored in the order of fields in the "sort" configuration. So there is no need to fix their order in URLs

app/service/sorting.go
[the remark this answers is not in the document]
zenovich: It cannot be "id" agnostic because it implicitly adds the "id" to the end of the "sort" string

app/service/sorting_test.go
zenovich: fixed

PR #75, opened 2019-03-12
Do not allow using from.* fields that are not listed in "sort" even if they are accepted fields

zenovich:


app/service/sorting.go
```diff
@@ -129,7 +129,7 @@ func applyOrder(query *database.DB, usedFields []string, acceptedFields map[stri
 // parsePagingParameters returns a slice of values provided for paging in a request URL (none or all should be present)
 // The values are in the order of the 'usedFields' and converted according to 'FieldType' of 'acceptedFields'
 func parsePagingParameters(r *http.Request, usedFields []string,
-	acceptedFields map[string]*FieldSortingParams) ([]interface{}, error) {
+	acceptedFields map[string]*FieldSortingParams, fieldsDirections map[string]int) ([]interface{}, error) {
```
smadbe: Probably this remark was more for the previous PR but I would not encode directions to 1/-1 explicitly but to a real enum with "asc", "desc" and maybe "unknown", and so the auth values would be one of these... (you would even be able to create a method on this type to return you the SQL direction as a string 😎).

app/service/sorting_test.go
```diff
@@ -137,13 +137,13 @@ func TestApplySorting(t *testing.T) {
 			shouldPanic: errors.New(`unsupported type "interface{}" for field "name"`)},
 		{name: "unallowed from fields",
 			args: args{
-				urlParameters: "?from.field=Joe&from.version=2",
+				urlParameters: "?from.field=Joe&from.version=2&from.name=Jane",
 				acceptedFields: map[string]*FieldSortingParams{
 					"name": {ColumnName: "sName", FieldType: "interface{}"},
```
smadbe: don't use `FieldType: "interface{}"` here as it is not what you are testing

app/service/sorting_test.go
```diff
@@ -137,13 +137,13 @@ func TestApplySorting(t *testing.T) {
 			shouldPanic: errors.New(`unsupported type "interface{}" for field "name"`)},
 		{name: "unallowed from fields",
 			args: args{
-				urlParameters: "?from.field=Joe&from.version=2",
+				urlParameters: "?from.field=Joe&from.version=2&from.name=Jane",
```
smadbe: As there is this "version", it may test something else. Probably it should be 2 different tests (?)

app/service/sorting_test.go
zenovich: fixed

app/service/sorting.go
zenovich: done

smadbe: 👍🙂

app/service/sorting_test.go
zenovich: Both 'version' and 'name' are not listed in `sort`, so here they should be similar (`acceptedFields` should not matter). That's why they are both in the same test.

PR #77, opened 2019-03-14
Cache successfully loaded configs

zenovich:


smadbe:
I haven't checked it into details yet (and I shouldn't have the time to check today)... but aren't you re-inventing the wheel of what the `viper` lib is trying to do? Especially, `viper` will not cache config provided through ENV on purpose to allow changing the config live or at least to set them once the app is already started.

zenovich:
I would prefer never change config at runtime. We can restart the app to apply a new configuration. Hostings like Heroku do that automatically when an env variable is changed.

zenovich:
Without this patch I'm getting 
```
WARNING: DATA RACE
Write at 0x00c0001d6cf0 by goroutine 28:
  runtime.mapassign_faststr()
      /usr/local/opt/go/libexec/src/runtime/map_faststr.go:190 +0x0
  github.com/spf13/viper.(*Viper).SetDefault()
      /Users/dzenovich/gowork/src/github.com/spf13/viper/viper.go:1178 +0x20d
  github.com/spf13/viper.SetDefault()
      /Users/dzenovich/gowork/src/github.com/spf13/viper/viper.go:1167 +0x72
  github.com/France-ioi/AlgoreaBackend/app/config.setDefaults()
      /Users/dzenovich/gowork/src/github.com/France-ioi/AlgoreaBackend/app/config/config.go:95 +0x5f
```

smadbe:
"I would prefer never change config at runtime." > Ok but that's already there, removing it and adding a cache in front is an actual solution IMO.
"Without this patch I'm getting" -> in what case you get this? How can I reproduce it?

zenovich:
The goal of this PR is to prevent config reloading which currently happens every time we access config values.

> in what case you get this? How can I reproduce it?
Just call config.Load() concurrently:
```
config.Load()
for i:=0; i<1000; i++ {
  go config.Load()
}
```

zenovich:
> Ok but that's already there
Is it even possible to change another process' env variables at runtime?

smadbe:
Regarding this PR.
I don't believe in adding more "global" to solve a test concurrency issue. Ok this work, no doubt, but it may bring up more problems than it solves. (actually, the problem is probably already due to something which is "global")
Maybe we should wait for a fix for concurrency from viper. But is it so annoying? Do you really execute tests in parallel?  (on the CI it should not)  

[...] after a bit of investigation:
From what I see, not working on the global var of viper but creating an instance works as well. A much cleaner solution. I recreate one PR with that? I will add a test for concurrency (the loop) if it can help.

zenovich:
For now, we reload the config file every time we want to get a value from the config. I think we should avoid that because it is slow. I suppose, it is more important than the concurrency issues. This PR solves the issue.

PR #80, opened 2019-03-15
Handle errors of db.Begin() in DB.inTransaction()

zenovich:


app/database/db.go
```diff
@@ -37,24 +37,26 @@ func Open(source interface{}) (*DB, error) {
 }
 
 func (conn *DB) inTransaction(txFunc func(*DB) error) (err error) {
+	var transactionStarted = false
 	var txDB = conn.db.Begin()
-	if err != nil {
-		return err
+	if txDB.Error != nil {
```
smadbe: Haha it was so stupid 👍  (due to a change to the initial code). Such a shame that Go does not help spotting this kind of mistake.

app/database/db.go
```diff
@@ -37,24 +37,26 @@ func Open(source interface{}) (*DB, error) {
...
+		return txDB.Error
 	}
+	transactionStarted = true
 	defer func() {
 		if p := recover(); p != nil {
 			// ensure rollback is executed even in case of panic
-			txDB.Rollback()
+			if transactionStarted {
```
smadbe: How can we arrive here with `! transactionStarted` as the `defer` is defined after the "begin/return error"?

zenovich: This function is a closure, so it sees all the variables from the scope that defines it.

zenovich: You're right, we cannot. I'll fix it right now

PR #84, opened 2019-03-19
Fix Config concurrency problem + try to load it less 

smadbe:
To be reviewed commit per commit. 
The abusive calls of `config.Load()` by other parts of the code will be addressed in a separated PR.

### DIFF
--- a/app/config/config_test.go
+++ b/app/config/config_test.go
@@ -43,3 +43,13 @@ func TestLoadConfig(t *testing.T) {
 	// test env
 	assert.EqualValues(999, conf.Server.WriteTimeout)
 }
+
+func TestLoadConfig_Concurrent(t *testing.T) {
+	assert := assertlib.New(t)
+	assert.NotPanics(func() {
+		_, _ = Load()
+		for i := 0; i < 1000; i++ {
+			go func() { _, _ = Load() }()
+		}
+	})
+}
--- a/app/app.go
+++ b/app/app.go
@@ -30,6 +30,9 @@ func New() (*Application, error) {
 		return nil, err
 	}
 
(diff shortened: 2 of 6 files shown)


app/config/config_test.go
```diff
@@ -43,3 +43,13 @@ func TestLoadConfig(t *testing.T) {
 	// test env
 	assert.EqualValues(999, conf.Server.WriteTimeout)
 }
+
+func TestLoadConfig_Concurrent(t *testing.T) {
+	assert := assertlib.New(t)
+	assert.NotPanics(func() {
+		Load()
```
golangcibot: Error return value of `Load` is not checked (from `errcheck`)

app/config/config_test.go
```diff
@@ -43,3 +43,13 @@ func TestLoadConfig(t *testing.T) {
...
+		for i := 0; i < 1000; i++ {
+			go Load()
```
golangcibot: Error return value of `Load` is not checked (from `errcheck`)

app/config/config_test.go
```diff
@@ -43,3 +43,13 @@ func TestLoadConfig(t *testing.T) {
 	// test env
 	assert.EqualValues(999, conf.Server.WriteTimeout)
 }
+
+func TestLoadConfig_Concurrent(t *testing.T) {
+	assert := assertlib.New(t)
+	assert.NotPanics(func() {
+		_, _ = Load()
```
zenovich: This line is probably unnecessary.

smadbe: it is not (to make it fail when it was failing) but don't ask me why 😁

PR #89, opened 2019-03-21
remove forgotten unused monkey-patch

smadbe:
I forgot to remove these few lines of code after updating the test.


smadbe:
True. I thought there was a fallback solution when there was no config but that's not the case here.

PR #94, opened 2019-03-28
Retry transactions on getting deadlocks

zenovich:


zenovich:
There is a good reason for sleeping before transaction retrying: TestGroupGroupStore_CreateNewAncestors_Concurrent doesn't work without it. The jitter is nice to have too, so I added it.

PR #99, opened 2019-04-04
Change group password inside a transaction

zenovich:


app/api/groups/change_password.go
```diff
@@ -26,23 +27,26 @@ func (srv *Service) changePassword(w http.ResponseWriter, r *http.Request) servi
...
 
 	var newPassword string
-	for retryCount := 1; ; retryCount++ {
-		if retryCount > 3 {
-			return service.ErrUnexpected(errors.New("the password generator is broken"))
-		}
+	service.MustNotBeError(srv.Store.InTransaction(func(store *database.DataStore) error {
+		for retryCount := 1; ; retryCount++ {
+			if retryCount > 3 {
+				return errors.New("the password generator is broken")
```
smadbe: This specific error should be logged as well. (using `GetLogEntry(r)....`).
(yeah I know it is not directly related to the change.)

zenovich: fixed

PR #100, opened 2019-04-05
Implement groupAddChild service

zenovich:


app/database/group_group_store.go
```diff
@@ -22,3 +27,53 @@ func (s *GroupGroupStore) WhereUserIsMember(user *User) *DB {
...
+
+func (s *GroupGroupStore) CreateRelation(parentGroupID, childGroupID int64) (err error) {
+	s.mustBeInTransaction()
+	defer recoverPanics(&err)
+
+	mustNotBeError(s.WithNamedLock(s.tableName, groupsRelationsLockTimeout, func(store *DataStore) (err error) {
+		var rows []interface{}
+		mustNotBeError(s.WithWriteLock().
+			Select("ID").
+			Where("idGroupChild = ? AND idGroupParent = ?", childGroupID, parentGroupID).
```
smadbe: If it exists but is "removed" I supposed it can be recreated.

app/database/group_group_store.go
```diff
@@ -22,3 +27,53 @@ func (s *GroupGroupStore) WhereUserIsMember(user *User) *DB {
...
+			Limit(1).
+			Scan(&rows).Error())
```
smadbe: In MySQL there are no perf difference with a count? (or other constructs)

app/api/groups/add_child.feature
```diff
@@ -0,0 +1,42 @@
...
+    And the response body should be, in JSON:
+    """
+    {
+      "success": true,
+      "message": "created"
+    }
+    """
+    And the table "groups_groups" should be:
+      | idGroupParent | idGroupChild | iChildOrder | sType  | sRole  |
+      | 13            | 11           | 1           | direct | member |
```
smadbe: I would expect tests checking the order computation is valid in practice.

app/database/group_group_store.go
```diff
@@ -22,3 +27,53 @@ func (s *GroupGroupStore) WhereUserIsMember(user *User) *DB {
...
+			return ErrRelationCycle
+		}
+
+		groupGroupStore := store.GroupGroups()
+
+		mustNotBeError(groupGroupStore.db.Exec(
+			"SET @maxIChildOrder = IFNULL((SELECT MAX(iChildOrder) FROM `groups_groups` WHERE `idGroupParent` = ? FOR UPDATE), 0)",
+			parentGroupID).Error)
+
+		mustNotBeError(groupGroupStore.retryOnDuplicatePrimaryKeyError(func(db *DB) error {
```
smadbe: Shouldn't `retryOnDuplicatePrimaryKeyError` take care of providing the ID directly as this is the only usage of this function?

app/api/groups/add_child.robustness.feature
```diff
@@ -0,0 +1,140 @@
...
+    And the table "groups_groups" should stay unchanged
+    And the table "groups_ancestors" should stay unchanged
+
+  Scenario: A group cannot become an ancestor of itself
+    Given I am the user with ID "4"
+    When I send a POST request to "/groups/11/add_child/13"
+    Then the response code should be 403
+    And the response error message should contain "A group cannot become an ancestor of itself"
+    And the table "groups_groups" should stay unchanged
+    And the table "groups_ancestors" should stay unchanged
```
smadbe: missing test: 

- [ ] given parent and child are the same. I know it should fail on ancestors for the moment but if the ancestor cache stuff changes, this should still be prevented, so at least a test should validate it.
- [ ] the "removed" sType case (maybe in `add_child.feature`)

app/database/group_group_store.go
zenovich: There is a bug in 'gorm'. It skips the write lock ('FOR UPDATE') when counting rows. Probably we should create a PR with the bugfix. It's just 3 lines of code. But without this patch we cannot use Count() here :(

app/api/groups/add_child.feature
```diff
@@ -0,0 +1,42 @@
...
+      | ID | sName   | sType     |
+      | 11 | Group A | Class     |
+      | 13 | Group B | Class     |
+      | 14 | Group C | UserAdmin |
+      | 22 | Owned   | UserAdmin |
+    And the database has the following table 'groups_ancestors':
+      | idGroupAncestor | idGroupChild | bIsSelf |
+      | 21              | 21           | 1       |
+      | 22              | 11           | 0       |
+      | 22              | 13           | 0       |
```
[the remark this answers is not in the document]
zenovich: fixed

app/database/db.go
```diff
@@ -391,6 +391,23 @@ func (conn *DB) WithWriteLock() *DB {
...
+		err := f(conn)
+		if err != nil {
+			if e, ok := err.(*mysql.MySQLError); ok && e.Number == 1062 && strings.Contains(e.Message, "for key 'PRIMARY'") {
+				continue // retry with a new ID
+			}
+			return err
+		}
+		return nil
+	}
+	return errors.New("cannot generate a new ID")
```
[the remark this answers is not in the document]
zenovich: added logging here

app/api/groups/add_child.feature
zenovich: fixed

app/database/group_group_store.go
zenovich: I have done it this way because (imho) DB should not depend on DataStore

app/api/groups/add_child.robustness.feature
zenovich: The first one actually returns "insufficient access rights" error

zenovich: added a scenario for the first case

zenovich: added a scenario for relation rewriting

app/database/group_group_store.go
zenovich: from now it will always recreate a relation, no matter which type it has

app/database/db_test.go
```diff
@@ -507,15 +507,33 @@ func TestDB_Take(t *testing.T) {
...
+	db, mock := NewDBMock()
+	defer func() { _ = db.Close() }()
+
+	mock.ExpectExec(regexp.QuoteMeta("DELETE FROM `myTable`") + `\s+` +
+		regexp.QuoteMeta("WHERE (ID = 1)")).
+		WillReturnResult(sqlmock.NewResult(-1, 1))
+
+	db = db.Table("myTable")
+
+	deleteDB := db.Delete(nil, "ID = 1")
```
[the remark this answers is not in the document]
zenovich: improved the test

PR #104, opened 2019-04-12
Always call createNewAncestors() if at least one relation is modified in GroupGroupStore.Transition()

zenovich:
We need to do this because our triggers mark relations as "todo" no matter what sType they have.


zenovich:
> Cannot you add a test failing before and fixed by this PR? (to avoid regressions)

will do right now

zenovich:
updated the test

PR #106, opened 2019-04-17
Implement Db.Pluck() and fix small issues

zenovich:


app/database/db.go
```diff
@@ -314,6 +314,17 @@ func (conn *DB) Count(dest interface{}) *DB {
 	return newDB(conn.db.Count(dest))
 }
 
+// Pluck used to query single column from a model as a map
```
smadbe: I know that is from gorm but I do not find this sentence very clear.

app/database/db.go
```diff
@@ -314,6 +314,17 @@ func (conn *DB) Count(dest interface{}) *DB {
...
+//     var ages []int64
+//     db.Find(&users).Pluck("age", &ages)
+func (conn *DB) Pluck(column string, value interface{}) *DB {
+	reflectValue := reflect.ValueOf(value).Elem()
+	if reflectValue.Len() > 0 {
+		reflectValue.Set(reflect.MakeSlice(reflectValue.Type(), 0, reflectValue.Cap()))
```
smadbe: What (and why) are you doing here? (a comment in the code may be interesting as well)

app/database/db.go

app/database/db.go

app/database/db.go
```diff
@@ -314,14 +314,17 @@ func (conn *DB) Count(dest interface{}) *DB {
...
 
-// Pluck used to query single column from a model as a map
-//     var ages []int64
-//     db.Find(&users).Pluck("age", &ages)
+// Pluck is used to query a single column into a slice of values
+//     var ids []int64
+//     db.Table("users").Pluck("ID", &ids)
 func (conn *DB) Pluck(column string, value interface{}) *DB {
+	// If 'value' is not empty, wipe its content by replacing it with an empty slice.
+	// Otherwise we would get new values mixed with old values.
```
smadbe: Isn't it the responsibility of the caller? If so, it should panic instead.

zenovich: I think, it is responsibility of the method because it supports passing a non-initiazed slice as a parameter.

smadbe: But you should support a nil value and any invalid value then ;-)

zenovich: Pointers to nil are supported (and were supported), nils are not supported (and were not supported), types that are not pointers to a slice of values are not supported (and were not supported). Nothing’s changed here, we just remove all elements from non-empty slices.

zenovich: I mean, comparing to gorm.

PR #107, opened 2019-04-17
Extract service.checkThatUserHasRightsForDirectRelation() method + fix small issues

zenovich:


app/api/groups/groups.go
```diff
@@ -44,3 +44,35 @@ func checkThatUserOwnsTheGroup(store *database.DataStore, user *database.User, g
...
+func checkThatUserHasRightsForDirectRelation(store *database.DataStore, user *database.User, parentGroupID, childGroupID int64) service.APIError {
+	groupStore := store.Groups()
+
+	var groupData []struct {
+		ID   int64  `gorm:"column:ID"`
+		Type string `gorm:"column:sType"`
+	}
+
+	err := groupStore.OwnedBy(user).
+		WithWriteLock().
```
smadbe: don't you have to check that it is run in a transaction?

app/database/user_test.go
```diff
@@ -12,7 +12,9 @@ import (
 // NewMockUser creates a mock user, to be used for testing
 //func NewMockUser(id, selfGroupID, ownedGroupID, defaultLanguageID int64) *User {
```
smadbe: is this line still relevant?

app/api/groups/groups.go
zenovich: 1. WithWriteLock() panics itself when it is called without a transaction.
2. The method is private (unexported) so I don't think we should have a test for this.

app/database/user_test.go
zenovich: fixed

app/api/groups/groups.go
zenovich: And we have a test that checks WithWriteLock() panics when outside a transaction

PR #109, opened 2019-04-18
Handle incorrect types of destination in DB.Pluck, add more tests

zenovich:


app/database/db.go
```diff
@@ -317,15 +317,22 @@ func (conn *DB) Count(dest interface{}) *DB {
 // Pluck is used to query a single column into a slice of values
 //     var ids []int64
 //     db.Table("users").Pluck("ID", &ids)
-func (conn *DB) Pluck(column string, value interface{}) *DB {
-	// If 'value' is not empty, wipe its content by replacing it with an empty slice.
+// The 'values' parameter should be a pointer to a slice
+func (conn *DB) Pluck(column string, values interface{}) *DB {
```
smadbe: Should not it be called "dest" as some other places? "Dest" looks clearer than "value(s)" for this usage.

zenovich: In other places "dest" has another meanings. Scan() wants it to be *struct or *[]struct, ScanIntoSliceOfMaps accepts *[]map[string]interface{} instead. Pluck() always returns only one column as an array of values. So 'values' is a proper name for it.

zenovich: In gorm it is named as 'value' which, I believe, is worse.

PR #116, opened 2019-04-25
Implement current-user/memberships

zenovich:


app/api/user/user.go
```diff
@@ -18,4 +18,5 @@ func (srv *Service) SetRoutes(router chi.Router) {
 	router.Use(render.SetContentType(render.ContentTypeJSON))
 	router.Use(auth.UserIDMiddleware(&srv.Config.Auth))
 	router.Get("/user/invitations", service.AppHandler(srv.getInvitations).ServeHTTP)
+	router.Get("/user/memberships", service.AppHandler(srv.getMemberships).ServeHTTP)
```
smadbe: As it's not just the memberships of users in general but of the authenticated one, shouldn't we do `/user/me/memberships` (which would allow `/user/{user_id}/memberships` later as well)

zenovich: I meant 'user'='me' otherwise it would be 'users'. I can replace 'user' by 'me' if you will.

zenovich: '/users/{user_id}/' is possible, but '/user/{user_id}' is not

zenovich: I like 'user' better than 'me'

zenovich: Do we really need to have '/users/{user_id}/memberships'?

zenovich: Fixed. It will be /current-user/memberships

PR #122, opened 2019-05-07
Add support for multiple tables in data fixtures, introduce testhelpers.SetupDBWithFixtureString()

zenovich:


zenovich:
It's just more convenient to have an entire fixture (with all tables) in one file or even in a string literal inside a test.

PR #140, opened 2019-06-05
Implement itemAttemptsView service

zenovich:
[itemAttemptsView](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit#heading=h.4ftyl7688840)

### DIFF
--- a/app/database/group_attempt_store.go
+++ b/app/database/group_attempt_store.go
@@ -66,3 +66,32 @@ func (s *GroupAttemptStore) GetAttemptItemIDIfUserHasAccess(attemptID int64, use
 	mustNotBeError(err)
 	return true, itemID, nil
 }
+
+// ByUserAndItemID returns a composable query for getting groups_attempts with the following access rights
+// restrictions:
+// 1) the user should have at least partial access rights to the groups_attempts.idItem item,
+// 2) the user is able to see answers related to his group's attempts, so:
+//   (a) if items.bHasAttempts = 1, then the user should be a member of the groups_attempts.idGroup team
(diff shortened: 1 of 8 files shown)


app/database/group_attempt_store.go
```diff
@@ -66,3 +66,32 @@ func (s *GroupAttemptStore) GetAttemptItemIDIfUserHasAccess(attemptID int64, use
...
 	return true, itemID, nil
 }
+
+// ByUserAndItemID returns a composable query for getting groups_attempts with the following access rights
+// restrictions:
+// 1) the user should have at least partial access rights to the groups_attempts.idItem item,
+// 2) the user is able to see answers related to his group's attempts, so:
+//   (a) if items.bHasAttempts = 1, then the user should be a member of the groups_attempts.idGroup team
+//   (b) if items.bHasAttempts = 0, then groups_attempts.idGroup should be equal to the user's self group
+func (s *GroupAttemptStore) ByUserAndItemID(user *User, itemID int64) *DB {
```
smadbe: I thought first it was "by user and by item" but actually, it is not really "by user", it is more "visible by the user and by item id ". Correct me if I'm wrong. 
If correct, the naming is quite confusing (I expect "by-A-and-B" mainly for something filter exactly on A and B (typically keys)). 
I think something like "VisibleAndByItemID()" would make more sense. I know it is quite longish but I find it less confusing.

zenovich: fixed

PR #145, opened 2019-06-13
Implement the itemEdit service

zenovich:
[itemEdit](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit#heading=h.j38lxuhhu6ni)


app/api/items/update_item.feature
```diff
@@ -0,0 +1,186 @@
...
+      | 50           | 12          |
+      | 50           | 34          |
+    And the table "items_ancestors" should be:
+      | idItemAncestor | idItemChild |
+      | 21             | 60          |
+      | 50             | 12          |
+      | 50             | 34          |
+    And the table "groups_items" should stay unchanged
+
+  Scenario: Valid with empty full_screen
```
smadbe: I don't get in what it is an interesting scenario. Probably there is but I don't see it.

app/api/items/update_item.feature
```diff
@@ -0,0 +1,186 @@
...
+    """
+    {
+    }
+    """
+    Then the response should be "updated"
+    And the table "items" at ID "50" should be:
+      | ID | sType  | sUrl | idDefaultLanguage | bTeamsEditable | bNoScore | sTextID | bTitleBarVisible | bCustomChapter | bDisplayDetailsInParent | bUsesAPI | bReadOnly | sFullScreen | bShowDifficulty | bShowSource | bHintsAllowed | bFixedRanks | sValidationType | iValidationMin | idItemUnlocked | iScoreMinUnlock | sTeamMode | bTeamsEditable | idTeamInGroup | iTeamMaxMembers | bHasAttempts | sAccessOpenDate      | sDuration | sEndContestDate      | bShowUserInfos | sContestPhase | iLevel | bNoScore | groupCodeEnter |
+      | 50 | Root   | null | 1                 | 0              | 0        | null    | 1                | 0              | 0                       | 1        | 0         | default     | 0               | 0           | 0             | 0           | All             | null           | null           | 100             | null      | 0              | null          | 0               | 0            | null                 | null      | null                 | 0              | Running       | null   | 0        | 0              |
```
smadbe: I cannot see the type before (see above remark) so I do know it is valid, but anyway, shouldn't all tables just be "unchanged" for this scenario?

app/api/items/update_item.feature
```diff
@@ -0,0 +1,186 @@
...
+
+Background:
+  Given the database has the following table 'users':
+    | ID | sLogin | tempUser | idGroupSelf | idGroupOwned |
+    | 1  | jdoe   | 0        | 11          | 12           |
+  And the database has the following table 'groups':
+    | ID | sName      | sType     |
+    | 11 | jdoe       | UserSelf  |
+    | 12 | jdoe-admin | UserAdmin |
+  And the database has the following table 'items':
```
smadbe: For a service updating the attributes and being able to show that some have changed, some haven't, I thing it would make sense to put value in the column... at least in some of them.

app/api/items/update_item.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/go-chi/render"
+	"gopkg.in/go-playground/validator.v9"
+
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/formdata"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+type itemWithDefaultLanguageIDAndOptionalType struct {
+	Item              item   `json:"item,squash"`
```
smadbe: Same remark about override/extension instead

app/api/items/update_item.feature
```diff
@@ -0,0 +1,186 @@
...
+    | 50 | Course | null | 1                 | 0              | 0        | null    | 1                | 0              | 0                       | 1        | 0         | default     | 0               | 0           | 0             | 0           | All             | null           | null           | 100             | null      | 0              | null          | 0               | 0            | null                 | null      | null                 | 0              | Running       | null   | 0        | 0              |
+  And the table "items_strings" should stay unchanged
+  And the table "items_items" should be:
+    | idItemParent | idItemChild |
+    | 21           | 60          |
```
smadbe: If you do not specify the children, they should not be removed. not specified <> []  !!!

app/api/items/add_item.go
```diff
@@ -57,11 +56,16 @@ type item struct {
 	GroupCodeEnter  *bool      `json:"group_code_enter" sql:"column:groupCodeEnter"`
 }
 
+type itemWithRequiredType struct {
```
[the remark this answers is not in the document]
zenovich: No, sorry. What you're suggesting will not work since we ignore non-exported fields and fields without the 'json' tag. Note that 'Type' is not twice in the struct (I removed it from the 'item' struct)

app/api/items/update_item.feature
zenovich: It is interesting because the sFullScreen column is defined as
```
`sFullScreen` enum('forceYes','forceNo','default','') NOT NULL DEFAULT 'default'
```
So an empty string should be valid.

app/api/items/update_item.feature
zenovich: fixed

app/api/items/update_item.feature
zenovich: fixed

app/api/items/update_item.go
zenovich: (see my reply above)

app/api/items/update_item.feature
zenovich: fixed

zenovich:
fixed

PR #149, opened 2019-06-14
groupEdit service in OpenAPI

smadbe:
Document the first service "groupEdit" in OpenAPI.
On master, it will deploy the spec file so that it can directly be integrated into the dev doc.


app/api/groups/update_group.go
```diff
@@ -1,5 +1,26 @@
...
+// swagger:operation PUT /groups/{group_id} groupEdit
+// ---
+// summary: Edit group information
+// description: Edit group information.
+//   Requires the user to be the owner of the group.
+// parameters:
+// - name: group_id
+//   in: path
+//   required: true
+//   type: string
```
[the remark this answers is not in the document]
smadbe: Changed to "integer". Actually would be probably possible by updating the commented format by struct code, but I prefer that way for the path parameters.

app/api/groups/update_group.go
```diff
@@ -13,18 +34,28 @@ import (
...
+	FreeAccess bool `json:"free_access" sql:"column:bFreeAccess"`
+	// Duration after the first use of the password it will expire
+	// Nullable
+	// pattern: \d{2}:\d{2}:\d{2}
+	// example: 79:56:22
+	PasswordTimer *string `json:"password_timer" sql:"column:sPasswordTimer" validate:"omitempty,duration" x-nullable:"true" nullable:"true"`
+	// Nullable
+	PasswordEnd *time.Time `json:"password_end" sql:"column:sPasswordEnd"`
+	// Nullable
+	RedirectPath *string `json:"redirect_path" sql:"column:sRedirectPath" validate:"omitempty,redirect_path"`
```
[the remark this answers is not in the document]
smadbe: I don't think that's possible.

As stated in the other PR on OpenAPI, there are many repetitions, but here the goal is to do a bit better than fully writing the doc by hand.

app/api/groups/update_group.go
```diff
@@ -13,18 +34,28 @@ import (
...
+	Grade int32  `json:"grade" sql:"column:iGrade"`
+	// Nullable
+	Description *string `json:"description" sql:"column:sDescription"`
+	Opened      bool    `json:"opened" sql:"column:bOpened"`
+	// If changed from true to false, automatically switch all requests to join this group from requestSent to requestRefused
+	FreeAccess bool `json:"free_access" sql:"column:bFreeAccess"`
+	// Duration after the first use of the password it will expire
+	// Nullable
+	// pattern: \d{2}:\d{2}:\d{2}
+	// example: 79:56:22
```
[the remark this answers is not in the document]
smadbe: Fixed. 
Btw, this "nullable"  is supported by OpenAPI but I cannot make go-swagger to put the right tags into the spec (I still have open questions on their slack)

app/api/groups/update_group.go
```diff
@@ -48,12 +48,13 @@ type groupUpdateInput struct {
 	FreeAccess bool `json:"free_access" sql:"column:bFreeAccess"`
 	// Duration after the first use of the password it will expire
 	// Nullable
-	// pattern: \d{2}:\d{2}:\d{2}
+	// pattern: ^\d{2}:\d{2}:\d{2}$
```
zenovich: This pattern is not really correct. Values like 1:2:3 should be considered as correct, while values having more than 59 minutes/seconds should not (see https://github.com/France-ioi/AlgoreaBackend/blob/master/app/formdata/validations.go#L11). It is one more reason to add it as a custom type somehow.

smadbe: Actually, it is a doc, that's not very important (it is should show the format, a bit as an example would do). I improved the pattern to avoid rejecting valid cases. 
Yes, of course, custom types would be nice, but if it is not possible in the current go-swagger implem, that does not worth spending hours implementing this.

app/api/groups/update_group.go
```diff
@@ -17,9 +17,15 @@ package groups
 //     "$ref": "#/definitions/groupUpdateInput"
 // responses:
 //   "200":
-//     "$ref": "#/responses/updated"
-//   default:
-//     "$ref": "#/responses/error"
+//     "$ref": "#/responses/updatedResponse"
+//   "400":
+//     "$ref": "#/responses/badRequestPOSTPUTPATCHResponse"
+//   "401":
```
zenovich: Not sure if we ever return 401 somewhere currently

smadbe: I think we do in the auth middleware

PR #152, opened 2019-06-18
Implement userNotificationReadDateUpdate service

zenovich:
[userNotificationReadDateUpdate](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit#heading=h.de3aphmygrrv)


app/api/currentuser/update_notification_read_date.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/jinzhu/gorm"
+
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+func (srv *Service) updateNotificationReadDate(w http.ResponseWriter, r *http.Request) service.APIError {
+	user := srv.GetUser(r)
+	err := user.Load()
+	if err == database.ErrUserNotFound {
```
smadbe: I would not do `InsufficientAccessRightsError` as it is not really a problem of "forbidden" and normally if his token was accepted (in the middleware), it exists.  
Actually, you could do it by checking the row affected by the update instead (no extra query).
But I think I would let the update flow silently in any case. Tell me if you disagree.
(if you do so, don't forget to remove "403" from the possible responses in the doc.)

zenovich: fixed

PR #154, opened 2019-06-19
Add swagger specs for itemAnswersView, itemGetAnswerToken, itemAnswerUpdateCurrent, itemAnswerGet, acceptGroupInvitation; add swagger tags

zenovich:


app/api/answers/get.go
```diff
@@ -36,3 +64,39 @@ func (srv *Service) get(rw http.ResponseWriter, httpReq *http.Request) service.A
 	render.Respond(rw, httpReq, convertedResult)
 	return service.NoError
 }
+
+// OK. Normal response of the itemAnswersGet service
+// swagger:response itemAnswerGetResponse
+type itemAnswerGetResponse struct { //nolint:unused,deadcode
```
smadbe: Why doesn't any linter complain about the fact the comment does not start with "itemAnswerGetResponse"? 
When unused, an alternative (better or not?) would be to directly put it in the service comment directly (not through refs).  Cfr what I have written in the comments on top of `doc/errors.go`.
Strange to have an extra import for something unused... 

(open question, may require to discuss it)

app/api/answers/get.go
```diff
@@ -2,13 +2,41 @@ package answers
...
+// parameters:
+// - name: answer_id
+//   in: path
+//   type: integer
+//   required: true
+// responses:
+//   "200":
+//     "$ref": "#/responses/itemAnswerGetResponse"
+//   "400":
+//     "$ref": "#/responses/badRequestPOSTPUTPATCHResponse"
```
smadbe: Cannot be used the GET, the message is not appropriate I think. Or at least its name ;-)

app/api/answers/get_answers.go
```diff
@@ -68,23 +104,43 @@ type rawAnswersData struct {
...
+	// enum: Submission,Saved,Current
+	Type string `json:"type"`
+	// Nullable
+	LangProg *string `json:"lang_prog,omitempty"`
+	// required: true
+	SubmissionDate string `json:"submission_date"`
+	// Nullable
+	Score *float32 `json:"score,omitempty"`
+	// Nullable
+	Validated *bool `json:"validated,omitempty"`
```
smadbe: Why is doing already now with this "omitempty"? I don't get it. Should not be omitted.

app/api/answers/get_answers.go
```diff
@@ -68,23 +104,43 @@ type rawAnswersData struct {
...
+
+	// required: true
 	User answersResponseAnswerUser `json:"user"`
 }
 
+// OK. Normal response of the itemAnswersView service
+// swagger:response itemAnswersViewResponse
+type itemAnswersViewResponse struct { // nolint:unused,deadcode
```
smadbe: You can put directly this construct in the service spec to avoid polluting the code with this. Cfr what I have written in the comments on top of `doc/errors.go`.

app/api/answers/submit.go
```diff
@@ -12,10 +12,43 @@ import (
...
 )
 
+// swagger:operation POST /answers answers items itemGetAnswerToken
+// ---
+// summary: Generate answer token
+// description: Generate a token that can be used by the TaskGrader to ensure task parameters have not been altered.
+//   A task_token has to be given to this service.
+//
+//
+//   This service corresponds to task.php/askValidation <https://github.com/France-ioi/AlgoreaPlatform/blob/master/task/task.php#L202>
```
smadbe: Is this comment future-proof enough? Wasn't this link provided to help you implement it (but so irrelevant in a doc). Should we link to dev doc?

app/api/answers/submit.go
```diff
@@ -148,3 +185,15 @@ func (requestData *SubmitRequest) Bind(r *http.Request) error {
 var (
 	_ render.Binder = (*SubmitRequest)(nil)
 )
+
+// Created. Normal response of the itemGetAnswerToken service
```
smadbe: Don't get "Normal response"

app/doc/responses.go
```diff
@@ -16,3 +16,16 @@ type updatedResponse struct {
 		Success bool `json:"success"`
 	}
 }
+
+// CreatedResponse returned when the request has successfully created the object
```
smadbe: The comment here is intended to be displayed in the doc, so it should the same format as the other. So "The request has successfully created the object". There is no lint check in the doc package anyway.

app/doc/responses.go
```diff
@@ -16,3 +16,16 @@ type updatedResponse struct {
...
+// swagger:response updatedResponse
+type CreatedResponse struct {
+	// description: The request has successfully created the object
```
smadbe: line 23 to be removed

app/doc/responses.go
```diff
@@ -18,7 +18,6 @@ type updatedResponse struct {
 }
 
 // CreatedResponse returned when the request has successfully created the object
-// swagger:response updatedResponse
```
smadbe: if it could be used alone, you can leave the `// swagger:response updatedResponse`

zenovich: If I leave it (swagger:response CreatedResponse), the validator warns (because it is unused)

app/doc/responses.go
zenovich: fixed

app/doc/responses.go
zenovich: Actually it is never displayed anywhere, so I removed it.

app/api/answers/submit.go
zenovich: fixed

app/api/answers/submit.go
zenovich: fixed

app/api/answers/get_answers.go
zenovich: discussed

smadbe: I didn't see it was a list, sorry.

app/api/answers/get_answers.go
zenovich: fixed

app/api/answers/get.go
```diff
@@ -36,3 +64,39 @@ func (srv *Service) get(rw http.ResponseWriter, httpReq *http.Request) service.A
...
+		// required:true
+		ItemID int64 `json:"item_id,string"`
+		// Nullable
+		// format:integer
+		AttemptID *string `json:"attempt_id,string"`
+		// required:true
+		// enum:Submission,Saved,Current
+		Type string `json:"type"`
+		// Nullable
+		State *string `json:"state,omitempty"`
```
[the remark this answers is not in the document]
zenovich: fixed

app/api/items/get_navigation_data.feature
```diff
@@ -290,7 +293,8 @@ Feature: Get item for tree navigation
                   "partial_access": true,
                   "full_access": true,
                   "gray_access": true
-                }
+                },
+                "children": null
```
smadbe: is there a why this "children" is null while the prev one is empty array? (I've checked into details yet)

app/api/answers/get.go
zenovich: > Why doesn't any linter complain about the fact the comment does not start with "itemAnswerGetResponse"?

Probably it is because itemAnswerGetResponse is unexported

zenovich: moved itemAnswerGetResponse into app/doc/responses.go to eliminate extra imports for the unused code

app/api/items/get_navigation_data.feature
zenovich: discussed in Slack

smadbe:
dle [9:43 AM]
for the `// OK. Success response of the ABC service`. Don't you think it may be useful to describe the actual response content? Such as "an answer tuple" or "an answer token "
Dmitry Zenovich [9:46 AM]
Yeah, probably it will be helpful

zenovich:
fixed

PR #157, opened 2019-06-26
DB optimizations

zenovich:


app/database/ancestors.go
```diff
@@ -95,58 +93,89 @@ func (s *DataStore) createNewAncestors(objectName, upObjectName string) { /* #no
 	}
 
 	// For every object marked as 'processing', we compute all its ancestors
-	query = `
-		INSERT IGNORE INTO ` + objectName + `_ancestors
+	insertQueries := make([]string, 0, 4)
+	insertQueries = append(insertQueries, `
```
golangcibot: G202: SQL string concatenation (from `gosec`)

app/database/ancestors.go
```diff
@@ -95,58 +93,89 @@ func (s *DataStore) createNewAncestors(objectName, upObjectName string) { /* #no
...
+			`+relationsTable+`.id`+upObjectName+`Child
+			`+bIsSelfValue+`
+		FROM `+relationsTable+`
+		JOIN `+objectName+`_propagate
+		ON (
+			`+relationsTable+`.id`+upObjectName+`Child = `+objectName+`_propagate.ID
+		)
+		WHERE
+			`+objectName+`_propagate.sAncestorsComputationState = 'processing'`+groupsAcceptedCondition)
+	insertQueries = append(insertQueries, `
```
golangcibot: G202: SQL string concatenation (from `gosec`)

app/database/ancestors.go
```diff
@@ -95,58 +93,89 @@ func (s *DataStore) createNewAncestors(objectName, upObjectName string) { /* #no
...
 	if objectName == groups {
-		query += `
+		insertQueries[2] += `
 			AND (
-				groups_groups_join.sType = 'invitationAccepted' OR
-				groups_groups_join.sType = 'requestAccepted' OR
-				groups_groups_join.sType = 'direct'
+				groups_groups_join.sType IN('invitationAccepted', 'requestAccepted', 'direct')
+			)`
+		insertQueries = append(insertQueries, `
```
golangcibot: G202: SQL string concatenation (from `gosec`)

db/migrations/2_create_index_on_history_groups_attempts_id.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+ALTER TABLE `history_groups_attempts` ADD INDEX `ID` (`ID`);
+
+-- +migrate Down
+ALTER TABLE `history_groups_attempts` DROP INDEX `ID`;
```
smadbe: I think all these tables can disappear, so if it causes you perf pb, I can start to see if we can do it now.

cmd/serve.go
```diff
@@ -11,17 +15,37 @@ import (
...
-		Use:   "serve",
+		Use:   "serve [environment]",
 		Short: "start http server",
+		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			var err error
 
 			var application *app.Application
-			application, err = app.New()
+			environment := "dev"
```
golangcibot: string `dev` has 2 occurrences, make it a constant (from `goconst`)

app/logging/logger.go
```diff
@@ -79,6 +81,5 @@ func ResetShared() {
 
 func new() *Logger {
 	logger := logrus.New()
-	log.SetOutput(logger.Writer())
```
smadbe: I am pretty sure there was a reason why it was there actually.

Ok, you do this, then the two  remaining lines should be merged.

zenovich: fixed

Makefile
```diff
@@ -40,9 +40,13 @@ gen-keys:
 	openssl genpkey -algorithm RSA -out private_key.pem 2>/dev/null | openssl genrsa -out private_key.pem 1024
 	openssl rsa -pubout -in private_key.pem -out public_key.pem
 db-restore: $(BIN_PATH)
-	$(BIN_PATH) db-restore
+	$(BIN_PATH) db-restore test
 db-migrate: $(BIN_PATH)
-	$(BIN_PATH) db-migrate
+	$(BIN_PATH) db-migrate test
```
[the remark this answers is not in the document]
zenovich: fixed

cmd/common.go
```diff
@@ -0,0 +1,14 @@
+package cmd
+
+import "os"
+
+func resolveEnvironment(args []string, defaultEnvironment string) string {
```
smadbe: At quite short term, I think this will not be the right place for this function as it may be required by the config, app, ... Probably it will have to be in a "common" package or a "utils" (if not dependant from "ALGOREA"). 
But anyway, the move may be done later.

cmd/common.go
```diff
@@ -0,0 +1,14 @@
...
+	environment := defaultEnvironment
```
smadbe: I think the defaultEnvironment should be a constant for the project, I don't see why it should be specific to the caller (meaning different part of the app may do different env assumptions at the same time).

And one of the goal of using env variable, it is not have to pass them by arg with all the "complex" machinery of passing them everywhere. If the visibility of env var is problematic through the Makefile, you can call it directly with the makefile  `ALGOREA_ENV=test make something` or `make something ALGOREA_ENV=test`

So I would remove most of the code of this commit (sorry 😕 but it is just a few lines 🙂). But maybe I would add a check that it is among "prod", "test, "dev", and otherwise default to "dev" (or other default if you prefer). Then you could add shortcut function "inProdEnv() bool", "inTestEnv()", ...

app/app.go
```diff
@@ -27,40 +28,42 @@ type Application struct {
 func New(environment string) (*Application, error) {
 	var err error
 
+	log.Println("Starting application: environment =", environment)
```
smadbe: As said in the previous review, I would not pollute all functions everywhere with this "environment" variable while it is already available for all.

smadbe:
@zenovich I did the change for the env stuff. I still get a very weird behavior in the makefile with dot env var containing the db data on circle-ci. So for now the "make test" does not force the env to "test". :-/

cmd/db_migrate.go
```diff
@@ -9,21 +9,27 @@ import (
...
+		Use:   "db-migrate",
 		Short: "apply schema-change migrations to the database",
-		Args:  cobra.MaximumNArgs(1),
 		Long:  `migrate uses go-pg migration tool under the hood supporting the same commands and an additional reset command`,
 		Run: func(cmd *cobra.Command, args []string) {
 			var err error
 
+			common.SetDefaultEnvToTest()
+			if !common.IsEnvTest() {
+				fmt.Println("'db-migrate' can only be run in 'test' env")
```
zenovich: Why so?

cmd/db_restore.go
```diff
@@ -9,21 +9,27 @@ import (
...
-		Use:   "db-restore [environment]",
+		Use:   "db-restore",
 		Short: "load the last db schema",
-		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			var err error
 
+			common.SetDefaultEnvToTest()
+			if !common.IsEnvTest() {
+				fmt.Println("'db-migrate' can only be run in 'test' env")
```
zenovich: Why so and why does `db-restore` mention `db-migrate`?

cmd/db_migrate.go
```diff
@@ -9,21 +9,27 @@ import (
...
 	"github.com/spf13/cobra"
 
+	"github.com/France-ioi/AlgoreaBackend/app/common"
 	"github.com/France-ioi/AlgoreaBackend/app/config"
 )
 
 func init() { // nolint:gochecknoinits
 
 	var migrateCmd = &cobra.Command{
-		Use:   "db-migrate [environment]",
```
zenovich: `./bin/AlgoreaBackend db-migrate dev` is much easier to type than `./bin/AlgoreaBackend db-migrate ALGOREA_ENV=dev`. Can we keep the command line arguments?

app/common/environment.go
```diff
@@ -0,0 +1,45 @@
+package common
+
+import "os"
+
+const envVarName = "ALGOREA_ENV"
+const productionEnv = "prod"
+const developementEnv = "dev"
+const testEnv = "test"
```
zenovich: Why do you think that only these three environments are possible? I would like to be able to create separate environments for task, for example. I can have 'dev_with_many_users' + 'dev_many_teams' (each uses its own DB) which are both kind of 'dev'. So, I think we should not constrain the list of possible environments. The only "special" environment that the code should know is "test". Also, I think we don't need the IsEnvDev() function.

smadbe: IsEnvDev is for your `if environment == devEnvironment {`

zenovich: But will the IsEnvDev() return true for 'dev_with_many_users'? I think we should add a config variable 'stub_auth' for this and use it instead of `if environment == devEnvironment`

smadbe: This is what you had done. Anyway, it will change in a few days with the new auth...

PR #161, opened 2019-07-02
Implement groupTeamProgress service

zenovich:
a half of [groupTeamUserProgress](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit#heading=h.c7xpcaxby5ym)

### DIFF
--- a/app/api/groups/get_team_progress.go
+++ b/app/api/groups/get_team_progress.go
@@ -22,6 +22,7 @@ import (
 //   required: true
 // - name: parent_item_ids
 //   in: query
+//   required: true
 //   type: array
 //   items:
 //     type: integer
--- a/app/api/groups/get_team_progress.go
+++ b/app/api/groups/get_team_progress.go
@@ -1,0 +1,99 @@
+package groups
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/go-chi/render"
+	"github.com/jinzhu/gorm"
+
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+func (srv *Service) getTeamProgress(w http.ResponseWriter, r *http.Request) service.APIError {
+	user := srv.GetUser(r)
+
+	groupID, err := service.ResolveURLQueryPathInt64Field(r, "group_id")
+	if err != nil {
+		return service.ErrInvalidRequest(err)
+	}
+
+	if apiError := checkThatUserOwnsTheGroup(srv.Store, user, groupID); apiError != service.NoError {
+		return apiError
+	}
+
+	itemParentIDs, err := service.ResolveURLQueryGetInt64SliceField(r, "parent_item_ids")
+	if err != nil {
+		return service.ErrInvalidRequest(err)
+	}
+
+	itemsVisibleToUserSubQuery := srv.Store.GroupItems().AccessRightsForItemsVisibleToUser(user).SubQuery()
+
+	// Preselect IDs of end member for that we will calculate the stats.
+	// There should not be too many of end members on one page.
+	var teamIDs []interface{}
+	teamIDQuery := srv.Store.GroupAncestors().
+		Joins("JOIN groups ON groups.ID = groups_ancestors.idGroupChild AND groups.sType = 'Team'").
+		Where("groups_ancestors.idGroupAncestor = ?", groupID).
+		Where("groups_ancestors.idGroupChild != groups_ancestors.idGroupAncestor")
+	teamIDQuery, apiError := service.ApplySortingAndPaging(r, teamIDQuery, map[string]*service.FieldSortingParams{
+		// Note that we require the 'from.name' request parameter although the service does not return group names
+		"name": {ColumnName: "groups.sName", FieldType: "string"},
+		"id":   {ColumnName: "groups.ID", FieldType: "int64"},
+	}, "name,id")
+	if apiError != service.NoError {
+		return apiError
+	}
+	teamIDQuery = service.NewQueryLimiter().Apply(r, teamIDQuery)
+	service.MustNotBeError(teamIDQuery.
+		Pluck("groups.ID", &teamIDs).Error())
+
+	if len(teamIDs) == 0 {
+		render.Respond(w, r, []map[string]interface{}{})
+		return service.NoError
+	}
+
+	itemsQuery := srv.Store.ItemItems().
+		Select("items_items.idItemChild").
+		Where("idItemParent IN (?)", itemParentIDs).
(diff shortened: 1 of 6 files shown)


app/api/groups/get_team_progress.go
```diff
@@ -0,0 +1,99 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/go-chi/render"
+	"github.com/jinzhu/gorm"
+
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+func (srv *Service) getTeamProgress(w http.ResponseWriter, r *http.Request) service.APIError {
```
smadbe: Doc/spec missing. You should probably do it while writing the code so that you can add every point you are thinking about when implementing the service.

app/api/groups/get_team_progress.go
```diff
@@ -0,0 +1,99 @@
...
+		Select(`
+			items.ID AS idItem,
+			groups.ID AS idGroup,
+			IFNULL(attempt_with_best_score.iScore, 0) AS iScore,
+			IFNULL(attempt_with_best_score.bValidated, 0) AS bValidated,
+			(SELECT MAX(sLastActivityDate) FROM groups_attempts WHERE idGroup = groups.ID AND idItem = items.ID) AS sLastActivityDate,
+			IFNULL(attempt_with_best_score.nbHintsCached, 0) AS nbHintsRequested,
+			IFNULL(attempt_with_best_score.nbSubmissionsAttempts, 0) AS nbSubmissionAttempts,
+			IF(attempt_with_best_score.idGroup IS NULL,
+				0,
```
smadbe: For instance, for the doc/spec, probably interesting to mention that if not started the timeSpend is considered "0"

app/api/groups/get_team_progress.go
```diff
@@ -0,0 +1,99 @@
...
+				(
+					SELECT IF(MAX(bValidated),
```
smadbe: You already have this value (`MAX(bValidated)`) from `attempt_with_best_score.bValidated`. Wouldn't be more efficient to use it so that the subquery does not recompute it again for each row?
And BTW... it looks more efficient (but you never know ;-)) to JOIN "first_attempt" in the same way as "attempt_with_best_score" and then use these directly without any need for the subquery with aggregation for each row (looks expensive).

zenovich: Fixed the first part. For the second part, JOIN will not help here (and we don't need to optimize the query since it takes just about 0.5s to complete on the large data set).

app/api/groups/get_team_progress.go
```diff
@@ -10,6 +10,48 @@ import (
...
+// ---
+// summary: Display the current progress of teams on a subset of items
+// description: For all children of items from the parent_item_id list,
+//              display the result of each team among the descendants of the group.
+// parameters:
+// - name: group_id
+//   in: path
+//   type: integer
+//   required: true
+// - name: parent_item_ids
```
smadbe: I suppose it is required as well

app/doc/responses.go
```diff
@@ -121,3 +121,44 @@ type groupsGetGroupProgressResponse struct {
...
+	Body []struct {
+		// The team’s `group_id`
+		// required:true
+		GroupID int64 `json:"group_id,string"`
+		// required:true
+		ItemID int64 `json:"item_id,string"`
+		// The best score across all attempts, or 0
+		// required:true
+		Score float32 `json:"score"`
+		// Max of `bValidated` across all attempts, or `false`
```
smadbe: -> "Whether the team has validated the item"

app/doc/responses.go
```diff
@@ -121,3 +121,44 @@ type groupsGetGroupProgressResponse struct {
...
+// swagger:response groupsGetTeamProgressResponse
+type groupsGetTeamProgressResponse struct {
+	// in: body
+	Body []struct {
+		// The team’s `group_id`
+		// required:true
+		GroupID int64 `json:"group_id,string"`
+		// required:true
+		ItemID int64 `json:"item_id,string"`
+		// The best score across all attempts, or 0
```
smadbe: Was more a way to explain to the implementer.
We can say "Current score.  If there are no attempts, the score is 0."

app/doc/responses.go
```diff
@@ -121,3 +121,44 @@ type groupsGetGroupProgressResponse struct {
...
+		// required:true
+		Score float32 `json:"score"`
+		// Max of `bValidated` across all attempts, or `false`
+		// required:true
+		Validated bool `json:"validated"`
+		// Max across all attempts of the team (`null` if no attempts)
```
smadbe: I think the description can be removed. And you have to add `Nullable`

app/doc/responses.go
```diff
@@ -121,3 +121,44 @@ type groupsGetGroupProgressResponse struct {
...
+		// required:true
+		LastActivityDate time.Time `json:"last_activity_date"`
+		// `groups_attempts.nbHintsCached` of the attempt with the best score (0 if no attempts;
+		// if several with the same score, we use the first attempt chronologically on `sBestAnswerDate`)
```
smadbe: --> "Number of hints requested for the attempt with the best score (if multiple, take the first one, chronologically). If there are no attempts, the number of hints is 0."

app/doc/responses.go
```diff
@@ -121,3 +121,44 @@ type groupsGetGroupProgressResponse struct {
...
+		// required:true
+		HintsRequested int32 `json:"hints_requested"`
+		// `groups_attempts.nbSubmissionsAttempts` of the attempt with the best score (0 if no attempts;
```
smadbe: "Number of submissions for the attempt with the best score (if multiple, take the first one, chronologically). If there are no attempts, the number of submissions is 0."

app/api/groups/get_team_progress.go
zenovich: fixed

app/api/groups/get_team_progress.go
zenovich: fixed

app/api/groups/get_team_progress.go
zenovich: fixed

app/doc/responses.go
zenovich: fixed

app/doc/responses.go
zenovich: fixed

app/doc/responses.go
zenovich: fixed

app/doc/responses.go
zenovich: fixed

PR #164, opened 2019-07-05
Fix cucumber tests instability caused by db concurrency issues

zenovich:
Can be tested with
```
for i in `seq 1 1000`; do godog --stop-on-failure ./app/api/groups/reject_requests.feature; if [[ "$?" != '0' ]]; then break; fi; done
```


zenovich:
This PR doesn't really fix the issue. The real fix is in #166.

PR #176, opened 2019-07-12
Rework UserIDMiddleware to use sessions stored in the DB and the 'Authorization' header

zenovich:
[Token-based authentication](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit#heading=h.znnhk55p9ro)


app/auth/middleware_test.go
```diff
@@ -1,43 +1,138 @@
...
+			expectedBody:             "No access token provided",
+		},
+		{
+			name:                     "token is too long (should not query the DB)",
+			authHeader:               "Bearer " + strings.Repeat("1", 256),
+			expectedStatusCode:       401,
+			expectedServiceWasCalled: false,
+			expectedBody:             "The access token has expired",
+		},
+	}
```
smadbe: From what I understand, you loop through all values of the authorization header. I do not see that tested here.

app/auth/middleware.go
```diff
@@ -62,3 +63,20 @@ func UserIDMiddleware(sessionStore *database.SessionStore) func(next http.Handle
 		})
 	}
 }
+
+// SetAuthorizationHeaderFromQueryMiddleware is a middleware that copies an access token
```
smadbe: If you use appropriate development tools, you can set the header correctly, I don't like the idea of using completely different stuffs in dev, I have always put the Authorization headers myself as a developer. I don't think it push toward good dev practices, and we should as the other devs are not really used to.

app/auth/middleware.go
```diff
@@ -14,55 +17,47 @@ const (
...
-			}
-			if auth.UserID <= 0 {
-				http.Error(w, "Invalid response from auth server. No error by userID:"+string(auth.UserID), http.StatusBadGateway)
+			if !authorized {
+				http.Error(w, "The access token has expired", http.StatusUnauthorized)
 				return
 			}
 
-			ctx := context.WithValue(r.Context(), ctxUserID, auth.UserID)
+			ctx := context.WithValue(r.Context(), ctxUserID, userID)
```
smadbe: Shouldn't we join with the user directly and store the user directly? 
Now we do a request to get some user info, we could fetch the user data. It would allow to get rid of the lazy loading.

app/auth/middleware.go
```diff
@@ -14,55 +17,47 @@ const (
...
-				http.Error(w, err.Error(), http.StatusBadGateway)
-				return
+			if len(accessToken) < 255 {
+				err := sessionStore.Where("sAccessToken = ?", accessToken).
+					Where("sExpirationDate > NOW()").PluckFirst("idUser", &userID).
+					Error()
+				authorized = err == nil
+				if err != nil && !gorm.IsRecordNotFoundError(err) {
+					logging.Errorf("Can't validate an access token: %s", err)
+					http.Error(w, "Can't validate the access token", http.StatusBadGateway)
```
smadbe: All these errors (401 or 500) should return a proper JSON response in the same format as the other. From what I understand here, it is not the case. 

BTW, could we do a *feature test with 401? I am a bit missing it here :-)

app/auth/middleware.go
```diff
@@ -14,55 +17,47 @@ const (
...
-				return
-			}
-			if auth.Error != "" {
-				http.Error(w, "Unable to validate the session ID: "+auth.Error, http.StatusUnauthorized)
-				return
-			}
-			if auth.UserID <= 0 {
-				http.Error(w, "Invalid response from auth server. No error by userID:"+string(auth.UserID), http.StatusBadGateway)
+			if !authorized {
+				http.Error(w, "The access token has expired", http.StatusUnauthorized)
```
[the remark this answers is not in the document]
zenovich: fixed

app/auth/middleware_test.go
zenovich: added the test

app/auth/middleware.go
zenovich: reverted this commit

app/auth/middleware.go
zenovich: done

app/database/user_answer_store_integration_test.go
```diff
@@ -332,7 +332,8 @@ func TestUserAnswerStore_Visible(t *testing.T) {
 				test.fixture)
 			defer func() { _ = db.Close() }()
 			store := database.NewDataStore(db)
-			user := database.NewUser(test.userID, store.Users(), nil)
+			user := &database.User{}
+			assert.NoError(t, user.LoadByID(store, test.userID))
```
golangcibot: user.LoadByID undefined (type *database.User has no field or method LoadByID) (from `typecheck`)

app/auth/middleware.go
zenovich: fixed

app/api/currentuser/get_info.go
```diff
@@ -52,7 +52,7 @@ func (srv *Service) getInfo(w http.ResponseWriter, r *http.Request) service.APIE
 		AllowSubgroups    *bool   `gorm:"column:allowSubgroups" json:"allow_subgroups"`
 	}
 
-	err := srv.Store.Users().ByID(user.UserID).
+	err := srv.Store.Users().ByID(user.ID).
```
smadbe: The test coverage for this file I think. And probably it is because your "user not existing" scenarios are now covered by the middleware. So probably `err == gorm.ErrRecordNotFound` is irrelevant now. But you should add a short comment explaining the `gorm.ErrRecordNotFound` is irrelevant in the case as are sure that the user exists because of the middleware.
... hum actually that's not true as, because of race condition, the user may have been deleted in between. But you should still have to add a comment saying it is very unlikely... and still find a way to test it :-)

zenovich:
```	if err == gorm.ErrRecordNotFound {
		return service.InsufficientAccessRightsError
	}
	service.MustNotBeError(err)```

PR #180, opened 2019-07-24
Improve mocking of DB time

zenovich:


app/database/db.go
```diff
@@ -411,6 +411,23 @@ func (conn *DB) Exec(sqlQuery string, values ...interface{}) *DB {
 	return newDB(conn.db.Exec(sqlQuery, values...))
 }
 
+var nowExpr = gorm.Expr("NOW()")
```
smadbe: The `db.go` file is already quite long, shouldn't you move it to something like `db_now.go` or `db_functions.go`?

PR #185, opened 2019-07-28
Introduce a separate config setting for enabling logging of low-level DB calls (+ update README.md)

zenovich:


app/logging/db_logger.go
```diff
@@ -10,18 +10,19 @@ type DBLogger interface {
 }
 
 // NewDBLogger returns a logger for the database and the `logmode`, according to the config
```
smadbe: Probably needs to be updated as well.

zenovich: fixed

PR #201, opened 2019-08-14
Add swagger specs for groupsMembershipHistory

zenovich:


app/api/currentuser/get_group_memberships_history.go
```diff
@@ -8,6 +8,51 @@ import (
 	"github.com/France-ioi/AlgoreaBackend/app/service"
 )
 
+// swagger:operation GET /current-user/group-memberships-history groups users groupsMembershipHistory
+// ---
+// summary: Display a history of invitations/requests for the current user
```
smadbe: Display -> return

app/doc/models.go
```diff
@@ -15,3 +17,26 @@ type userCreateTmpResponse struct {
...
+	// `groups_groups.sStatusDate`
+	// required: true
+	StatusDate time.Time `json:"status_date"`
+	// `groups_groups.sType`
+	// required: true
+	// enum: invitationSent,requestSent,invitationAccepted,requestAccepted,invitationRefused,requestRefused,removed,left
+	Type string `json:"type"`
+
+	// required: true
+	Group struct {
```
smadbe: It does not seems to be really reusage in other context (as there are only name and type)...
Actually, where is it used?

app/api/currentuser/get_group_memberships_history.go
smadbe: Or just "Get a history ..." as display is already displayed at the next line.

zenovich: fixed

app/doc/models.go
zenovich: It is used only here (by the groupsMembershipHistory service)

PR #211, opened 2019-08-18
Implement the contestListMembersAdditionalTime service

zenovich:
[contestListMembersAdditionalTime](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit?pli=1#heading=h.nax77vx2xko6)


app/api/contests/contests.go
```diff
@@ -22,6 +23,21 @@ func (srv *Service) SetRoutes(router chi.Router) {
 
 	router.Put("/contests/{item_id}/additional-time", service.AppHandler(srv.setAdditionalTime).ServeHTTP)
 	router.Get("/contests/{item_id}/group-by-name", service.AppHandler(srv.getGroupByName).ServeHTTP)
+	router.Get("/contests/{item_id}/members-additional-times", service.AppHandler(srv.getMembersAdditionalTimes).ServeHTTP)
```
smadbe: For all these services, I wonder (for a while), if we shouldn't do `/contests/{item_id}/groups/{group_id}/members-additional-times" instead of the query parameter :-/
What do you think?

zenovich: The problem is that setAdditionalTime sets the additional time for the specified group while
getMembersAdditionalTimes returns additional times for sub-groups of the specified group. So GET & PUT for the same URL would work with different objects which is strange.

smadbe: The url can still show the difference with something like `GET /contests/:item_id/groups/:group_id/members/additional_time` vs `PUT /contests/{item_id}/groups/:group_id/additional-time` . 
May look complicated at first sight but make sense.

zenovich: done

app/api/contests/contests.go
```diff
@@ -21,9 +21,12 @@ func (srv *Service) SetRoutes(router chi.Router) {
 	router.Use(render.SetContentType(render.ContentTypeJSON))
 	router.Use(auth.UserMiddleware(srv.Store.Sessions()))
 
-	router.Put("/contests/{item_id}/additional-time", service.AppHandler(srv.setAdditionalTime).ServeHTTP)
 	router.Get("/contests/{item_id}/group-by-name", service.AppHandler(srv.getGroupByName).ServeHTTP)
-	router.Get("/contests/{item_id}/members-additional-times", service.AppHandler(srv.getMembersAdditionalTimes).ServeHTTP)
+
```
smadbe: Should "/contests/{item_id}/group-by-name" be "/contests/{item_id}/groups/by-name" as well?

zenovich: fixed

app/api/contests/contests.go
```diff
@@ -20,9 +21,28 @@ func (srv *Service) SetRoutes(router chi.Router) {
 	router.Use(render.SetContentType(render.ContentTypeJSON))
 	router.Use(auth.UserMiddleware(srv.Store.Sessions()))
 
+	router.Get("/contests/{item_id}/groups/by-name", service.AppHandler(srv.getGroupByName).ServeHTTP)
+
 	router.Get("/contests/administered", service.AppHandler(srv.getAdministeredList).ServeHTTP)
-	router.Put("/contests/{item_id}/additional-time", service.AppHandler(srv.setAdditionalTime).ServeHTTP)
-	router.Get("/contests/{item_id}/group-by-name", service.AppHandler(srv.getGroupByName).ServeHTTP)
+
+  router.Put("/contests/{item_id}/groups/{group_id}/additional-times",
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	router.Put("/contests/{item_id}/groups/{group_id}/additional-times",
```

PR #212, opened 2019-08-20
Implement the groupsJoinByPassword service

zenovich:
[groupsJoinByPassword](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit?pli=1#heading=h.olscmtd3x2yx)


app/api/currentuser/current_user.go
```diff
@@ -33,6 +33,7 @@ func (srv *Service) SetRoutes(router chi.Router) {
 	router.Post("/current-user/group-requests/{group_id}", service.AppHandler(srv.sendGroupRequest).ServeHTTP)
 
 	router.Get("/current-user/group-memberships", service.AppHandler(srv.getGroupMemberships).ServeHTTP)
+	router.Post("/current-user/group-memberships", service.AppHandler(srv.joinGroupByPassword).ServeHTTP)
```
smadbe: As discussed in the doc, probably it should be `POST /current-user/group-memberships/by-password`

zenovich: fixed

smadbe:
Note that some elements of my last reviews are not addressed nor answered, so I am still waiting for them.

zenovich:
> Shouldn't we call it "code" directly everywhere (in the type, service url, ...) and do we wait for the DB schema to be updated?

I would prefer to wait for the DB schema to be updated.

> Note that some elements of my last reviews are not addressed nor answered, so I am still waiting for them.

No problem. I hope it's not me you're waiting for in this PR.

smadbe:
> > Note that some elements of my last reviews are not addressed nor answered, so I am still waiting for them.
> 
> No problem. I hope it's not me you're waiting for in this PR.
Yeah, it is you 😀


> * I discussed with Mathias for the type. We'll call it `joinedByCode` or `joinedByPassword`(cfr next remark).
Still not addressed

> > Shouldn't we call it "code" directly everywhere (in the type, service url, ...) and do we wait for the DB schema to be updated?
> 
> I would prefer to wait for the DB schema to be updated.

Meaning we would need to change the new type in the enum.

zenovich:
> The set `'invitationAccepted', 'requestAccepted', 'joinedByCode'` is repeated at least 15x through the code (in SQL queries) (with or without `'direct'`), you should define them somewhere so that they do not have to be repeated.

fixed

app/database/group_groups.go
```diff
@@ -0,0 +1,21 @@
+package database
+
+// GroupRelationIsActiveCondition is an SQL condition restricting a value
+// to be one of ('direct', 'invitationAccepted', 'requestAccepted', 'joinedByCode')
```
smadbe: Instead of writing the values it matches (that the reader can read at the next line), I think we should mention what it means (so that other dev know if they are supposed to use it or not).
So, something like `GroupRelationIsActiveCondition is an SQL condition restricting group types to those matching a current actual membership of a user or group within a group.`
And for the next one:  `GroupUserRelationIsActiveCondition is an SQL condition restricting group types to those matching a current actual user membership within a group.`

zenovich: fixed

PR #213, opened 2019-08-20
Disallow leaving a group if now() < lockUserDeletionDate

zenovich:


app/api/currentuser/current_user.go
```diff
@@ -71,6 +71,14 @@ func (srv *Service) performGroupRelationAction(w http.ResponseWriter, r *http.Re
 		if !found {
 			return service.InsufficientAccessRightsError
 		}
+	} else if action == leaveGroupAction {
+		var found bool
+		found, err = srv.Store.Groups().ByID(groupID).
+			Where("lockUserDeletionDate IS NULL OR lockUserDeletionDate <= NOW()").HasRows()
+		service.MustNotBeError(err)
+		if !found {
+			return service.InsufficientAccessRightsError
```
smadbe: The message should give the reason, otherwise there is a high risk of not understanding why it fails and big headache to try to debug it.

app/api/currentuser/accept_group_invitation.go
```diff
@@ -24,7 +24,9 @@ import (
 //   required: true
 // responses:
 //   "201":
-//     "$ref": "#/responses/updatedOrNotChangedResponse"
+//     "$ref": "#/responses/updatedResponse"
+//   "205":
```
smadbe: If you do not change tests and do not change the service to return 205, the behaviour is not changed, isn't it?
BTW, I don't get why 205 meaning "Reset Content" (you even use the ` http.StatusResetContent` constant for it) is used for "unchanged".

zenovich: I didn't change the service. I changed only its swagger specs. The service always returned 205 as 'not changed'. Unfortunately, there is no such HTTP status "Not Changed" so many people use 205.

app/api/currentuser/current_user.go
zenovich: fixed

app/api/currentuser/accept_group_invitation.go
smadbe: Would you have a reference for that?

zenovich: The only link I can find now is this one: https://stackoverflow.com/a/14224164
I thought this way: 1) 205 is an acceptable status code, 2) Most of the time, the "not changed" response is caused by stale data in a browser so we should refresh the data. Although refresh != reset, it is still better than just "OK"

zenovich: The issue with 205 is `Since the 205 status code implies that no additional content will be
   provided, a server MUST NOT generate a payload in a 205 response.`

zenovich: Should we return 200 for "not changed"?

zenovich: fixed (will be 200 for PUT/DELETE and 201 for POST)

zenovich:
> ok, just make a consistent message "not changed" vs "not_changed" vs "unchanged"

will be done in #214

PR #214, opened 2019-08-21
Change success status codes of current-user sub-routes for groups relations

zenovich:
groupInvitationAccept: 201 created -> 200 updated
groupInvitationReject:  201 created -> 200 updated


app/api/currentuser/render_group_group_transition_result_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	for _, tt := range tests {
+		tt := tt
+		if len(tt.actions) == 0 {
+			tt.actions = []userGroupRelationAction{acceptInvitationAction, rejectInvitationAction, createGroupRequestAction, leaveGroupAction}
+		}
+		for _, action := range tt.actions {
+			t.Run(tt.name+": "+string(action), func(t *testing.T) {
+				var fn service.AppHandler = func(respW http.ResponseWriter, req *http.Request) service.APIError {
+					return RenderGroupGroupTransitionResult(respW, req, tt.result, action)
```
golangcibot: Using the variable on range scope `action` in function literal (from `scopelint`)

app/api/currentuser/accept_group_invitation.feature
```diff
@@ -23,13 +23,13 @@ Feature: User accepts an invitation to join a group
 
   Scenario: Successfully accept an invitation
     Given I am the user with ID "1"
-    When I send a POST request to "/current-user/group-invitations/11/accept"
```
smadbe: Why changing it? actually for REST "actions", I usually find better to use POST. This is an action with the several impacts, it is not just an update of an attribute by one given as parameter.

zenovich: It's kinda strange to have POST services returning 201 with message "updated" instead of "created". These two services actually update relations, not create new ones.

zenovich: `groupRequestCreate` stays POST

zenovich: okay, now it's fixed

PR #215, opened 2019-08-21
Add swagger specs for rejectGroupInvitation

zenovich:


app/api/currentuser/reject_group_invitation.go
```diff
@@ -6,6 +6,36 @@ import (
...
+//
+//   * If `groups_groups.sType` is `invitationRefused` already, the "unchanged" (200) response is returned.
+// parameters:
+// - name: group_id
+//   in: path
+//   type: integer
+//   required: true
+// responses:
+//   "200":
+//     "$ref": "#/responses/updatedOrNotChangedResponse"
```
smadbe: "Unchanged" instead of "NoChanged" here as well ? 🤔

zenovich: fixed

PR #216, opened 2019-08-21
Add swagger specs for groupsAvailableSearch

zenovich:


zenovich:
> Shouldn't we add a note in the specs about the potential perf issues?

PR #217, opened 2019-08-21
Add a unique index on groups.sPassword

zenovich:
+ remove an unused test step


zenovich:
The problem is that the groupChangePassword service will work incorrectly without this unique index :/

zenovich:
> I am pretty sure there will be some collisions (maybe just with empty strings) on some of the groups

AlgoreaBackend will not work properly if so.

smadbe:
Of course it won't work properly with collisions, but we cannot merge it for now for the reason I mentioned.

PR #218, opened 2019-08-21
Add swagger specs for groupRequestCreate

zenovich:


app/api/currentuser/send_group_request.go
```diff
@@ -6,6 +6,42 @@ import (
 	"github.com/France-ioi/AlgoreaBackend/app/service"
 )
 
+// swagger:operation POST /current-user/group-requests/{group_id} groups users groupRequestCreate
+// ---
+// summary: Send a request to join a group
```
smadbe: Let's use "create a request" instead of "send a request" (x2) so that the reader does not think there is email sent or whatever.

app/doc/responses.go
```diff
@@ -87,6 +87,20 @@ type CreatedResponse struct {
...
 }
 
+// The request has succeeded. The message shows if the object has been deleted.
+// swagger:response createdOrNotChangedResponse
+type createdOrNotChangedResponse struct {
+	// in: body
+	Body struct {
+		// enum: created,not changed
+		// required: true
+		Message string `json:"message"`
```
smadbe: Same remark as I did yesterday about the boolean. I have to retrieve the PR to see if you have answered... 😛

zenovich: I don't remember there was a comment about the boolean (don't see a boolean here)...

app/api/currentuser/send_group_request.go
zenovich: fixed

app/doc/responses.go
smadbe: I mentioned somewhere that it might be useful (for the frontend typically) to have a boolean in the "data" map to know if it is "created" of "not changed", so that the client app does not have to string comparison on the message.

PR #220, opened 2019-08-22
Add swagger spec for groupRequestsReject

zenovich:


app/api/groups/reject_requests.go
```diff
@@ -6,6 +6,43 @@ import (
...
+//   Lets an admin reject requests to join a group.
+//   On success the service sets `groups_groups.sType` to "requestRefused" and `sStatusDate` to current UTC time
+//   for each of `group_ids`.
+//
+//
+//   The authenticated user should be an owner of the `parent_group_id`, otherwise the 'forbidden' error is returned.
+//
+//
+//   The input `group_id` should have the input `parent_group_id` as a parent group and the
+//   `groups_groups.sType` should be "requestSent", otherwise the `group_id` gets skipped with
```
smadbe: Not clear what the http code in case of "gets skipped with `unchanged`"

zenovich: It is a mass update action, so the http code doesn't depend on per-group results.

smadbe: Ok then it is not clear what is the return code in which case. If it always return 200, even if all "unchanged" or "invalid", it should be written.

zenovich: fixed

PR #239, opened 2019-09-02
Add swagger docs for itemCreate

zenovich:


app/api/items/add_item.go
```diff
@@ -112,6 +156,50 @@ func (in *NewItemRequest) canCreateItemsRelationsWithoutCycles(store *database.D
...
 
+// swagger:operation POST /items items itemCreate
+// ---
+// summary: Create an item
+// description: >
+//
+//   Creates an item with parameters from the input data with `items.idDefaultLanguage` = `language_id`.
+//   Also it
+//
+//     * inserts a row into `users_items` with given `language_id`, `title`, `image_url`, `subtitle`, `description`,
```
smadbe: in `items_strings` you mean?

app/api/items/add_item.go
```diff
@@ -112,6 +156,50 @@ func (in *NewItemRequest) canCreateItemsRelationsWithoutCycles(store *database.D
...
+//
+//     * creates a new `groups_items` row with: `idItem` = `items.ID`, `idGroup` = `idGroupSelf` of the current user,
```
smadbe: Let's give the meaning instead (keep the tech details in parenthesis if you think it is relevant): "give full access to the item to the current user".

app/formdata/form_data_integration_test.go
```diff
@@ -471,7 +471,7 @@ func TestFormData_ConstructMapForDB(t *testing.T) {
 			map[string]interface{}{},
 		},
 		{
-			"skips unexported fields",
+			"skips fields without json tag",
```
smadbe: So maybe you should export it (uppercase) so that you do not test 2 things at the same time.

app/api/items/add_item.go
```diff
@@ -112,6 +156,50 @@ func (in *NewItemRequest) canCreateItemsRelationsWithoutCycles(store *database.D
...
+//     * adds new relations for the parent and (optionally) children items into `items_items` and propagates `groups_items`.
+//
+//   The user should be an owner/manager of
+//
+//     * the `parent_item_id`,
+//     * `children` items (if any),
+//     * `unlocked_item_ids` items (if any),
+//
+//   and be an owner of `team_in_group_id` (if given),
+//   otherwise the "bad request" response is returned.
```
smadbe: "forbidden" would be more appropriate for most of these. Shouldn't we change?

app/formdata/form_data_integration_test.go
zenovich: We don't skip unexported fields anymore, so this test checks only one thing.

app/api/items/add_item.go
zenovich: I don't think "forbidden" would be more appropriate here because it is just one of many possible errors of the form. If the validation fails, this error gets listed in `errors` among other form-related errors.

app/api/items/add_item.go
```diff
@@ -17,63 +17,107 @@ import (
...
+	TeamInGroupID   *int64 `json:"team_in_group_id" sql:"column:idTeamInGroup" validate:"team_in_group_id"`
+	TeamMaxMembers  int32  `json:"team_max_members" sql:"column:iTeamMaxMembers"`
+	TitleBarVisible bool   `json:"title_bar_visible" sql:"column:bTitleBarVisible"`
+	HasAttempts     bool   `json:"has_attempts" sql:"column:bHasAttempts"`
+	// Nullable
+	AccessOpenDate *time.Time `json:"access_open_date" sql:"column:sAccessOpenDate"`
+	// Nullable
+	//
+	// MySQL time
+	Duration *string `json:"duration" sql:"column:sDuration" validate:"duration"`
```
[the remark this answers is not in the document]
zenovich: fixed

app/api/items/add_item.go
zenovich: fixed

app/api/items/add_item.go
zenovich: fixed

app/api/items/add_item.go
```diff
@@ -112,6 +158,50 @@ func (in *NewItemRequest) canCreateItemsRelationsWithoutCycles(store *database.D
...
+// ---
+// summary: Create an item
+// description: >
+//
+//   Creates an item with parameters from the input data with `items.idDefaultLanguage` = `language_id`.
+//   Also it
+//
+//     * inserts a row into `items_strings` with given `language_id`, `title`, `image_url`, `subtitle`, `description`,
+//
+//     * gives full access to the item for the current user (creates a new `groups_items` row with: `idItem` = `items.ID`, `idGroup` = `idGroupSelf` of the current user,
```
golangcibot: line is 169 characters (from `lll`)

app/formdata/form_data_integration_test.go
smadbe: This test differs in 2 points from other tests:
* it has no tags
* "name" is not exported

So your test, as it is, does show the effect of either one or the other, without guaranteeing that it is because of "no tags" that we get this result.

zenovich: improved the test

smadbe: Ok, that's better, but why still doing two tests in the same test. Reading the code, I don't know if unexported attributes are converted or not. 😕
If you want to put unexported attribute is a test, that's a good idea (that's why the initial test was about that), just do a one test with an unexported attribute and another one with no tags.

Sorry to bother you with such details 😞

zenovich: fixed :)

app/formdata/form_data_integration_test.go
```diff
@@ -536,11 +536,19 @@ func TestFormData_ConstructMapForDB(t *testing.T) {
...
+				Name string
 			}{},
-			`{"name":"test"}`,
+			`{"Name":"Test"}`,
+			map[string]interface{}{},
+		},
+		{
+			"skips unexported attributes",
+			&struct {
+				name string `json:"name" gorm:"column:name"`
```
golangcibot: structtag: struct field name has json tag but is not exported (from `govet`)

PR #243, opened 2019-09-04
Fix itemsBreadCrumbsData and add swagger docs for it

zenovich:


smadbe:
Please refactor the service to achieve 100% coverage.

zenovich:
> Please refactor the service to achieve 100% coverage.

done

app/api/items/get_breadcrumbs.go
```diff
@@ -0,0 +1,113 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/go-chi/render"
+	"github.com/jinzhu/gorm"
+
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+// swagger:operation GET /items items itemsBreadCrumbsData
```
smadbe: Looks weird to have `GET /items` for getting breadcrumb over an always-given set of ids. 
Don't you think we should pass the ids in the URL, so getting `GET /items/1,2,3` instead? Or even `GET /items/{item_ids:[\d,]+}/breadcrumb` as the ids cannot even be random ids (must be in hierarchy) and the data not really all item info.

zenovich: fixed

PR #245, opened 2019-09-04
Return "forbidden" error if a parent group is a team/user or a child group is a user in groupAddChild & groupRemoveChild

### LINKED ISSUE #244: In groupAddChild: reject if the parent is a team and if parent or child group is a userself (as the service is for non-user membership only)

zenovich:
Fixes #244 

The groupRemoveChild service was changed too for the sake of symmetry.

### DIFF
--- a/app/api/groups/add_child.go
+++ b/app/api/groups/add_child.go
@@ -20,8 +20,8 @@ import (
 //   Restrictions (otherwise the 'forbidden' error is returned):
 //     * the authenticated user should be an owner of both `parent_group_id` and `child_group_id,
 //     * the authenticated user should have `users.allowSubgroups` set to 1,
-//     * the parent group should not be of type "UserSelf",
-//     * the child group should not be of types "Base" or "UserAdmin",
+//     * the parent group should not be of type "UserSelf" or "Team",
+//     * the child group should not be of types "Base", "UserSelf", or "UserAdmin",
 //     * the action should not create cycles in the groups relations graph.
 // parameters:
 // - name: parent_group_id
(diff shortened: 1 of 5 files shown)


app/api/groups/add_child.go
```diff
@@ -20,8 +20,8 @@ import (
 //   Restrictions (otherwise the 'forbidden' error is returned):
 //     * the authenticated user should be an owner of both `parent_group_id` and `child_group_id,
 //     * the authenticated user should have `users.allowSubgroups` set to 1,
-//     * the parent group should not be of type "UserSelf",
-//     * the child group should not be of types "Base" or "UserAdmin",
+//     * the parent group should not be of type "UserSelf" or "Team",
+//     * the child group should not be of types "Base", "UserSelf", or "UserAdmin",
```
smadbe: We should add a note that users are not allowed as child as they need to use the createInvite/request services for that. Otherwise, we could wonder why "UserSelf" is not allowed here (the others are invalid relations while here it is just because it is handled differently)

app/api/groups/remove_child.go
```diff
@@ -31,8 +31,8 @@ import (
 //
 //   Restrictions (otherwise the 'forbidden' error is returned):
 //     * the authenticated user should be an owner of both `parent_group_id` and `child_group_id,
-//     * the parent group should not be of type "UserSelf",
-//     * the child group should not be of types "Base" or "UserAdmin",
+//     * the parent group should not be of type "UserSelf" or "Team",
+//     * the child group should not be of types "Base", "UserSelf" or "UserAdmin",
```
smadbe: same remark as above

app/api/groups/add_child.go
zenovich: done

app/api/groups/remove_child.go
zenovich: done

PR #253, opened 2019-09-09
Fix comments of GroupStore.TeamGroupForTeamItemAndUser() and GroupStore.TeamGroupForItemAndUser()

zenovich:


app/database/group_store.go
```diff
@@ -27,7 +27,7 @@ func (s *GroupStore) TeamGroupForTeamItemAndUser(itemID int64, user *User) *DB {
 }
 
 // TeamGroupForItemAndUser returns a composable query for getting a team (the first one in the order of`groups.ID`)
-// for the current user by one of team's items
+// for the given user by one of team's items
```
smadbe: This sentence is quite hard to understand out of context (the one from TeamGroupForTeamItemAndUser as well). After a few minutes on it, I had to find its usage to understand what it truly means. Could we make it a bit more clear? (can be longer that's not a problem)

zenovich: fixed

PR #261, opened 2019-09-11
Add swagger docs for the itemView service

zenovich:


app/api/items/get_item.go
```diff
@@ -11,106 +11,239 @@ import (
...
+// ---
+// summary: Get an item
+// description: Returns data related to the specified item, its children,
+//              and the current user's interactions with them
+//              (from tables `items`, `items_items`, `items_string`, and `users_items`).
+//
+//
+//              * If the specified item is not visible by the current user, the 'not found' response is returned.
+//
+//              * If the current user has only grayed access on the specified item, the 'forbidden' error is returned.
```
smadbe: why is there field with "only if not greyed" then? 🤔

zenovich: The root item cannot be grayed, but children still can

app/api/items/get_item.go
```diff
@@ -11,106 +11,239 @@ import (
...
+	// required: true
+	Title string `json:"title"`
+	// Nullable
+	// required: true
+	ImageURL *string `json:"image_url"`
 }
 
 type itemStringNotGrayed struct {
-	Subtitle    *string `json:"subtitle"`
+	// Nullable; only if not grayed
```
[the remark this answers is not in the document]
smadbe: Ok. Still not sure the "only if ..." is clear, but that should be good enough

PR #270, opened 2019-09-20
Add a new column items_items.partial_access_propagation superseding always_visible & access_restricted

### LINKED ISSUE #267: Partial propagation as a single attribute
The way partial access is propagated from an item to his child is defined is  `items_items.bAlwaysVisible` and `items_items.bAccessRestricted`. In addition to be not very clear, they allow expressing 4 "cases" while there are actually 3 cases.

So, convert `items_items.bAlwaysVisible` and `items_items.bAccessRestricted` to one single attribute  `items_items.partial_access_propagation` of type `ENUM('none','as grey','as partial')`.

Map the existing data:
* ! restricted -> 'as partial'
* restricted & always_visible -> 'as grey'
* restricted & !always_visible -> 'no'

zenovich:
Fixes #267


db/migrations/44_replace_items_items_always_visible_and_access_restricted_with_partial_access_propagation.sql
```diff
@@ -0,0 +1,269 @@
...
+                NEW.`partial_access_propagation`,NEW.`difficulty`);
+    INSERT IGNORE INTO `groups_items_propagate`
+        SELECT `id`, 'children' as `propagate_access` FROM `groups_items`
+            WHERE `groups_items`.`item_id` = NEW.`item_parent_id`
+            ON DUPLICATE KEY UPDATE propagate_access='children';
+END
+-- +migrate StatementEnd
+DROP TRIGGER IF EXISTS `before_update_items_items`;
+-- +migrate StatementBegin
+CREATE DEFINER=`algorea`@`%` TRIGGER `before_update_items_items` BEFORE UPDATE ON `items_items` FOR EACH ROW BEGIN
```
smadbe: Waw I didn't expect such a big migration. 

Do we want to keep these triggers (ignoring history stuffs)? Maybe that's better if we consider we'll add FK with possible "delete cascade" which may be hard to address in services.

zenovich: These triggers are mostly related to items access rights propagation and items ancestors creation.

app/api/items/get_item.go
```diff
@@ -164,13 +164,10 @@ type itemChildNode struct {
 	// `items_items.category`
 	// required: true
 	Category string `json:"category"`
-	// `items_items.always_visible`
-	// enum: Undefined,Discovery,Application,Validation,Challenge
+	// `items_items.partial_access_propagation`
+	// enum: None,AsGrayed,AsPartial
```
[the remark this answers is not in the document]
zenovich: fixed

PR #276, opened 2019-09-24
Remove groups_items.propagate_access column

### LINKED ISSUE #274: Remove remains of sPropagateAccess in `groups_items`
The table `groups_items_propagate` has been created to put away of `groups_items` the attribute `sPropagateAccess`, in particular to avoid changing `groups_items` too often, which run a trigger and store more entries.

However `sPropagateAccess` is still stored in both `groups_items` and  `groups_items_propagate` and still  `groups_items.sPropagateAccess` is still used in the propagation algorithm while it should not.

Remove the usage of `groups_items.sPropagateAccess` in the "compute all access" propagation (without breaking it 😁) and remove the attribute from `groups_items`.

zenovich:
Fixes #274


db/migrations/48_remove_groups_items_propagate_access.sql
```diff
@@ -0,0 +1,278 @@
...
+                NEW.`cached_access_solutions_date`,NEW.`cached_grayed_access_date`,NEW.`cached_full_access`,
+                NEW.`cached_partial_access`,NEW.`cached_access_solutions`,NEW.`cached_grayed_access`,
+                NEW.`cached_manager_access`
+            );
+    END IF;
+END
+-- +migrate StatementEnd
+DROP TRIGGER IF EXISTS `after_update_groups_items`;
+-- +migrate StatementBegin
+CREATE DEFINER=`algorea`@`%` TRIGGER `after_update_groups_items` AFTER UPDATE ON `groups_items` FOR EACH ROW BEGIN
```
smadbe: These trigger names are quite bad names, they do not define what the function does but when it is run, which already present in the trigger metadata. I don't know if it make sense to do it from now on (for new ones). A bit related with my other comments about comments,  maybe if it has a good name, it does not require comments.

app/database/group_item_store_compute_all_access.go
```diff
@@ -192,11 +178,12 @@ func (s *GroupItemStore) computeAllAccess() {
...
 	for hasChanges {
-		_, err = stmtInsertMissingChildren.Exec()
+		_, err = stmtDropTemporaryTable.Exec()
 		mustNotBeError(err)
-		_, err = stmtInsertMissingPropagate.Exec()
+		_, err = stmtCreateTemporaryTable.Exec()
 		mustNotBeError(err)
-		_, err = stmtUpdatePropagateAccess.Exec()
+		mustNotBeError(s.Exec(queryInsertMissingChildren).Error())
+		_, err = stmtDropTemporaryTable.Exec()
```
smadbe: Meaning your drop will be run twice in a row each time with a single create. If you want to have a "drop if exist" before your "create", just put the drop after the loop.

app/database/group_item_store_compute_all_access.go
```diff
@@ -192,11 +178,12 @@ func (s *GroupItemStore) computeAllAccess() {
...
 	hasChanges := true
 	for hasChanges {
-		_, err = stmtInsertMissingChildren.Exec()
+		_, err = stmtDropTemporaryTable.Exec()
 		mustNotBeError(err)
-		_, err = stmtInsertMissingPropagate.Exec()
+		_, err = stmtCreateTemporaryTable.Exec()
 		mustNotBeError(err)
-		_, err = stmtUpdatePropagateAccess.Exec()
+		mustNotBeError(s.Exec(queryInsertMissingChildren).Error())
```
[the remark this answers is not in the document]
zenovich: It cannot be prepared because it uses a temporary table which doesn't exist when we prepare statements, so mysql errs.

app/database/group_item_store_compute_all_access.go
```diff
@@ -39,35 +53,7 @@ func (s *GroupItemStore) computeAllAccess() {
...
-		UPDATE groups_items
-		SET propagate_access='done'
-		WHERE propagate_access='self'`
-	stmtUpdatePropagateAccess, err = s.db.CommonDB().Prepare(queryUpdatePropagateAccess)
-	mustNotBeError(err)
-	defer func() { mustNotBeError(stmtUpdatePropagateAccess.Close()) }()
+		JOIN parents_propagate ON parents_propagate.id = parents.id`
 
 	// mark as 'done' groups_items_propagate that shouldn't propagate (having items.custom_chapter=1)
 	const queryMarkDoNotPropagate = `
```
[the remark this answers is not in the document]
zenovich: > Wouldn't be faster (maybe slower actually) to join with groups_items_propagate and only apply this to propagate_access!='done' ?

Probably it would be worse: with the proposed modification MySQL uses an internal temporary table:
```mysql> explain INSERT INTO groups_items_propagate (id, propagate_access) SELECT groups_items.id AS id, 'done' as propagate_access FROM groups_items JOIN items ON groups_items.item_id = items.id AND items.custom_chapter LEFT JOIN groups_items_propagate AS prop ON prop.id = groups_items.id AND prop.propagate_access = 'done' WHERE prop.id IS NULL ON DUPLICATE KEY UPDATE propagate_access='done';
+----+-------------+------------------------+------------+--------+--------------------------+------------+---------+--------------------------------------+------+----------+------------------------------+
| id | select_type | table                  | partitions | type   | possible_keys            | key        | key_len | ref                                  | rows | filtered | Extra                        |
+----+-------------+------------------------+------------+--------+--------------------------+------------+---------+--------------------------------------+------+----------+------------------------------+
|  1 | INSERT      | groups_items_propagate | NULL       | ALL    | NULL                     | NULL       | NULL    | NULL                                 | NULL |     NULL | NULL                         |
|  1 | SIMPLE      | groups_items           | NULL       | index  | item_id,itemtem_id       | itemtem_id | 8       | NULL                                 |    1 |   100.00 | Using index; Using temporary |
|  1 | SIMPLE      | items                  | NULL       | eq_ref | PRIMARY                  | PRIMARY    | 8       | algorea_db_test.groups_items.item_id |    1 |   100.00 | Using where                  |
|  1 | SIMPLE      | prop                   | NULL       | eq_ref | PRIMARY,propagate_access | PRIMARY    | 8       | algorea_db_test.groups_items.id      |    1 |   100.00 | Using where; Not exists      |
+----+-------------+------------------------+------------+--------+--------------------------+------------+---------+--------------------------------------+------+----------+------------------------------+```

File: db/migrations/48_remove_groups_items_propagate_access.sql (hunk 1 above)
Comment (zenovich) [reply db/migrations/48_remove_groups_items_propagate_access.sql]: For me, `after_update_groups_items` is a perfect name. It really describes the trigger (no matter what it does inside).

File: app/database/group_item_store_compute_all_access.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (zenovich) [reply app/database/group_item_store_compute_all_access.go]: > For the next next, queryMarkExistingChildren, it is over-complicated by the fact they want a (completely useless) ID for their entry 😕. Meaning also much more complicated to understand than what it is really.

They are not completely useless

File: app/database/group_item_store_compute_all_access.go [hunk 5]
```diff
@@ -39,35 +53,7 @@ func (s *GroupItemStore) computeAllAccess() {
...
-	// Set groups_items as set up for propagation
-	// (switch groups_items.propagate_access from 'self' to 'done')
-	const queryUpdatePropagateAccess = `
-		UPDATE groups_items
-		SET propagate_access='done'
-		WHERE propagate_access='self'`
-	stmtUpdatePropagateAccess, err = s.db.CommonDB().Prepare(queryUpdatePropagateAccess)
-	mustNotBeError(err)
-	defer func() { mustNotBeError(stmtUpdatePropagateAccess.Close()) }()
+		JOIN parents_propagate ON parents_propagate.id = parents.id`
```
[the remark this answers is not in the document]
zenovich: > I suggest an easy reorganisation in this function:
> I suppose you do not want to extract requests out of the function as it would not make things easier.
> If so, I would suggest to only declare statement without executing any of them in the first part of the function... then draw clearly a line a start the processing, and run the statements that has to be run before the `for hasChanges` loop.
> It will make the "workflow" much clearer.
> 
> BTW, are you sure these requests has necessarily to be run twice (as they are run before the loop and then again as `hasChanged` is true at first.

I don't get what you want to change. We only run queries inside the loop, there are no queries executed before the loop.

app/database/group_item_store_compute_all_access.go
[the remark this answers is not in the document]
smadbe: The key should be `group_id,item_id`. Here it is not clear that it is actually a key for the other table :-/, but it could be worst, with an ID + groups_items_id, true.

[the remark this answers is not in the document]
smadbe: But it forces going through the groups_items while it may not be needed.

app/database/group_item_store_compute_all_access.go
zenovich: fixed (did you mean 'before')?

db/migrations/48_remove_groups_items_propagate_access.sql
```diff
@@ -0,0 +1,278 @@
...
+    IF NOT (NEW.`full_access_date` <=> OLD.`full_access_date`AND NEW.`partial_access_date` <=> OLD.`partial_access_date`AND
```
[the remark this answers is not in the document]
zenovich: Do you want me to add something like `-- Mark that the groups_items row needs access rights propagation` right before the ```INSERT INTO `groups_items_propagate` ```?

app/database/group_item_store_compute_all_access.go
smadbe: Moving the first one before, or the second one after, it is all the same.

db/migrations/48_remove_groups_items_propagate_access.sql
zenovich: Also, note that in AlgoreaPlatform these triggers are generated by the PHP code.

app/database/group_item_store_compute_all_access.go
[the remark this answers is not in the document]
zenovich: added more comments

app/database/group_item_store_compute_all_access.go
```diff
@@ -174,6 +177,9 @@ func (s *GroupItemStore) computeAllAccess() {
 	mustNotBeError(err)
 	defer func() { mustNotBeError(stmtMarkChildrenItems.Close()) }()
 
+	// ------------------------------------------------------------------------------------
+	// Here we execute the statements
```
smadbe: 👍

app/database/group_item_store_compute_all_access.go
```diff
@@ -39,35 +53,7 @@ func (s *GroupItemStore) computeAllAccess() {
...
-	const queryUpdatePropagateAccess = `
-		UPDATE groups_items
-		SET propagate_access='done'
-		WHERE propagate_access='self'`
-	stmtUpdatePropagateAccess, err = s.db.CommonDB().Prepare(queryUpdatePropagateAccess)
-	mustNotBeError(err)
-	defer func() { mustNotBeError(stmtUpdatePropagateAccess.Close()) }()
+		JOIN parents_propagate ON parents_propagate.id = parents.id`
 
 	// mark as 'done' groups_items_propagate that shouldn't propagate (having items.custom_chapter=1)
```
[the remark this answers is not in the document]
smadbe: But then, you still have a `groups_items_propagate` table growing forever while it should always be very small. The updates were design for the case when it was just a column.

Can you explain your risk of deadlocks?

db/migrations/48_remove_groups_items_propagate_access.sql
zenovich: http://www.dba-oracle.com/t_trigger_naming_standards.htm

app/database/group_item_store_compute_all_access.go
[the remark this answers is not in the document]
zenovich: The table isn't growing forever because it has same IDs as `groups_items`, so it has the same number of rows.

db/migrations/48_remove_groups_items_propagate_access.sql
zenovich: Actually people use this weird naming for triggers because it was impossible to have more than one trigger for the same table and the same event before MySQL 5.7.2. Although it is possible with 5.7.2+, the tradition persists. I would keep the naming just for the sake of consistency (all our triggers are named in that way).

db/migrations/48_remove_groups_items_propagate_access.sql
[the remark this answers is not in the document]
zenovich: fixed

app/database/group_item_store_compute_all_access.go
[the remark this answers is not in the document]
smadbe: Yes, which may be large >< very short if we keep only the ongoing ones. It is a 1<->0-1 relation, these "done" entries do not make sense anymore. 

Can you explain your risk of deadlocks?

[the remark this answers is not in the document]
zenovich: It's all instinct at this point, but we will see soon. Fixed.

app/database/group_item_store_compute_all_access.go
```diff
@@ -190,13 +171,18 @@ func (s *GroupItemStore) computeAllAccess() {
...
-		_, err = stmtInsertMissingChildren.Exec()
-		mustNotBeError(err)
-		_, err = stmtInsertMissingPropagate.Exec()
+		_, err = stmtCreateTemporaryTable.Exec()
 		mustNotBeError(err)
-		_, err = stmtUpdatePropagateAccess.Exec()
+		mustNotBeError(s.Exec(queryInsertMissingChildren).Error())
+		_, err = stmtDropTemporaryTable.Exec()
 		mustNotBeError(err)
 		_, err = stmtMarkDoNotPropagate.Exec()
```
smadbe: I would expect `stmtMarkDoNotPropagate` to be run before creating the missing children (2 steps earlier), so that possibly many children are not created for "nothing". Do you agree?

app/database/group_item_store_compute_all_access.go
```diff
@@ -102,11 +86,8 @@ func (s *GroupItemStore) computeAllAccess() {
 	mustNotBeError(err)
 	defer func() { mustNotBeError(stmtMarkExistingChildren.Close()) }()
 
-	// marking 'children' groups_items_propagate as 'done'
-	const queryMarkFinishedItems = `
-		UPDATE groups_items_propagate
-		SET propagate_access = 'done'
-		WHERE propagate_access = 'children'`
+	// deleting 'children' groups_items_propagate
+	const queryMarkFinishedItems = `DELETE FROM groups_items_propagate WHERE propagate_access = 'children'`
```
smadbe: Good.
So now you have delete, all those with "children", the one only remaining in `groups_items_propagate` are those with 'self'. So: 
* `// computation for groups_items marked as 'self' in groups_items_propagate.` -> `// computation for groups_items marked as 'self' in groups_items_propagate (so all of them).`
* `LEFT JOIN groups_items_propagate` -> remove
* `(groups_items_propagate.propagate_access = 'self' OR groups_items_propagate.id IS NULL) ` -> remove
* `WHERE groups_items_propagate.propagate_access = 'self'`  -> remove
* maybe I would move the `JOIN groups_items_propagate USING(id)` to the second line of the request, so that it is clearer that we work only the propagating ones.

Actually, for the  `LEFT JOIN groups_items_propagate` and its "where", it was quite useless. Do you find a reason?

smadbe: Probably you should rename `stmtMarkFinishedItems` -> `stmtRemoveProcessedItems` 
and `stmtMarkDoNotPropagate` -> `stmtDeleteDoNotPropagate`

db/migrations/48_remove_groups_items_propagate_access.sql
```diff
@@ -116,7 +118,7 @@ CREATE DEFINER=`algorea`@`%` TRIGGER `after_insert_items_items` AFTER INSERT ON
     INSERT INTO `groups_items_propagate`
     SELECT `id`, 'children' as `propagate_access` FROM `groups_items`
     WHERE `groups_items`.`item_id` = NEW.`parent_item_id`
-    ON DUPLICATE KEY UPDATE propagate_access = IF(propagate_access = 'done', 'children', propagate_access);
+    ON DUPLICATE KEY UPDATE propagate_access = IF(propagate_access != 'self', 'children', propagate_access);
```
smadbe: I don't get your modification. Of course, if `propagate_access != 'self'` so is equals to 'children', put 'children", so put the same value in any case ... so it is just a "on duplicate key -> ignore"

same x3

smadbe:
BTW, the comment above of `computeAllAccess` has to be adapted a bit.

app/database/group_item_store_compute_all_access.go
zenovich: No, I do not agree here. Although we do not propagate access from custom chapters to their children, we still need to propagate access to custom chapters themselves.

app/database/group_item_store_compute_all_access.go
zenovich: That would be true if we had a single-task system. But in AlgoreaBackend parellel requests can insert new rows with `propagate_access` = 'children' between the DELETE and UPDATE. And the UPDATE statement will see and process those new rows. So we still need to filter by `propagate_access` :(

app/database/group_item_store_compute_all_access.go
zenovich: > No, I do not agree here. Although we do not propagate access from custom chapters to their children, we still need to propagate access to custom chapters themselves.

At least the 'database schema comments' doc says that. Nevertheless, AlogoreaPlatform even doesn't propagate access rights from to custom chapters themselves and we implemented that in the same way. Probably it's a bug. How should it work?

app/database/group_item_store_compute_all_access.go
smadbe: There is no lock anymore? 🤔 So it can be completely interleaved? How do you ensure that work correctly?
Ok, so you can take these comments, probably some still make a bit of sense.  And the left join is still strange to me.

zenovich: On the other hand, we should never commit a transaction with unprocessed `groups_items_propagate` rows...

zenovich: So should be okay unless someone inserts items via mysql console

zenovich: Fixed

zenovich: Fixed

db/migrations/48_remove_groups_items_propagate_access.sql
zenovich: fixed

app/database/group_item_store_compute_all_access.go
smadbe: Ah I see you did it at the end? 🤔
The change I was suggesting does not change anything to the logic, but just was not creating "empty" groups_items for nothing.  (except if I did a mistake)

Ah you have added `AND propagate_access = 'children'`. It looks like a good idea to make it correct, right.

app/database/group_item_store_compute_all_access.go
```diff
@@ -112,14 +113,12 @@ func (s *GroupItemStore) computeAllAccess() {
...
 				ON items_items.child_item_id = child.item_id
-			LEFT JOIN groups_items_propagate
-				ON groups_items_propagate.id = child.id
 			JOIN groups_items AS parent
 				ON parent.item_id = items_items.parent_item_id AND parent.group_id = child.group_id
 			JOIN items AS parent_item
 				ON parent_item.id = items_items.parent_item_id
 			WHERE
-				(groups_items_propagate.propagate_access = 'self' OR groups_items_propagate.id IS NULL) AND
+				child.id = groups_items_propagate.id AND
```
smadbe: Is it useful? The "using (id)" should do the job correctly, but it depends if it succeeds doing it in an efficient way.

app/database/group_item_store_compute_all_access.go
```diff
@@ -163,13 +160,13 @@ func (s *GroupItemStore) computeAllAccess() {
 	}()
 
 	// marking 'self' groups_items_propagate as 'children'
-	const queryMarkChildrenItems = `
+	const queryMarkSelfAsChildren = `
```
smadbe: that's a better name 👍

app/database/group_item_store_compute_all_access.go
zenovich: It was added to optimize the query. Otherwise MySQL aggregated info for all the groups_items.

zenovich: With filtering:
```
mysql> explain UPDATE groups_items JOIN groups_items_propagate USING(id) LEFT JOIN LATERAL ( SELECT child.id, MIN(parent.cached_full_access_date) AS cached_full_access_date, MIN(IF(items_items.partial_access_propagation = 'AsPartial', parent.cached_partial_access_date, NULL)) AS cached_partial_access_date, MAX(parent.cached_manager_access) AS cached_manager_access, MIN(IF(items_items.partial_access_propagation = 'AsGrayed', parent.cached_partial_access_date, NULL)) AS cached_grayed_access_date, MIN(parent.cached_access_solutions_date) AS cached_access_solutions_date, CONCAT('From ancestor group(s) ', GROUP_CONCAT( DISTINCT IF(parent.access_reason = '', NULL, parent.access_reason) ORDER BY parent_item.id SEPARATOR ', ' )) AS access_reason_ancestors FROM groups_items AS child JOIN items_items ON items_items.child_item_id = child.item_id JOIN groups_items AS parent ON parent.item_id = items_items.parent_item_id AND parent.group_id = child.group_id JOIN items AS parent_item ON parent_item.id = items_items.parent_item_id WHERE child.id = groups_items_propagate.id AND ( parent.cached_full_access_date IS NOT NULL OR (parent.cached_partial_access_date IS NOT NULL AND items_items.partial_access_propagation != 'None') OR parent.cached_access_solutions_date IS NOT NULL OR parent.cached_manager_access ) AND parent_item.custom_chapter = 0 GROUP BY child.id ) AS new_data USING(id) SET groups_items.cached_full_access_date = LEAST( IFNULL(new_data.cached_full_access_date, groups_items.full_access_date), IFNULL(groups_items.full_access_date, new_data.cached_full_access_date) ), groups_items.cached_partial_access_date = LEAST( IFNULL(new_data.cached_partial_access_date, groups_items.partial_access_date), IFNULL(groups_items.partial_access_date, new_data.cached_partial_access_date) ), groups_items.cached_manager_access = GREATEST( IFNULL(new_data.cached_manager_access, 0), groups_items.manager_access ), groups_items.cached_access_solutions_date = LEAST( IFNULL(new_data.cached_access_solutions_date, groups_items.access_solutions_date), IFNULL(groups_items.access_solutions_date, new_data.cached_access_solutions_date) ), groups_items.cached_grayed_access_date = new_data.cached_grayed_access_date, groups_items.cached_access_reason = new_data.access_reason_ancestors;
+----+-------------------+------------------------+------------+--------+-----------------------------------------------------------------------------------------+------------------+---------+---------------------------------------------------------------------------+--------+----------+-----------------------------------------+
| id | select_type       | table                  | partitions | type   | possible_keys                                                                           | key              | key_len | ref                                                                       | rows   | filtered | Extra                                   |
+----+-------------------+------------------------+------------+--------+-----------------------------------------------------------------------------------------+------------------+---------+---------------------------------------------------------------------------+--------+----------+-----------------------------------------+
|  1 | PRIMARY           | groups_items_propagate | NULL       | index  | PRIMARY                                                                                 | propagate_access | 1       | NULL                                                                      | 200191 |   100.00 | Using index; Rematerialize (<derived2>) |
|  1 | UPDATE            | groups_items           | NULL       | eq_ref | PRIMARY                                                                                 | PRIMARY          | 8       | algorea_example.groups_items_propagate.id                                 |      1 |   100.00 | NULL                                    |
|  1 | PRIMARY           | <derived2>             | NULL       | ref    | <auto_key0>                                                                             | <auto_key0>      | 8       | algorea_example.groups_items_propagate.id                                 |      2 |   100.00 | NULL                                    |
|  2 | DEPENDENT DERIVED | child                  | NULL       | eq_ref | PRIMARY,item_id,version,group_id,itemtem_id,full_access,access_solutions,partial_access | PRIMARY          | 8       | algorea_example.groups_items_propagate.id                                 |      1 |   100.00 | NULL                                    |
|  2 | DEPENDENT DERIVED | items_items            | NULL       | ref    | parent_item_id,child_item_id,parent_child,parent_version                                | child_item_id    | 8       | algorea_example.child.item_id                                             |      1 |   100.00 | NULL                                    |
|  2 | DEPENDENT DERIVED | parent                 | NULL       | eq_ref | item_id,group_id,itemtem_id                                                             | item_id          | 16      | algorea_example.items_items.parent_item_id,algorea_example.child.group_id |      1 |    99.99 | Using where                             |
|  2 | DEPENDENT DERIVED | parent_item            | NULL       | eq_ref | PRIMARY                                                                                 | PRIMARY          | 8       | algorea_example.items_items.parent_item_id                                |      1 |    10.00 | Using where                             |
+----+-------------------+------------------------+------------+--------+-----------------------------------------------------------------------------------------+------------------+---------+---------------------------------------------------------------------------+--------+----------+-----------------------------------------+
7 rows in set, 2 warnings (0.02 sec)
```
Without filtering:
```
mysql> explain UPDATE groups_items JOIN groups_items_propagate USING(id) LEFT JOIN ( SELECT child.id, MIN(parent.cached_full_access_date) AS cached_full_access_date, MIN(IF(items_items.partial_access_propagation = 'AsPartial', parent.cached_partial_access_date, NULL)) AS cached_partial_access_date, MAX(parent.cached_manager_access) AS cached_manager_access, MIN(IF(items_items.partial_access_propagation = 'AsGrayed', parent.cached_partial_access_date, NULL)) AS cached_grayed_access_date, MIN(parent.cached_access_solutions_date) AS cached_access_solutions_date, CONCAT('From ancestor group(s) ', GROUP_CONCAT( DISTINCT IF(parent.access_reason = '', NULL, parent.access_reason) ORDER BY parent_item.id SEPARATOR ', ' )) AS access_reason_ancestors FROM groups_items AS child JOIN items_items ON items_items.child_item_id = child.item_id JOIN groups_items AS parent ON parent.item_id = items_items.parent_item_id AND parent.group_id = child.group_id JOIN items AS parent_item ON parent_item.id = items_items.parent_item_id WHERE ( parent.cached_full_access_date IS NOT NULL OR (parent.cached_partial_access_date IS NOT NULL AND items_items.partial_access_propagation != 'None') OR parent.cached_access_solutions_date IS NOT NULL OR parent.cached_manager_access ) AND parent_item.custom_chapter = 0 GROUP BY child.id ) AS new_data USING(id) SET groups_items.cached_full_access_date = LEAST( IFNULL(new_data.cached_full_access_date, groups_items.full_access_date), IFNULL(groups_items.full_access_date, new_data.cached_full_access_date) ), groups_items.cached_partial_access_date = LEAST( IFNULL(new_data.cached_partial_access_date, groups_items.partial_access_date), IFNULL(groups_items.partial_access_date, new_data.cached_partial_access_date) ), groups_items.cached_manager_access = GREATEST( IFNULL(new_data.cached_manager_access, 0), groups_items.manager_access ), groups_items.cached_access_solutions_date = LEAST( IFNULL(new_data.cached_access_solutions_date, groups_items.access_solutions_date), IFNULL(groups_items.access_solutions_date, new_data.cached_access_solutions_date) ), groups_items.cached_grayed_access_date = new_data.cached_grayed_access_date, groups_items.cached_access_reason = new_data.access_reason_ancestors;
+----+-------------+------------------------+------------+--------+-----------------------------------------------------------------------------------------+------------------+---------+---------------------------------------------------------------------------+--------+----------+-------------+
| id | select_type | table                  | partitions | type   | possible_keys                                                                           | key              | key_len | ref                                                                       | rows   | filtered | Extra       |
+----+-------------+------------------------+------------+--------+-----------------------------------------------------------------------------------------+------------------+---------+---------------------------------------------------------------------------+--------+----------+-------------+
|  1 | PRIMARY     | groups_items_propagate | NULL       | index  | PRIMARY                                                                                 | propagate_access | 1       | NULL                                                                      | 200191 |   100.00 | Using index |
|  1 | UPDATE      | groups_items           | NULL       | eq_ref | PRIMARY                                                                                 | PRIMARY          | 8       | algorea_example.groups_items_propagate.id                                 |      1 |   100.00 | NULL        |
|  1 | PRIMARY     | <derived2>             | NULL       | ref    | <auto_key0>                                                                             | <auto_key0>      | 8       | algorea_example.groups_items_propagate.id                                 |     10 |   100.00 | NULL        |
|  2 | DERIVED     | child                  | NULL       | index  | PRIMARY,item_id,version,group_id,itemtem_id,full_access,access_solutions,partial_access | PRIMARY          | 8       | NULL                                                                      | 205815 |   100.00 | NULL        |
|  2 | DERIVED     | items_items            | NULL       | ref    | parent_item_id,child_item_id,parent_child,parent_version                                | child_item_id    | 8       | algorea_example.child.item_id                                             |      1 |   100.00 | NULL        |
|  2 | DERIVED     | parent                 | NULL       | eq_ref | item_id,group_id,itemtem_id                                                             | item_id          | 16      | algorea_example.items_items.parent_item_id,algorea_example.child.group_id |      1 |    99.99 | Using where |
|  2 | DERIVED     | parent_item            | NULL       | eq_ref | PRIMARY                                                                                 | PRIMARY          | 8       | algorea_example.items_items.parent_item_id                                |      1 |    10.00 | Using where |
+----+-------------+------------------------+------------+--------+-----------------------------------------------------------------------------------------+------------------+---------+---------------------------------------------------------------------------+--------+----------+-------------+
7 rows in set, 1 warning (0.00 sec)```

zenovich: Note the "rows" column

PR #277, opened 2019-09-25
 Make sure the db seed schema match the production one

### LINKED ISSUE #272: Make sure the db seed schema match the production one.
I tried to apply our migrations on the loaded prod db on MySQL8 and it fails on some aspect of the schema and data (which is probably invalid in MySQL8).

1) I assume the schema we have does not match the prod one (strange, I don't know why), so it would need to be fixed
2) Some data (such as "0000-00-00 00:00" datetime) are not valid in MySQL8 and make the migrations fail. Probably it requires to add a first migration before all to fix these.

I had issues on migrations n°2, n°14 (for which I had to do `UPDATE history_groups SET sDateCreated = NULL WHERE sDateCreated < '1000-01-01'`), n°17, ...

---

Edit: 
a) The new prod export (from 22/09) should be emptied and considered as the the new "base schema" 
b) Then, we have to see what migrations fails with this schema (fixing error of type (1), see above)
c) Then, we have to see what may fail in prod by relaunching the migration and prepend a migration fixing data  (fixing error of type (2), see above)

Make sure to do (a), (b), and (c) in this order and in different commits.

zenovich:
Fixes #272


app/api/items/get_item.go
```diff
@@ -13,8 +13,9 @@ import (
 type itemStringCommon struct {
 	// required: true
 	LanguageID int64 `json:"language_id,string"`
+	// Nullable
```
smadbe: You haven't modified the add/update items where the title could be "null" as well then :-/ (even I don't really know what it means to have a `null` title)

db/migrations/18_do_not_mark_ancestors_as_todo_in_groups_groups_triggers.sql
```diff
@@ -1,12 +1,12 @@
 -- +migrate Up
 DROP TRIGGER `before_update_groups_groups`;
 -- +migrate StatementBegin
-/*!50003 CREATE*/ /*!50017 DEFINER=`algorea`@`%`*/ /*!50003 TRIGGER `before_update_groups_groups`
+/*!50003 CREATE*/ /*!50003 TRIGGER `before_update_groups_groups`
```
smadbe: Without `/*!50017` comment, probably it is useless to close the 50003 comment to reopen it directly, so it should probably be `/*!50003 CREATE TRIGGER ...`
BTW, I think all these comments are irrelevant in the case of our migrations. In my understanding, they are there to prevent error if we would apply this to a version of MySQL not supporting these, but in our case, we would prefer an error than ignoring the trigger silently. 
I think all comment marks should all be removed in our migrations (and probably in the initial schema as well)

app/api/items/get_item.go
smadbe: FYI, it would mean that there is no translation for the title in this language and that the default language should be used. (no impact for us)

zenovich: > You haven't modified the add/update items where the title could be "null" as well then :-/ (even I don't really know what it means to have a `null` title)

I haven't modified add/update intentionally because we don't allow setting titles to NULL there. Should we (I doubt we want to allow that)?

smadbe: Ok, valid reason :-)

PR #279, opened 2019-09-27
Rename datetime columns

### LINKED ISSUE #271: Fix attribute names of time attributes
Most "datetime" attributes in the db are called "xyz_date" while it is not just a date. 
I think "xyz_datetime" is not appropriate as the goal is not to put the type in it but just to have to have something semantically correct.  
IMO, "_time" is correct (even if it is not a mysql "time" as well), as a time semantically may be a precise date.
But in most of the time, the attribute naming should not require any suffix, "partial_access_from" would be more appropriate. 

BTW, for the durations, I would prefer having them in decimal seconds instead of using the MySQL time format.  

Let me know what you plan to do before starting.

zenovich:
Fixes #271


db/migrations/49_rename_datetime_columns.sql
```diff
@@ -0,0 +1,876 @@
+-- +migrate Up
+
+
+ALTER TABLE `filters`
+    RENAME COLUMN `start_date` TO `started_at`;
+ALTER TABLE `groups`
+    RENAME COLUMN `date_created` TO `created_at`,
+    RENAME COLUMN `code_end` TO `code_expires_at`,
```
smadbe: That's a much better name 😄

db/migrations/49_rename_datetime_columns.sql
```diff
@@ -0,0 +1,876 @@
...
+    RENAME COLUMN `validation_date` TO `validated_at`,
+    RENAME COLUMN `finish_date` TO `finished_at`,
+    RENAME COLUMN `last_activity_date` TO `last_activity_at`,
+    RENAME COLUMN `thread_start_date` TO `thread_started_at`,
+    RENAME COLUMN `best_answer_date` TO `best_answer_at`,
+    RENAME COLUMN `last_answer_date` TO `last_answer_at`,
+    RENAME COLUMN `last_hint_date` TO `last_hint_at`,
+    RENAME COLUMN `contest_start_date` TO `contest_started_at`;
+ALTER TABLE `groups_groups`
+    RENAME COLUMN `status_date` TO `status_changed_at`;
```
smadbe: Still not clear when I read the name, isn't it `latest_status_changed_at` ?

db/migrations/49_rename_datetime_columns.sql
```diff
@@ -0,0 +1,876 @@
...
+    RENAME COLUMN `expiration_date` TO `expires_at`,
+    RENAME INDEX `expiration_date` TO `expires_at`;
+ALTER TABLE `messages`
+    RENAME COLUMN `submission_date` TO `submitted_at`;
+ALTER TABLE `sessions`
+    RENAME COLUMN `expiration_date` TO `expires_at`,
+    RENAME COLUMN `issued_at_date` TO `issued_at`,
+    RENAME INDEX `expiration_date` TO `expires_at`;
+ALTER TABLE `threads`
+    RENAME COLUMN `last_activity_date` TO `last_activity_at`;
```
smadbe: `latest_activity_at` would be more correct

The same apply also to many other "last_"" attributes  (will not repeat)

smadbe:
Not directly a *date attribute, but `code_timer` would require a better name as well. `code_validity_duration` ? 🤔 Any better idea? (as a reminder, it is "How much time after the first use of the password, it will expire")

db/migrations/49_rename_datetime_columns.sql
zenovich: let's skip the 'latest_' prefix when possible, otherwise we would have a lot of them...

smadbe: Ok, but anyway "status" does not mean anything here. Either it is "type_changed_at" or just "changed_at" considering that it is quite implicit (as the other attributes are not really changed, we can consider it is)

zenovich: fixed (will be `type_changed_at`)

db/migrations/49_rename_datetime_columns.sql
zenovich: fixed

zenovich:
> Not directly a *date attribute, but `code_timer` would require a better name as well. `code_validity_duration` ? 🤔 Any better idea? (as a reminder, it is "How much time after the first use of the password, it will expire")

fixed (will be `code_lifetime`)

PR #288, opened 2019-10-07
Handle expiration for groups membership

### LINKED ISSUE #281: Handle expiration for group membership
Contest management will be handled through expiration on group memberships (cfr contest explanation).
So let's first handle expiration without contest.
Let's consider the existing services do not need to set or get this expiration, but membership is only considered as actual if the expiration is not past. 
Maybe to make this cleaner in regards of the existing statuses, it would be interesting to invent a new "contestEntered" (or just "system"?) membership, which would be the only one with this expiration for the moment.

zenovich:
Fixes #281


app/database/group_group_store_ancestors_integration_test.go
```diff
@@ -145,10 +158,103 @@ func TestGroupGroupStore_CreateNewAncestors_ProcessesOnlyDirectRelationsOrAccept
...
+	assert.NoError(t, groupGroupStore.Where("parent_group_id=1 AND child_group_id=3").
+		UpdateColumn("expires_at", "3019-10-20 14:13:55").Error())
+	assert.NoError(t, groupGroupStore.Where("parent_group_id=1 AND child_group_id=4").
+		UpdateColumn("expires_at", "3020-09-15 21:13:59").Error())
+	assert.NoError(t, groupGroupStore.Where("parent_group_id=2 AND child_group_id=3").
+		UpdateColumn("expires_at", "3020-10-20 10:23:40").Error())
+	assert.NoError(t, groupGroupStore.Where("parent_group_id=2 AND child_group_id=4").
+		UpdateColumn("expires_at", "3021-08-10 12:11:45").Error())
+	assert.NoError(t, groupGroupStore.Where("parent_group_id=3 AND child_group_id=4").
+		UpdateColumn("expires_at", "3022-05-11 17:43:24").Error())
```
smadbe: I don't get why you make vary the year, month, day, hour, min, sec. (anything to be validated really?). I would expect only one varying (for instance all different years and that's all) to be able to save 80% of time in trying to verify something here. 🙂
Or maybe I missed some interesting hidden logic 🤔 ... actually I don't care the day,  etc changes, but at least they should all be on different years so that we do not have to check year and month together in the mapping. 

But that looks correct at the end, I just have a headache now 😛

db/migrations/57_add_expires_at_into_groups_groups_and_groups_ancestors.sql
```diff
@@ -0,0 +1,128 @@
...
+            `child_descendants`.`ancestor_group_id` = OLD.`child_group_id`;
+
+        INSERT IGNORE INTO `groups_propagate` (id, ancestors_computation_state) VALUES (NEW.child_group_id, 'todo')
+        ON DUPLICATE KEY UPDATE `ancestors_computation_state` = 'todo';
+    END IF;
+END
+-- +migrate StatementEnd
+DROP TRIGGER `before_delete_groups_groups`;
+-- +migrate StatementBegin
+CREATE DEFINER=`algorea`@`%` TRIGGER `before_delete_groups_groups` BEFORE DELETE ON `groups_groups` FOR EACH ROW BEGIN
```
smadbe: 2 INSERT and 4 DELETE at each update/delete. 😲
Have you tried deciphering if everything is still needed? I didn't. (you can answer "no", that's just a question)

app/api/groups/get_team_descendants.go
```diff
@@ -102,10 +103,15 @@ func (srv *Service) getTeamDescendants(w http.ResponseWriter, r *http.Request) s
...
+			JOIN groups_groups AS parent_links
+			ON parent_links.parent_group_id = groups.id AND
+				parent_links.type = 'direct' AND
+				NOW() < parent_links.expires_at AND
+				parent_links.child_group_id IN (?)`, groupIDs).
 		Joins(`
-			JOIN groups_ancestors AS parent_ancestors ON parent_ancestors.child_group_id = groups.id AND
+			JOIN groups_ancestors AS parent_ancestors
+			ON parent_ancestors.child_group_id = groups.id AND
+				NOW() < parent_ancestors.expires_at AND
```
smadbe: When I see all these `NOW() < parent_ancestors.expires_at ` added, I wonder if it wouldn't have been wiser to create views `groups_ancestors_notexpired` and `groups_groups_notexpired` and to use them across the requests. 
These conditions and polluting sometimes-already-complex requests and could easily be forgotten and removed by mistake. 
There are no test updates for them :-/ and probably it would have been too much to add tons of tests but then actually a mistake would be so easy here.  I think when there are too many repetitive changes for the same things to justify tests (and because testings ancestors membership expiration would be out of scope of the function), maybe it is good place to hide the "logic" into something that could be considered as tested separately. I am afraid of mistakes on this expiration (not necessarily now but more regressions). 
What are your thoughts about that?

db/migrations/57_add_expires_at_into_groups_groups_and_groups_ancestors.sql
```diff
@@ -0,0 +1,128 @@
+-- +migrate Up
+ALTER TABLE `groups_groups` ADD COLUMN `expires_at` datetime  NOT NULL DEFAULT '9999-12-31 23:59:59'
+    COMMENT 'If set, the group membership expires at the specified time';
```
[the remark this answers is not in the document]
zenovich: fixed

app/database/group_group_store_ancestors_integration_test.go
zenovich: fixed

db/migrations/57_add_expires_at_into_groups_groups_and_groups_ancestors.sql
zenovich: As you can see there had been 3 INSERTs + 4 DELETEs at each update and 2 INSERTs + 4 DELETES at each deletion. I got rid of some unnecessary queries and only 3 INSERTs + 1 DELETE at each update and 2 INSERTs + 1 DELETE at each deletion remained.

app/api/groups/get_team_descendants.go
zenovich: Fixed. Let's see if it is better with the views...

app/database/group_group_store_ancestors_integration_test.go
smadbe: much better :-)

PR #289, opened 2019-10-08
Update expiration of current contestants in contestSetAdditionalTime

### LINKED ISSUE #282: Update

zenovich:
Fixes #282

### DIFF
--- a/app/api/contests/set_additional_time.go
+++ b/app/api/contests/set_additional_time.go
@@ -175,6 +175,7 @@ func setAdditionalTimeForGroupInContest(
 				IFNULL(SUM(TIME_TO_SEC(groups_contest_items.additional_time)), 0) AS total_additional_time`).
 			WithWriteLock().QueryExpr()).Error())
 
+	//nolint:gosec
 	result := store.Exec(`
 		UPDATE groups_groups
 		JOIN contest_participations


app/api/contests/set_additional_time.go
```diff
@@ -57,59 +61,132 @@ import (
...
+		service.MustNotBeError(scope.UpdateColumn("additional_time",
+			gorm.Expr("SEC_TO_TIME(?)", additionalTimeInSeconds)).Error())
+	} else if additionalTimeInSeconds != 0 {
+		service.MustNotBeError(groupContestItemStore.Exec(
+			"INSERT INTO groups_contest_items (group_id, item_id, additional_time) VALUES(?, ?, SEC_TO_TIME(?))",
+			groupID, itemID, additionalTimeInSeconds).Error())
+	}
+
+	service.MustNotBeError(store.Exec("DROP TEMPORARY TABLE IF EXISTS total_additional_times").Error())
+	service.MustNotBeError(store.Exec(`
```
smadbe: This request probably needs some explanations (comments), especially as there are many kind of groups and it is very hard to follow.
For instance, indicating after L157 that, at this point, we have all groups participating to the contest and which are descendant of the group on which we change additional time (so, that will be affected)

app/api/contests/set_additional_time.go
```diff
@@ -57,59 +61,132 @@ import (
...
+					ON groups_contest_items.group_id = groups_ancestors.ancestor_group_id AND
+						groups_contest_items.item_id = contest_participations.item_id`).
+			Group("groups_groups.child_group_id").
+			Select(`
+				groups_groups.child_group_id,
+				IFNULL(SUM(TIME_TO_SEC(groups_contest_items.additional_time)), 0) AS total_additional_time`).
+			WithWriteLock().QueryExpr()).Error())
+
+	result := store.Exec(`
+		UPDATE groups_groups
```
smadbe: If I do the UPDATE using the SELECT request directly inside it, it seems to works on my db. (so no needing the temp table) I just replaced `JOIN total_additional_times` by `JOIN (SELECT ... ) AS total_additional_times`
Could you check? 
If so, probably there are also ways to write it in a bit cleaner way.

zenovich: Do you have triggers in your DB? (Initially I implemented this with nested select, but that didn't work)

zenovich: ("Can't update table in stored function/trigger because it is already used by statement which invoked this stored function/trigger")

app/api/contests/set_additional_time.go
zenovich: fixed

app/api/contests/set_additional_time.go
smadbe: Yes, I have the triggers and applied the migrations, strange. 🤔 The tables are empty, but this error should not depend on the actual data.

zenovich: Just tried this again: ```Data: {"success":false,"message":"Internal Server Error","error_text":"Error 1442: Can't update table 'groups_ancestors' in stored function/trigger because it is already used by statement which invoked this stored function/trigger."}```

PR #290, opened 2019-10-09
Test migrations on realistic datasets and migration-undo

### LINKED ISSUE #285: Test migration undo's
Our migrations are kind-of tested by applying them on the CI before launching the tests.
Our down-migrations, while they are always written, are not run at all by the CI, so they may be completely erroneous without us noticing it. 
It may be very unpleasant to discover that in prod during a critical moment.
So we should include that in the CI testing process.

smadbe:
Test migrations on realistic datasets and down migrations.

Fixes #285 and more.


smadbe:
Thx for fixing the migration already :-) 
Some of the commit were made to be rebased (merged) as I was not able to run the same things locally. I can still rebase when it has been reviewed.

PR #292, opened 2019-10-10
Implement the contestEnter (aka contestStart) service

zenovich:
[contestEnter / contestStart](https://docs.google.com/document/d/15Xxmatrs_7my5ioriikL6YgOU4I0sAVcq5ySKinc0Nw/edit?pli=1#heading=h.dmh59tzemeq3)

### DIFF
--- a/app/api/contests/enter.go
+++ b/app/api/contests/enter.go
@@ -2,7 +2,6 @@ package contests
 
 import (
 	"net/http"
-	"time"
 
 	"github.com/go-chi/render"
 
@@ -104,7 +103,7 @@ func (srv *Service) enter(w http.ResponseWriter, r *http.Request) service.APIErr
 
 	service.MustNotBeError(render.Render(w, r, service.CreationSuccess(map[string]interface{}{
 		"duration":   itemInfo.Duration,
-		"entered_at": time.Time(*itemInfo.Now).Format(time.RFC3339),
+		"entered_at": itemInfo.Now,
 	})))
 	return service.NoError
 }
--- a/app/api/contests/enter.go
+++ b/app/api/contests/enter.go
@@ -61,7 +61,7 @@ func (srv *Service) enter(w http.ResponseWriter, r *http.Request) service.APIErr
 			return apiError.Error
 		}
 
-		if qualificationState.State != ready {
+		if qualificationState.State != string(ready) {
 			apiError = service.InsufficientAccessRightsError
 			return apiError.Error
 		}
--- a/app/api/contests/enter.go
+++ b/app/api/contests/enter.go
@@ -1,0 +1,110 @@
+package contests
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/go-chi/render"
+
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+// swagger:operation POST /contests/{item_id}/groups/{group_id} contests groups contestEnter
+// ---
+// summary: Enter the contest
+// description: >
+//                Allows to enter a contest as a group (user self or team).
+//
+//
+//                Restrictions:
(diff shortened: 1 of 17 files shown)


app/api/contests/enter.go
```diff
@@ -0,0 +1,110 @@
...
+
+		service.MustNotBeError(store.Items().ByID(qualificationState.itemID).
+			Select("NOW() AS now, items.duration, items.contest_participants_group_id").
+			WithWriteLock().Take(&itemInfo).Error())
+
+		service.MustNotBeError(store.Exec(`
+			INSERT INTO contest_participations (group_id, item_id, entered_at)
+			VALUES(?, ?, ?)`, qualificationState.groupID, qualificationState.itemID, itemInfo.Now).Error())
+
+		if itemInfo.ContestParticipantsGroupID != nil {
```
smadbe: I don't know what to think about the possibility of itemInfo.ContestParticipantsGroupID == nil.
I think we cannot allow that in our system, even if it does not destroy anything or give extra perm in practice. 
In order to be able to detect the problem, we should "log.warn" the case for the moment.

app/api/contests/enter.go
```diff
@@ -0,0 +1,110 @@
...
+			var totalAdditionalTime int64
+			service.MustNotBeError(store.ActiveGroupAncestors().
+				Where("groups_ancestors_active.child_group_id = ?", qualificationState.groupID).
+				Joins(`
+					LEFT JOIN groups_contest_items
+						ON groups_contest_items.group_id = groups_ancestors_active.ancestor_group_id AND
+							groups_contest_items.item_id = ?`, qualificationState.itemID).
```
smadbe: You should "join" instead of "left join", you do not care about the ancestors with no `groups_contest_items` .
Hum 🤔 maybe your PluckFirst may not return any row then? Couldn't we handle that properly?

app/api/contests/enter.go
```diff
@@ -0,0 +1,110 @@
...
+				Group("groups_ancestors_active.child_group_id").
+				WithWriteLock().
+				PluckFirst("IFNULL(SUM(TIME_TO_SEC(groups_contest_items.additional_time)), 0)", &totalAdditionalTime).
+				Error())
```
smadbe: Hey, it makes me think that we could not have "duration" attribute at all and put the "default duration" as "additional time" for the contest participant group! 😁

app/api/contests/enter.go
```diff
@@ -0,0 +1,110 @@
...
+	})
+
+	if apiError != service.NoError {
+		return apiError
+	}
+	service.MustNotBeError(err)
+
+	service.MustNotBeError(render.Render(w, r, service.CreationSuccess(map[string]interface{}{
+		"duration":   itemInfo.Duration,
+		"entered_at": time.Time(*itemInfo.Now).Format(time.RFC3339),
```
smadbe: Shouldn't be the render job to display the datetime in a right format?
🤔 Actually duration is also a time? 
So our backend has two ways to return a time in json and the service has to choose?
Maybe at least `time.Time(*...).Format(time.RFC3339)` should be wrapped in a `service` function so that the service do not handle it differently?

app/api/contests/enter.go
zenovich: Do you want to remove `items.duration`? I think it's helpful, don't you?

app/api/contests/enter.go
zenovich: fixed

app/api/contests/enter.go
smadbe: Conceptually implementing the duration of a contest by an "additional time" on the participant group is a bit weird, I cannot convince myself it is a very good idea. 

For instance, let's consider the interface to set "additional time" to a group of users who has not started the contest yet. The contest duration will be at "0" for them as they are not in the participant group yet.

PR #301, opened 2019-10-21
Remove column groups_attempts.minus_score

### LINKED ISSUE #283: Clean "groups_attempts.minus_score"
About the "groups_attempts.minus_score" attribute which was added to make index efficient:
- Is it still required in MySQL8? 
- Could we do it a generated column so that it is not forgotten when updating data? 
- This can probably even be a virtual column (not stored) as they can be used for indices.
- In any case, if not deleted, this column should have a comment in the db, I was not remembering what this column was about.

zenovich:
Fixes #283

### DIFF
--- a/app/api/groups/get_group_progress.go
+++ b/app/api/groups/get_group_progress.go
@@ -204,19 +204,19 @@ func (srv *Service) getGroupProgress(w http.ResponseWriter, r *http.Request) ser
 						TIMESTAMPDIFF(SECOND, MIN(started_at), MIN(validated_at)),
 						TIMESTAMPDIFF(SECOND, MIN(started_at), NOW())
 					)
-					FROM groups_attempts FORCE INDEX (group_item_minus_score_best_answer_date_id)
+					FROM groups_attempts
 					WHERE group_id = end_members.id AND item_id = items.id
 				)
 			) AS time_spent
 		FROM ? AS end_members`, endMembers.SubQuery()).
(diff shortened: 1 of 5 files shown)


app/api/groups/get_group_progress.go
```diff
@@ -204,19 +204,19 @@ func (srv *Service) getGroupProgress(w http.ResponseWriter, r *http.Request) ser
...
-			ON attempt_with_best_score.id = (
-				SELECT id FROM groups_attempts FORCE INDEX (group_item_minus_score_best_answer_date_id)
+			LEFT JOIN LATERAL (
+				SELECT score, validated, hints_cached, submissions_attempts, group_id
+				FROM groups_attempts
 				WHERE group_id = end_members.id AND item_id = items.id
-				ORDER BY group_id, item_id, minus_score, best_answer_at LIMIT 1
-			)`)
+				ORDER BY group_id, item_id, score DESC, best_answer_at LIMIT 1
+			) AS attempt_with_best_score ON 1`)
```
smadbe: What the benefit here? You are not using the attribute from previous queries the left join was anyway limiting every to 1 row.

zenovich: `WHERE group_id = end_members.id AND item_id = items.id` uses two attributes from previous queries. There were two joins actually (LEFT JOIN + the nested SELECT), now it is just one LEFT JOIN LATERAL.

PR #304, opened 2019-10-26
Increase transactionDelayBetweenRetries to 1 second to make concurrent transactions more reliable

zenovich:


zenovich:
> What can be the consequence in prod?

It will be more robust in the situatiation where there are many concurrent transactions causing deadlocks.

smadbe:
But it gives you something very slow for a web service. If a service is going to sleep for seconds, it may be better to abort it and let the client retries.
So you sometimes get deadlock (or just long locks?)? Very problematic.

zenovich:
> But it gives you something very slow for a web service. If a service is going to sleep for seconds, it may be better to abort it and let the client retries.

It is up to (transactionRetriesLimit = 30) * (transactionDelayBetweenRetries = 1s) = 30s. We can make transactionRetriesLimit twice lower.

zenovich:
> So you sometimes get deadlock (or just long locks?)? Very problematic.

I get deadlocks, but it is not very problematic. It is normal in MySQL and retrying is the recommended way of dealing with that. The 100ms delay is too small (at least on my computer), I often get 30 deadlocks in a row in unit tests that check concurrency. At the same time, the 1s delay works great.

smadbe:
Come on, deadlocks are not perfectly normal 🤨, they are due to mistakes. 
Anyway, I understand that we cannot avoid any bugs and so deadlocks, so maybe not "very problematic".
My point was that for waiting 1sec in a web service, it might be more interesting to just abort it. It is not about running the tests but for the final product. 
We can continue this discussion on Slack maybe.

zenovich:
"In a transaction system, deadlocks are a fact of life and not completely avoidable." (see https://www.percona.com/blog/2014/10/28/how-to-deal-with-mysql-deadlocks/)

zenovich:
1s delay doesn't seem to be too much for services that modify data in the DB. Other services highly unlikely get into deadlocks.

PR #322, opened 2019-11-16
Set items_items.*_propagation fields correctly in itemCreate & itemEdit

zenovich:


app/database/permission_generated_store_test.go
```diff
@@ -18,7 +18,11 @@ func TestPermissionGeneratedStore_AccessRightsForItemsVisibleToUser(t *testing.T
 	defer clearAllPermissionEnums()
 
 	mock.ExpectQuery("^"+regexp.QuoteMeta(
-		"SELECT item_id, MAX(can_view_generated_value) AS can_view_generated_value "+
+		"SELECT item_id, MAX(can_view_generated_value) AS can_view_generated_value, "+
```
smadbe: I am wondering: Are these kind of tests are really useful?  🤔
These are copy from the code (not dry), if there is a mistake it is will be there anyway as it is a copy paste, and make changes "heavy". That's nice, but isn't it a case where too many tests may harm testing? (just a question)

What is the benefit of to put the full query into the test? To check that the query has been run? Are the other tests on the results enough for that?

app/api/items/items.go
```diff
@@ -59,3 +60,87 @@ func checkHintOrScoreTokenRequiredFields(user *database.User, taskToken *token.T
...
+	ChildItemID                int64
+	Order                      int32
+	ContentViewPropagation     string
+	UpperViewLevelsPropagation string
+	GrantViewPropagation       bool
+	WatchPropagation           bool
+	EditPropagation            bool
+}
+
+func constructItemsItemsForChildren(childrenPermissions []permission, children []itemChild,
```
smadbe: Even if private, most non-trivial function should a small spec / comment describing what it does.

app/database/permission_generated_store_test.go
zenovich: These are not copied from the code: the code looks like
```
	return s.
		Select(`
			item_id,
			MAX(can_view_generated_value) AS can_view_generated_value`).
		Joins(`
			JOIN (
				SELECT * FROM groups_ancestors_active
				WHERE groups_ancestors_active.child_group_id = ?
			) AS ancestors
			ON ancestors.ancestor_group_id = permissions.group_id`, groupID).
		WherePermissionIsAtLeast(permissionKind, neededPermission).
		Group("permissions.item_id")
```

zenovich: The benefit of putting the whole query is to check that the query is completely correct and there are no missing parts.

zenovich: > That's nice, but isn't it a case where too many tests may harm testing? (just a question)

Yes, probably it is the case. Looks like this one http://xunitpatterns.com/Fragile%20Test.html#Overspecified%20Software

app/api/items/items.go
zenovich: added comments

app/database/permission_generated_store_test.go
smadbe: "the query is completely correct and there are no missing parts." -> if we assume we are not testing gorm, there is not really to check that there is nothing missing.
Of course, that is for the case where there is no query composition from what you get from functions. In such cases, it may still make sense to check that the end result is what you expect. 
IMO, using `Select`, `Joins`, ... functions should not be different from directly the SQL string in terms of testing.

Ok when I'll have time, I'll check what we could do... 
(so not blocking for this PR, was just a question :-))

app/api/items/items.go
smadbe: 👍 
Try, by default, to add a bit more comments in the code (not especially code spec but a few words of explanation for a non-trivial block of code).

PR #324, opened 2019-11-20
Do not add the 'id' column implicitly either into the list of accepted fields or into the sorting spec in service.ApplySortingAndPaging() + provide a way to prohibit setting the 'sort' parameter in some services

### LINKED ISSUE #323: Do not add the 'id' column implicitly either into the list of accepted fields or into the sorting spec  in service.ApplySortingAndPaging()
Additionally provide a way to prohibit setting the 'sort' parameter in some services

zenovich:
Fixes #323

### DIFF
--- a/app/service/sorting.go
+++ b/app/service/sorting.go
@@ -43,7 +43,8 @@ func (d sortingDirection) conditionSign() string {
 }
 
 // ApplySortingAndPaging applies ordering and paging according to given accepted fields and sorting rules
-// taking into the account the URL parameters 'from.*'
+// taking into the account the URL parameters 'from.*'.
+// When the `skipSortParameter` is true, the 'sort' request parameter is ignored.
 func ApplySortingAndPaging(r *http.Request, query *database.DB, acceptedFields map[string]*FieldSortingParams,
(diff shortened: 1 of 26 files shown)


app/service/sorting.go
```diff
@@ -42,13 +42,11 @@ func (d sortingDirection) conditionSign() string {
...
 }
 
-const idFieldName = "id"
-
 // ApplySortingAndPaging applies ordering and paging according to given accepted fields and sorting rules
 // taking into the account the URL parameters 'from.*'
 func ApplySortingAndPaging(r *http.Request, query *database.DB, acceptedFields map[string]*FieldSortingParams,
-	defaultRules string) (*database.DB, APIError) {
-	sortingRules, acceptedFields := prepareSortingRulesAndAcceptedFields(r, acceptedFields, defaultRules)
+	defaultRules string, skipSortParameter bool) (*database.DB, APIError) {
```
smadbe: Reading the definition + spec of the function, I cannot understand what this `skipSortParameter`is for.

zenovich: added the comment

PR #325, opened 2019-11-21
Clean some groups_attempts attributes

smadbe:
Clean some attributes of `groups_attempts`: remove some obsolete ones, rename others.


docker-compose.yml
```diff
@@ -8,6 +8,8 @@ services:
     ports:
       - "3306:3306"
       - "33060:33060"
+    security_opt:
```
zenovich: Why is this?

db/migrations/66_clean_groups_attempts.sql
```diff
@@ -8,7 +8,10 @@ ALTER TABLE `groups_attempts`
   DROP COLUMN `thread_started_at`,
   DROP COLUMN `all_lang_prog`,
 
-  RENAME COLUMN `submissions_attempts` TO `submissions`;
+  RENAME COLUMN `submissions_attempts` TO `submissions`,
+
+  -- rename and change comment
+  CHANGE COLUMN `key_obtained` `unlocks_items` tinyint(1) NOT NULL DEFAULT '0' COMMENT 'Whether the score of this attempts allows unlocking other items (score >= items.score_min_unlock)';
```
zenovich: I think `key_obtained` is good enough, but, if you don't like it, it probably should be something in the past tense, `unlocked_items` maybe. But still as I see it, `unlocked_items` should list items while this column does not. Maybe `obtained_key` would be better?

docker-compose.yml
smadbe: Hum probably a tab fix by the IDE. I'll remove it from the commit.

db/migrations/66_clean_groups_attempts.sql
smadbe: The initial idea is of course `unlocked_items` but the problem is that in English it means `items which are unlocked`. Here I mean that the attempt is current in a state which "unlocks items". This two reasons explains the present tense but I am open to a better name :-).
("key obtained" does not make sense IMO (Michel raised this as well), there is no concept of "key" here except that if there is a lock there is a key ;-).)

zenovich: Yes, 'if there is a lock there is a key' is exactly how I understand the current name. Some other possible names could be `unlocked_next_tasks`, `unlocked_subitems`, `achievement_unlocked` :)

smadbe: It is not really "next tasks", nor "subitems", not even achievements, as it could be even "hidden" from the task owner (it will be moved from `items` to a mapping table soon). Not a dependency as well. :-/
So what we want to express is "this attempt has obtained a score high enough to unlock some other items"... that is why I still find "unlocks_items" (i.e. "does this attempt unlock other items"), but I agree that the present tense is not great. `unlocking_items` is not much better I suppose? 🤔

zenovich: `goal_achieved` maybe?

smadbe: But while the related attribute is called `items.unlocked_item_ids`, I think it has to use a very similar name, without any interpretation such as a lock needs a key or whatever. "unlock" should appear in it, without doubt.

zenovich: Okay, let's at least call it `unlocked_items` then?

zenovich: The present tense leads to much more confusion (IMO)

smadbe: After slack discussion, let's go on `has_unlocked_items`

smadbe: Fixed

docker-compose.yml
smadbe: Fixed

PR #326, opened 2019-11-21
Fix groups_attempts.validated

smadbe:
`validated` and `validated_at` are supposed to be related values, in practice it was not the case. So, 
-  fix propagation which was putting `validated_at` for chapters even when there are not validated
- make `validated` computed from `validated_at` instead of independents values

After spending some time on it, I really cannot fix the unit tests of propagation in a reasonable amount of time 😔 (figuring out what you test exactly in each test, the machinery, the idea behind each dataset combination). @zenovich , could you take over this PR? So check what I've done and fix the propagation tests as they should (it should be quite fresh in your head).


smadbe:
👍

zenovich:
May I merge? @smadbe

smadbe:
Yes :-)

PR #337, opened 2019-12-03
Add columns related to approvals

### LINKED ISSUE #334: Group permissions, step 6: Approvals (partial support)
The end goal will be to implement the group permissions as described in https://france-ioi.github.io/algorea-devdoc/design/access-rights/groups/

Step 6: Approvals (partial support)
Description: Add everything in the DB to handle approvals. Do not change existing services so far.

Subtasks:
* add to `groups`:
  - require_personal_info_access_approval: enum(none, view, edit) (default: none)
  - require_lock_membership_approval_until (datetime) (default: null)
  - require_watch_approval (bool, default: null)
* add to `groups_groups`:
  - personal_info_access_approved_at (datetime) + personal_info_access_approved (computed `personal_info_access_approved_at IS NOT NULL`)
  - lock_membership_approved_at (datetime) + lock_membership_approved (computed)
  - watch_approved_at (datetime) + watch_approved (computed)
* add to `group_pending_requests`: (for"join requests)
  - personal_info_access_approved_at
  - lock_membership_approved_at
  - watch_approved_at
* add to `groups_managers`
  - can_edit_personal_info (bool, default: false)

zenovich:
Fixes #334

### DIFF
--- a/db/migrations/72_add_approval_related_columns.sql
+++ b/db/migrations/72_add_approval_related_columns.sql
@@ -1,0 +1,54 @@
+-- +migrate Up
+ALTER TABLE `groups`
+    ADD COLUMN `require_personal_info_access_approval` ENUM('none', 'view', 'edit') NOT NULL DEFAULT 'none'
+        AFTER `lock_user_deletion_until`,
+    ADD COLUMN `require_lock_membership_approval_until` DATETIME DEFAULT NULL
+        AFTER `require_personal_info_access_approval`,
+    ADD COLUMN `require_watch_approval` TINYINT(1) NOT NULL DEFAULT 0
+        AFTER `require_lock_membership_approval_until`;
+
+ALTER TABLE `groups_groups`
+    ADD COLUMN `personal_info_access_approved_at` DATETIME DEFAULT NULL AFTER `expires_at`,
+    ADD COLUMN `personal_info_access_approved` TINYINT(1)
+        AS (IFNULL(`personal_info_access_approved_at`, 0)) NOT NULL AFTER `personal_info_access_approved_at`,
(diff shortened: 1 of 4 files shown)


db/migrations/72_add_approval_related_columns.sql
```diff
@@ -0,0 +1,56 @@
...
+        AFTER `lock_user_deletion_until`,
+    ADD COLUMN `require_lock_membership_approval_until` DATETIME DEFAULT NULL
+        AFTER `require_personal_info_access_approval`,
+    ADD COLUMN `require_watch_approval` TINYINT(1) NOT NULL DEFAULT 0
+        AFTER `require_lock_membership_approval_until`;
+
+ALTER TABLE `groups_groups`
+    ADD COLUMN `personal_info_access_approved_at` DATETIME DEFAULT NULL AFTER `expires_at`,
+    ADD COLUMN `personal_info_access_approved` TINYINT(1)
+        AS (IFNULL(`personal_info_access_approved_at`, 0)) NOT NULL
```
smadbe: Are you sure `(IFNULL(personal_info_access_approved_at, 0))` is a better to compute `(personal_info_access_approved_at IS NOT NULL)`?
For me this looks like writing `if x == true return true; else return false` instead of `return x` directly.
Here it even has to cast the date to a int if not null... so it would return false even if it is not null but a date which equals "0", which I would expect to return TRUE. Probably it cannot happen anyway ;-) but just to say that it looks less readable and even "a bit less correct" (whatever it means;-)).

Ah there is even a `ISNULL()` fct in MySQL, didn't you want to that one? `NOT ISNULL(personal_info_access_approved_at)`

Maybe I just missed the reason why you did that.

zenovich: fixed

PR #340, opened 2019-12-04
update groups_groups table comment

smadbe:
Just a fix of the comment of `groups_`groups`.


smadbe:
@zenovich , could you review?
The codecov warning is just because there were other tests added to master in the meantime.

PR #349, opened 2019-12-06
Create contest participants group on item creation/editing + add swagger docs for itemEdit

### LINKED ISSUE #297: Create contest participant group at item creation/edition
When an item is created with a duration, create the participant group (with a new type "ContestParticipants", not allowed as input type from these service) as `contest_participants_group_id`. 
Give this group the partial access on the item.

On edit: 
* If a duration is added and `contest_participants_group_id` is NULL, do the same.
* If the duration is set to null, do nothing.  (keep the group)

Update the comment of items.contest_participants_group_id to "Group to which all the contest participants (users or teams) belong. Must not be null for a contest."

zenovich:
Fixes #297

### DIFF
--- a/app/api/items/update_item.go
+++ b/app/api/items/update_item.go
@@ -2,10 +2,12 @@ package items
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/go-chi/render"
+	"github.com/jinzhu/gorm"
 
 	"github.com/France-ioi/validator"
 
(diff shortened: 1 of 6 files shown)


app/api/items/update_item.go
```diff
@@ -69,18 +118,40 @@ func (srv *Service) updateItem(w http.ResponseWriter, r *http.Request) service.A
...
+
+		itemData := formData.ConstructPartialMapForDB("itemWithDefaultLanguageIDAndOptionalType")
+
+		if itemData["duration"] != nil && participantsGroupID == nil {
+			service.MustNotBeError(store.RetryOnDuplicatePrimaryKeyError(func(s *database.DataStore) error {
+				participantsGroupIDTmp := s.NewID()
+				participantsGroupID = &participantsGroupIDTmp
+				service.MustNotBeError(s.Groups().InsertMap(map[string]interface{}{
+					"id": participantsGroupID, "type": "ContestParticipants",
+					"name": fmt.Sprintf("%d-participants", itemID),
```
smadbe: This part of the code is repeated and might repeat at other places as well? (when starting a contest and there is no group for it, what happens again?) 
Should be most to the store probably (the test can be minimalist)

zenovich: > when starting a contest and there is no group for it

zenovich: Fixed. There were reasons to keep it inside app.api.items:
1) The function is probably not supposed to be used outside the package.
2) Moving it into another package would require us to implement cross-package error handling.
3) Also, if moving it into another package, we would have to write many tests that would check things already tested by scenarios + tests of error handling.

db/migrations/58_create_column_items_contest_participants_group_id.sql
```diff
@@ -1,6 +1,6 @@
 -- +migrate Up
 ALTER TABLE `items` ADD COLUMN `contest_participants_group_id` bigint(20) DEFAULT NULL
-    COMMENT 'id of the group to which all the contest participants (users or teams) belong.'
+    COMMENT 'Group to which all the contest participants (users or teams) belong. Must not be null for a contest.'
```
smadbe: 👍

PR #350, opened 2019-12-07
Include scores into attempt propagation + get rid of unlocking in saveGrade

### LINKED ISSUE #348: Include scores in attempt propagation
* Add a new attribute `items_items.weight` INT NOT NULL DEFAULT '1' COMMENT 'Weight of this child in his parent's score computation'.
* In attempt propagation, propagate scores to parents by computing the weighted average among all its children.
* Instead of unlocking by collecting first the items and then unlocking them "id IN (...)", do the unlock directly after each score update request for all processed attempts
* Don't do anything related to unlocking in saveGrade anymore
* Remove `has_unlocked_items` from saveGrade service output

zenovich:
Fixes #348


app/database/group_attempt_store_compute_all_group_attempts.go
```diff
@@ -160,24 +149,62 @@ func (s *GroupAttemptStore) ComputeAllGroupAttempts() (err error) {
...
+							'content',
+							NOW()
+						FROM groups_attempts
+						JOIN item_unlocking_rules ON item_unlocking_rules.unlocking_item_id = groups_attempts.item_id AND
+							item_unlocking_rules.score <= groups_attempts.score
+						JOIN ` + "`groups`" + ` ON groups_attempts.group_id = groups.id
+						WHERE groups_attempts.ancestors_computation_state = 'processing'
+					ON DUPLICATE KEY UPDATE
+						latest_update_on = IF(can_view = 'content', latest_update_on, NOW()),
+						can_view = GREATEST(can_view_value, 3 /* 'content' */)`
```
smadbe: As soon you'll have introduce the "origin" attribute, we know that it cannot be > than "can_view:content" so you can just ignore on duplicated. Maybe you should already do it this way directly.

app/database/group_attempt_store_compute_all_group_attempts.go
```diff
@@ -111,7 +111,8 @@ func (s *GroupAttemptStore) ComputeAllGroupAttempts() (err error) {
 								AS children_non_validated_categories,
 							MAX(aggregated_children_attempts.validated_at) AS max_validated_at,
 							MAX(IF(items_items.category = 'Validation', aggregated_children_attempts.validated_at, NULL))
-								AS max_validated_at_categories
+								AS max_validated_at_categories,
+							IFNULL(AVG(aggregated_children_attempts.score * items_items.weight), 0) AS average_score
```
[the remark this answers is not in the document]
zenovich: It will not be null actually, but it still will be wrong:
```
mysql> select avg(score) from (select score from groups_attempts union select null) as scores;
+------------+
| avg(score) |
+------------+
|         20 |
+------------+
1 row in set (0.00 sec)

mysql> select score from groups_attempts;                                       +-------+
| score |
+-------+
|    10 |
|    20 |
|    30 |
+-------+
3 rows in set (0.00 sec)
```

db/migrations/75_create_columns_items_items_weight.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+ALTER TABLE `items_items`
+    ADD COLUMN `weight` FLOAT NOT NULL DEFAULT '1' COMMENT 'Weight of this child in his parent\'s score computation'
```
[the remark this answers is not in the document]
zenovich: fixed both

app/database/group_attempt_store_compute_all_group_attempts.go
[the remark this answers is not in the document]
zenovich: fixed

app/database/group_attempt_store_compute_all_group_attempts.go
```diff
@@ -157,6 +159,8 @@ func (s *GroupAttemptStore) ComputeAllGroupAttempts() (err error) {
 							WHEN items.validation_type = 'One' AND children_stats.children_validated > 0 THEN children_stats.max_validated_at
 							ELSE NULL
 							END),
+						target_groups_attempts.score = IF(children_stats.id IS NOT NULL,
+							children_stats.average_score, target_groups_attempts.score),
```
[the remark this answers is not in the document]
zenovich: or there are no attempts, or because of 'no_score'

app/database/group_attempt_store_compute_all_group_attempts.go
zenovich: I'd prefer fixing it later together with introducing the 'origin' attribute.

app/database/group_attempt_store_compute_all_group_attempts.go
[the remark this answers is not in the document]
smadbe: Ok, then reset the score to "0" if there are no children (anymore), that's safer.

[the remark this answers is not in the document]
zenovich: This way we will wipe all the scores for tasks (since they don't have children items)

zenovich:
> Hum it seems you haven't understood correctly the concept of weight here (my bad, I didn't explain it, I thought there was a single interpretation of it).
> The weight is a INT (not a tiny) which may have any positive value.
> You compute the chapter score by `SUM(score_i * weight*i) / SUM(weight_i)`.
> So it means that if you have 2 subtasks, one with a weight 6 and a score of 10/100, and one with a weight 1 and a score of 80/100, you get `(10*6+80*1)/7 = 20/100` for the chapter. The calculation must make sure that if the total weight is "0", it does divide by 0. Can put 0 instead (btw, it is strange not to allow "null" for scores)
> Actually, you should even declare it `INT UNSIGNED`

fixed

app/database/group_attempt_store_compute_all_group_attempts.go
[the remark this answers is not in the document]
smadbe: Right... that's annoying 🤔   We have to find a solution for that.

BTW, you should not put score (so put 0) if "items.no_score"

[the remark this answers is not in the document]
zenovich: > BTW, you should not put score (so put 0) if "items.no_score"

done

app/database/group_attempt_store_compute_all_group_attempts.go
```diff
@@ -101,7 +101,8 @@ func (s *GroupAttemptStore) ComputeAllGroupAttempts() (err error) {
 							MAX(aggregated_children_attempts.validated_at) AS max_validated_at,
 							MAX(IF(items_items.category = 'Validation', aggregated_children_attempts.validated_at, NULL))
 								AS max_validated_at_categories,
-							IFNULL(AVG(IFNULL(aggregated_children_attempts.score, 0) * items_items.score_weight), 0) AS average_score
+							IFNULL(SUM(IFNULL(aggregated_children_attempts.score, 0) * items_items.score_weight), 0) /
```
smadbe: I do not get the outer "IFNULL". `IFNULL(aggregated_children_attempts.score, 0)` cannot be null, `items_items.score_weight`cannot be null, so the sum of the product cannot be null.

app/database/group_attempt_store_compute_all_group_attempts.go
```diff
@@ -101,7 +101,8 @@ func (s *GroupAttemptStore) ComputeAllGroupAttempts() (err error) {
...
+								COALESCE(NULLIF(IFNULL(SUM(items_items.score_weight), 1), 0), 1) AS average_score
```
smadbe: Same here, how can `SUM(items_items.score_weight)` be null with a `FROM items_items` and a score_weight which cannot be null?
Then, why does the `COALESCE` set "1/100" as score if the sum of weight is 0? Where does "1" come from?

app/database/group_attempt_store_compute_all_group_attempts.go
zenovich: aggregated_children_attempts.score is null when there are no attempts for the item

app/database/group_attempt_store_compute_all_group_attempts.go
zenovich: When the sum of weights is 0, the nominator is 0 too (because this means all the weights are zeros since weights cannot be negative). In this situation we use as the denominator to prevent division by 0.

app/database/group_attempt_store_compute_all_group_attempts.go
smadbe: "the **outer** "IFNULL"

app/database/group_attempt_store_compute_all_group_attempts.go
smadbe: Ok so you compute "0/1" which equals "0". Why not... ok.

There were 2 questions. The first one is unanswered:
"how can SUM(items_items.score_weight) be null with a FROM items_items and a score_weight which cannot be null?"

zenovich: SUM(items_items.score_weight) is null when there are no items_items matching the criteria.

app/database/group_attempt_store_compute_all_group_attempts.go
zenovich: fixed

app/database/group_attempt_store_compute_all_group_attempts.go
smadbe: How can it be when the query is on items_items?

zenovich:
```

Comment (zenovich) [reply app/database/group_attempt_store_compute_all_group_attempts.go]:
```

smadbe:
```
mysql> select sum(id) from items_items group by parent_item_id;
Empty set (0.00 sec)
```

zenovich: ok. fixed

app/database/group_attempt_store_compute_all_group_attempts.go
```diff
@@ -102,7 +102,7 @@ func (s *GroupAttemptStore) ComputeAllGroupAttempts() (err error) {
 							MAX(IF(items_items.category = 'Validation', aggregated_children_attempts.validated_at, NULL))
 								AS max_validated_at_categories,
 							SUM(IFNULL(aggregated_children_attempts.score, 0) * items_items.score_weight) /
-								COALESCE(NULLIF(IFNULL(SUM(items_items.score_weight), 0), 0), 1) AS average_score
+								COALESCE(NULLIF(SUM(items_items.score_weight), 0), 1) AS average_score
```
smadbe: usually you use `IFNULL` for doing `COALESCE`

zenovich: IFNULL and NULLIF are not the same thing

smadbe: I didn't say that. I say that `IFNULL` and `COALESCE` are the same thing. And at other places, you prefer using `IFNULL` for that use, so it is not consistent with what you do usually.

zenovich: Do you prefer 
```
IFNULL(NULLIF(SUM(items_items.score_weight), 0), 1)
```
to
```
COALESCE(NULLIF(SUM(items_items.score_weight), 0), 1)
```
?

smadbe: It is not a matter of preference, I just say that you usually use `IFNULL` for this usage. 
(and you did `IFNULL(NULLIF(` in the past already).
You decide.

zenovich: In this particulare case, for me, 'COALESCE(NULLIF' is more readable than 'IFNULL(NULLIF'

PR #364, opened 2019-12-17
Set missing users_answers.attempt_id where possible, make it NOT NULL, drop users_answers.item_id

### LINKED ISSUE #360: Attempts <-> Answers link
Clean the current data for the link between the `groups_attempts` and `users_answers`. 
All answers should match an attempt. 

For the cleaning: (just for info)
- all team answers were already with a attempt_id set in the initial db
- most of individual users' anwers don't.  But it is easy to match the answer to the attempt using the group/user and item.  
- Many answers do not match any attempt in any way but apparently that is due to a wrong cleaning in the past.

So in practice:
- for answers with attempt_id IS NULL and matching an attempt on attempt.group_id = answers.user_id and same item_id -> set the attempt.id in answers.attempt_id.
- delete all remaining answers with still no attempt_id (495k)
- make answers.attempt_id not nullable
- create FK from  answers.attempt_id  to attempt.id
- remove answers.item_id which becomes redundant with attempt_id

zenovich:
Fixes #360

### DIFF
--- a/db/migrations/77_make_users_answers_attempt_id_not_null.sql
+++ b/db/migrations/77_make_users_answers_attempt_id_not_null.sql
@@ -1,0 +1,51 @@
+-- +migrate Up
+
+/* 2000897 rows */
+UPDATE `users_answers`
+    JOIN `groups_attempts` ON `groups_attempts`.`group_id` = `users_answers`.`user_id` AND
+                              `groups_attempts`.item_id = `users_answers`.`item_id`
+SET `users_answers`.`attempt_id` = `groups_attempts`.`id`
+WHERE `users_answers`.`attempt_id` IS NULL;
+
+/* 35548 rows */
+UPDATE `users_answers`
+    JOIN `users_items` ON `users_items`.`user_id` = `users_answers`.`user_id` AND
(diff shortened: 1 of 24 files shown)


db/migrations/77_make_users_answers_attempt_id_not_null.sql
```diff
@@ -0,0 +1,51 @@
...
+    DROP COLUMN `item_id`,
+    ADD CONSTRAINT `fk_users_answers_attempt_id_groups_attempts_id`
+        FOREIGN KEY (`attempt_id`) REFERENCES `groups_attempts`(`id`) ON DELETE CASCADE;
+
+-- +migrate Down
+ALTER TABLE `users_answers`
+    DROP FOREIGN KEY `fk_users_answers_attempt_id_groups_attempts_id`,
+    ADD COLUMN `item_id` bigint(20) NOT NULL AFTER `user_id`,
+    ADD INDEX `item_id` (`item_id`),
+    MODIFY COLUMN `attempt_id` bigint(20) NOT NULL;
```
smadbe: ~~NOT~~ NULL

zenovich: fixed

PR #372, opened 2019-12-26
Update services using "has_attempts" so that "has_attempts" doesn't mean "only as a team" anymore

### LINKED ISSUE #368: Update attempts services using "has_attempts"
### GET /items/{item_id}/attempts -  itemAttemptsView
Take an optional `as_team_id` as input instead of relying on "has_attempts". 
If as_team_id is provided, this must be a team and the current user must be a member of it. 
If not provided, use the current user id as group_id. 
This group (team or user) should have at least 'content' permission on the task item, otherwise forbidden is returned.

### GET /items/{item_id}/task-token -  itemTaskTokenGet
Change the url to `GET /attempts/{attempt_id}/task-token` and remove all the part about attempts fetch/creation.
Just update the attempts:
* set `started_at` to now if null
* set `latest_activity_at` to now 
"the user should have at least 'content' access to the item" -> "the group of the attempt should have at least 'content' access to the item"

zenovich:
1. Introduce a new optional parameter 'as_team_id' in itemAttemptsView.
2. Get tokens by 'attempt_id' instead of 'item_id' in itemTaskTokenGet.

Fixes #368


app/api/items/get_attempts.go
```diff
@@ -39,18 +40,21 @@ type itemAttemptsViewResponseRow struct {
...
-// description: Returns attempts made by the current user (if `items.has_attempts` = 0) or his
-//              teams (if `items.has_attempts` = 1) while solving the task given in `item_id`.
+// description: Returns attempts made by the current user (if `as_team_id` is not given) or
+//              `as_team_id` (otherwise) while solving the task given in `item_id`.
 //
 //
-//              The current user should have at least 'content' permission on the task item,
-//              otherwise an empty list is returned.
+//              The current user (if no `as_team_id` given) or `as_team_id` (otherwise) should have
+//              at least 'content' permission on the task item, the 'forbidden' error is returned.
```
smadbe: .+ the user must be member of the team if `as_team_id` is provided

app/api/items/get_task_token.go
```diff
@@ -13,38 +12,34 @@ import (
...
-//   Get a task token with a refreshed active attempt.
+//   Get a task token with the refreshed active attempt.
 //
 //
 //   * If there is no row for the current user and the given item in `users_items`, the service creates one.
 //
 //   * If the active attempt (`active_attempt_id`) is not set in the `users_items` for the item and the user,
-//   the service chooses the most recent one among all the user's attempts (or the team's attempts if
-//   `items.has_attempts`=1)  for the given item. If no attempts found, the new one gets created and chosen as active.
+//   the service makes `attempt_id` active.
```
smadbe: At the end, we discussed with Michel that it is better that the backend does not auto-switch the active attempt, it allows the user to have 2 windows open on two different attempts without having the active attempt switch all the time.
(update on L25, L19, L29, code and tests)

app/api/items/get_attempts.go
zenovich: added the note

app/api/items/get_task_token.go
zenovich: fixed

PR #375, opened 2019-12-26
groupsMembersView: only display names if got user approval

### LINKED ISSUE #356: Member list: only display firstname/lastname if got user approval
In groupsMembersView, only display the firstname and lastname if got user approval, i.e. if there exists at least one group the auth user is manager of (so directly or from its ancestors) and on which the user has "personal_info_access_approved" (not necessarily the same group) as the request one.

zenovich:
Fixes #356

### DIFF
--- a/app/api/groups/get_members.go
+++ b/app/api/groups/get_members.go
@@ -118,39 +118,37 @@ func (srv *Service) getMembers(w http.ResponseWriter, r *http.Request) service.A
 		return apiError
 	}
 
-	query := srv.Store.Raw(`
-		WITH managed_groups_with_approval AS ? ?`,
-		srv.Store.ActiveGroupAncestors().ManagedByUser(user).
-			Joins(`
-				JOIN groups_groups_active
-					ON groups_groups_active.child_group_id = groups_ancestors_active.child_group_id AND
-						 groups_groups_active.personal_info_view_approved`).
-			Select("groups_groups_active.child_group_id AS id").
-			Group("groups_groups_active.child_group_id").
-			SubQuery(),
-		srv.Store.GroupGroups().
-			Select(`
-				groups_groups.id,
-				latest_change.at AS member_since,
-				latest_change.action,
-				users.group_id AS user__group_id,
-				users.login AS user__login,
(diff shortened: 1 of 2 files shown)


app/api/groups/get_members.go
```diff
@@ -117,27 +118,39 @@ func (srv *Service) getMembers(w http.ResponseWriter, r *http.Request) service.A
...
+				groups_groups.id,
+				latest_change.at AS member_since,
+				latest_change.action,
+				users.group_id AS user__group_id,
+				users.login AS user__login,
+				IF(managed_groups_with_approval.id IS NOT NULL, users.first_name, NULL) AS user__first_name,
+				IF(managed_groups_with_approval.id IS NOT NULL, users.last_name, NULL) AS user__last_name,
+				users.grade AS user__grade`).
+			Joins("LEFT JOIN users ON users.group_id = groups_groups.child_group_id").
+			Joins("LEFT JOIN managed_groups_with_approval ON managed_groups_with_approval.id = users.group_id").
```
smadbe: you cannot use the query directly instead of doing the "WITH"?

app/api/groups/get_members.go
```diff
@@ -117,27 +118,39 @@ func (srv *Service) getMembers(w http.ResponseWriter, r *http.Request) service.A
...
-				LIMIT 1
-			) AS latest_change ON 1`).
-		WhereGroupRelationIsActual().
-		Where("groups_groups.parent_group_id = ?", groupID)
+	query := srv.Store.Raw(`
+		WITH managed_groups_with_approval AS ? ?`,
+		srv.Store.ActiveGroupAncestors().ManagedByUser(user).
+			Joins(`
+				JOIN groups_groups_active
+					ON groups_groups_active.child_group_id = groups_ancestors_active.child_group_id AND
```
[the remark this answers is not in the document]
zenovich: fixed

app/api/groups/get_members.go
zenovich: I can use LEFT JOIN LATERAL (done)

PR #387, opened 2020-01-09
Generate predictable (where possible) or reproducible IDs for new rows inserted by DB migrations

zenovich:


db/migrations/60_merge_users_items_into_groups_attempts.sql
```diff
@@ -20,6 +21,7 @@ INSERT INTO `groups_attempts` (
     `autonomy`,`started_at`,`validated_at`,`finished_at`,`latest_activity_at`,`thread_started_at`,
     `best_answer_at`,`latest_answer_at`,`latest_hint_at`,`ranked`,`all_lang_prog`,`ancestors_computation_state`)
 SELECT
+    (FLOOR(RAND(1) * 1000000000) + FLOOR(RAND(2) * 1000000000) * 1000000000 + `groups_to_insert`.`id` + `items`.`id`) % 9223372036854775806 + 1,
```
smadbe: You could just hardcode the value you expect for `(FLOOR(RAND(1) * 1000000000) + FLOOR(RAND(2) * 1000000000) * 1000000000` ? Ah no you want different value for each row maybe... you should be able to get no collision with some arithmetic on group.id, item.id and users_items.id.

zenovich:
```
If an integer argument N is specified, it is used as the seed value:

With a constant initializer argument, the seed is initialized once when the statement is prepared, prior to execution.

With a nonconstant initializer argument (such as a column name), the seed is initialized with the value for each invocation of RAND().
```
So for each row it generates different values, but they are predictable because of seeds initialization and ORDER BY. I tried arithmetic on group.id & item.id (sum, product, sha1 of product), but this led to collissions. I also tried ```FLOOR(RAND(1) * 1000000000) + FLOOR(RAND(2) * 1000000000) * 1000000000``` without adding group.id & item.id, but it failed during reapplying.

zenovich:
> Globally I would have relied on arithmetic of the other ids which make the row unique instead of using rand.
Me too, but that didn't work (there were collisions).

> I assume you are 100% sure that rand(n) always give the same numbers on every system.
Yes, I'm 100% sure.

PR #390, opened 2020-01-14
Clean "answers" table and services

### LINKED ISSUE #379: Clean "answers" table and services
Overview: Multiple changes around the answers table and services. 

# Tasks:
* rename `users_answers` to `answers`
* rename `answers.user_id` to `author_id`
* remove `answers.name` (unused), `lang_prog` and `validated` (up to the attempt)
* rename `answers.submitted_at` to `created_at`
* add comment to `answers.type`: "'Submission' for answers submitted for grading, 'Saved' for manual backups of answers, 'Current' for automatic snapshots of the latest answers (unique for a user on an attempt) "
* create a new table `submissions` with attributes:
  - answer_id (PK, FK)
  - score (from answers), not null
  - graded_at (from answers), not null
* migrate data:
  - copy score and graded_at of answers with "Submission" type which have a score NOT NULL to the `submissions` table. For the few (1559) with a score but no `graded_at`, use `submitted_at`
* remove `score` and `graded_at` from answers
* remove the default value for `answer.type` (I do not find sane to allow not specifying the type for creating an answer)  

Misc service fixes: (can be done in separated PRs)
* list answers:
  - "Return answers (i.e., history of submissions and current answer) " -> "Return answers (i.e., saved answers, current answer and submissions) "
  - "If item_id and user_id are given, the authenticated user should have group_id equal to the input user_id or be an owner of a group containing the input user_id." -> "If item_id and user_id are given, the authenticated user should be the input user_id or a manager of a group containing the input user_id."
  - the user tuple should return first/lastname only if the auth user is allowed to see them
* get answer by id:
  - replace the check using `has_attempts` by a check that the attempt.group_id should be either the user or a team the user is member of
* itemGetAnswerToken:
  - change the swagger tag to "answers"
  - in the `SubmitNewAnswer`, put "Submission" as type explicitly
* saveGrade
  - saveNewScoreIntoUserAnswer (to be renamed probably): just create a "submission" row and fail if exists already (no need to check the attempt and user as the answer_id is not supposed to be modifiable, right?)
* itemAnswerUpdateCurrent
  - shouldn't the url be changed from (PUT) `/answers/current` to `/attempts/:id/answers/current` ? (open question)
  - For the has_attempts, same as above
  - This service has to ensure there is always at max 1 row for a given (user, attempt, 'Current') while updating `state`, `answer`, and `created_at` (! was not done). Instead of the current "get-or-create + update", I would do either a "delete-all + create" or "update + create-if-no-row-updated"

zenovich:
Fixes #379


app/database/answer_store_integration_test.go
```diff
@@ -0,0 +1,189 @@
+// +build !unit
```
golangcibot: /go/pkg/mod/github.com/jinzhu/gorm@v1.9.6/dialect_mysql.go:175:10: RuneCountInString not declared by package utf8 (from `typecheck`)

PR #395, opened 2020-01-15
Drop table users_items, remove updateActiveAttempt, update other services

### LINKED ISSUE #392: Remove `users_items`
Goal: The users_items to get the current attempt by (user,item) is just not enough to get the active attempt. It should depends as well on the context (team or user) and soon on parent attempt. Let's take an alternative approach which is to let the UI deal with the active attempts and default on the most recent attempt.

Tasks:
* remove the `users_items` table
* in getItem, remove `user_active_attempt` from the output
* in getNavData, remove `user_active_attempt` and return for each item:
  - "best_score" (max among attempts), 
  - "validated" (max among attempt or "validated" of the one with the best score)
* remove update active attempt
* in create attempt, don't set active attempt anymore

zenovich:
Fixes #392


app/api/items/navigation_data_mapper.go
```diff
@@ -58,20 +53,23 @@ func getRawNavigationData(dataStore *database.DataStore, rootID int64, user *dat
...
 			items.child_order AS child_order,
 			items.content_view_propagation,
 			items.parent_item_id AS parent_item_id,
 			items.item_grandparent_id AS item_grandparent_id,
 			items.can_view_generated_value
 		FROM ? items`, itemThreeGenQ.SubQuery()).
 		JoinsUserAndDefaultItemStrings(user).
-		Joins("LEFT JOIN users_items ON users_items.item_id=items.id AND users_items.user_id=?", user.GroupID).
-		Joins("LEFT JOIN attempts ON attempts.id=users_items.active_attempt_id").
+		Joins(`
```
smadbe: As a max will always return a row, you can do a inner join, so not using this `ON 1` insanity ;-).
`JOIN LATERAL( ... ) AS best_scores`

zenovich: As the query contains 'GROUP BY' it still can return an empty set :(

PR #403, opened 2020-01-16
 Clean items_* and languages tables

### LINKED ISSUE #385: Clean items_* and languages tables
## Goal 
Move towards more consistency and constraints enforced by the DB schema

### Languages
* Add comment to table: "Languages supported for content"
* In `languages` table, make `code`:
   - the first column
   - the PK
   - delete the "code" index
   - renamed to "tag"
   - with length 6
   - with comment "Language tag as defined in RFC5646"
* Create `items.default_language_tag` and `items_strings.language_tag`
* Match items.default_language_tag and `items_strings.language_tag` to the language tag via the `languages` table, put 'fr' for '0's. 
* Remove `languages.id`, `items.default_language_id` and `items_strings.language_id`

### items_strings
* Add FK from `items_strings.item_id` to `items.id`
* Add FK from `items_strings.language_tag` to `languages`
* Make (`item_id`, `language_tag`) the new PK of `items_strings`
* Remove ` items_strings.id`
* Remove  `items_strings.ranking_comment`

### items
* Remove `items` matching no `items_strings.item_id` (27 rows)
* Set `items.default_language_tag` which are NULL to one of the items_strings they have, if multiple choices, prefer 'fr', then 'en'.
* Make `items.default_language_tag` not null
* Add FK from `items.(id, default_language_tag)` to `items_strings.PK`.
* Remove `display_children_as_tabs`
* Add new attributes `is_root` bool, NOT NULL, default 0, comment "Whether this item is intended to be a root chapter (in order to detect real orphans more easily)"
* Set `is_root` to 1 when type = 'Root'
* Change all type IN ('Root', 'Category') to 'Chapter'
* Remove 'Root' and 'Category' types

### items_items
* Find a better naming? (`items_relationships`, `items_children`, `items_parents`, ... ?)
* Clean entries with the same `(child,parent)` (keep any of the duplicated row)
* Make `(child_item_id, parent_item_id)` the new PK
* Remove `id`
* make FK from `child_item_id` and `parent_item_id` to `items.id`
* Make a cleaning of the table indexes 

### items_ancestors
* Make `(child_item_id, ancestor_item_id)` the new PK
* Remove `id`
* make FK from `child_item_id` and `ancestor_item_id` to `items.id`
* Make a cleaning of the table indexes

zenovich:
Fixes #385


app/database/item_store_test.go
```diff
@@ -290,3 +291,44 @@ func TestItemStore_GetAccessDetailsForIDs_HandlesDBErrors(t *testing.T) {
 
 	assert.NoError(t, dbMock.ExpectationsWereMet())
 }
+
```
smadbe: As briefly discussed previously, I am not sure these unit tests (especially on stores) are very valuable when I see their cost of writing (quite lengthy) and maintenance. For instance, the fact they do the expected request depend on the "hasRows" implementation which is not the point of the implementation here, and anyway should be spotted by the BDD tests of the service using it.  
I would be in favor (for the future, the current ones are ok! 🙂) of just checking that running it on a mock db just not crash (that the query gen was able to generate something) and that a fail on the query is just returned, while ensuring 100% test coverage. This could be very generic to every such function and such tests could be reduced to 4-5 lines in total. Of course, in term of testing, that's not be perfect, but as all these functions are tested by services, I thing that is very acceptable. 
I am mainly worried that somebody else taking over this code will not maintain them otherwise (or will just delete them to prevent losing 30min in tests for a 2min fix of code), these tests convincing himself that testing is just a loss of time.

db/migrations/84_drop_column_items_items_id.sql
```diff
@@ -0,0 +1,48 @@
...
+        FROM `items_items`
+            JOIN (
+                SELECT `parent_item_id`, `child_item_id`, count(*) AS cnt
+                FROM `items_items`
+                GROUP BY `parent_item_id`, `child_item_id` HAVING cnt > 1
+            ) AS `duplicates` -- not unique (parent_item_id, child_item_id) pairs
+                ON `duplicates`.`parent_item_id` = `items_items`.`parent_item_id` AND
+                 `duplicates`.`child_item_id` = `items_items`.`child_item_id`
+    ) AS `duplicated_rows` -- ids of duplicated row with row numbers within (parent_item_id, child_item_id) group, ordered by child_order
+        ON `duplicated_rows`.`id` = `items_items`.`id` AND `duplicated_rows`.`number` > 1;  -- remove duplicates keeping first rows
```
smadbe: Seems to be a very complex version of the simple: 
```sql
DELETE items_items FROM items_items 
JOIN items_items orig USING (parent_item_id, child_item_id) 
WHERE items_items.id > orig.id;
```
😁

app/api/items/items.go
```diff
@@ -134,17 +134,17 @@ func insertItemItems(store *database.DataStore, spec []*insertItemItemsSpec) {
 	var values = make([]interface{}, 0, len(spec)*9)
 
 	for index := range spec {
-		values = append(values, store.NewID(),
+		values = append(values,
```
smadbe: The fact that this function is still called in the `RetryOnDuplicatePrimaryKeyError` in `add_item` (and probably others) cannot be a problem? In case there is a duplicate child, could it fail the same way in "loop" ?
It would be nice to prevent this issue not by checking manually there is no duplicate or by expecting the context to avoid such a case, but by failing properly on the DuplicatePrimaryKeyError on the `items_items` table (while keeping it on the `items` table)

(+ I assume the related services do test this scenario anyway)

zenovich: Fixed (although it could not cause problems)

db/migrations/84_drop_column_items_items_id.sql
zenovich: Mine orders by `child_order`

db/migrations/83_rework_languages.sql
```diff
@@ -0,0 +1,157 @@
...
+    ORDER BY `items_strings`.`language_tag` = 'fr' DESC, `items_strings`.`language_tag` = 'en' DESC
+    LIMIT 1
+) WHERE `items`.`default_language_tag` IS NULL;
+
+UPDATE `items` SET `is_root` = 1, `type` = 'Chapter' WHERE `type` IN ('Root', 'Category');
+
+DROP TRIGGER `before_insert_languages`;
+ALTER TABLE `languages`
+    DROP PRIMARY KEY,
+    ADD PRIMARY KEY (`tag`),
```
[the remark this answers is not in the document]
zenovich: fixed

app/api/items/update_item.go
```diff
@@ -180,21 +180,21 @@ func updateChildrenAndRunListeners(formData *formdata.FormData, store *database.
...
-	formData.RegisterTranslation("default_language_id",
+func registerDefaultLanguageTagValidator(formData *formdata.FormData, store *database.DataStore, itemID int64) {
+	formData.RegisterValidation("default_language_tag", constructDefaultLanguageTagValidator(formData, store, itemID))
+	formData.RegisterTranslation("default_language_tag",
 		"default language should exist and there should be item's strings in this language")
 }
 
-// constructDefaultLanguageIDValidator constructs a validator for the DefaultLanguageID field.
+// constructDefaultLanguageTagValidator constructs a validator for the DefaultLanguageTag field.
 // The validator checks that the language exists and there is an items_strings row in this language for the item.
```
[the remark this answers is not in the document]
zenovich: This way the error message would be 'internal server error'. Would it be okay?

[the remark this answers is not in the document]
smadbe: Internal server error has to be only for unexpected error. Here it is an expected error, but you can catch constraint error as you did in another recent PR, can't you?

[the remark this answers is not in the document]
zenovich: sure

app/api/items/items.go
smadbe: Why couldn't it cause problems?

db/migrations/84_drop_column_items_items_id.sql
smadbe:
```sql
DELETE items_items FROM items_items 
JOIN items_items orig USING (parent_item_id, child_item_id) 
WHERE items_items.order > orig.order;
```
Still much much less complex than yours :-P

smadbe: (you can leave it like that if you want, but I just saw it twice with something very complex)

app/api/items/items.go
zenovich: Oh, sorry. Yes, it really could cause problems because we still have the primary key in items_items (a composite one).

db/migrations/83_rework_languages.sql
```diff
@@ -0,0 +1,157 @@
...
+    DROP COLUMN `id`;
+
+ALTER TABLE `items`
+    MODIFY COLUMN `default_language_tag` VARCHAR(6) NOT NULL DEFAULT 'fr'
```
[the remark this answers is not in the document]
zenovich: fixed

db/migrations/83_rework_languages.sql
```diff
@@ -0,0 +1,157 @@
...
+        COMMENT 'Default language tag of this task (the reference, used when comparing translations)',
+    DROP COLUMN `default_language_id`,
+    ADD CONSTRAINT `fk_items_default_language_tag_languages_tag`
+        FOREIGN KEY (`default_language_tag`) REFERENCES `languages`(`tag`),
```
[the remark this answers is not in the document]
zenovich: fixed

db/migrations/84_drop_column_items_items_id.sql
zenovich: fixed

app/api/items/update_item.go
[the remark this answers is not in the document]
zenovich: fixed

app/api/items/items.go
smadbe: Wouldn't there be a way to make sure in `RetryOnDuplicatePrimaryKeyError` cannot go wrong in general. In addition to the sql error code, do you get a message? If so, you could match the PK id name (that you would give when calling the fct) maybe?

app/api/items/add_item.go
```diff
@@ -267,13 +267,15 @@ func (srv *Service) insertItem(store *database.DataStore, user *database.User, f
...
 
+	service.MustNotBeError(store.Exec("SET FOREIGN_KEY_CHECKS = 0").Error())
 	service.MustNotBeError(store.RetryOnDuplicatePrimaryKeyError(func(s *database.DataStore) error {
 		itemID = s.NewID()
 
 		itemMap["id"] = itemID
 		itemMap["default_language_tag"] = newItemRequest.LanguageTag
 		return s.Items().InsertMap(itemMap)
 	}))
+	service.MustNotBeError(store.Exec("SET FOREIGN_KEY_CHECKS = 0").Error())
```
smadbe: Didn't you want to set to "1" after the block?
BTW, if the goal if to set to 0 before without forgetting to set it back to 1, wouldn't it better to have a fct taking a block as arg? In Golang, this kind of construct are maybe not encouraged.

zenovich: fixed

app/api/items/items.go
zenovich: PK's name is always 'PRIMARY' in MySQL.

zenovich: > Wouldn't there be a way to make sure in RetryOnDuplicatePrimaryKeyError cannot go wrong in general.

zenovich:
> `get_breadcrumbs` mentions 'Root' type in its specs : to be changed

fixed

PR #417, opened 2020-01-23
Create missing attempts in ComputeAllAttempts()

### LINKED ISSUE #400: Fix result (attempt) propagation for non-existing attempts
### Goal
The results obtained on a task by a user should be propagated to every parents of this task, even it has not been explored yet (so if there are no attempt yet)

### Remark
In the next step, this will be improved by not creating the attempts for non-visible items. As it will be quite complex for the overall handling and just an "add-on" to the above propagation, it is good to complete this non-trivial task first.

### Subtasks
In attempt result propagation:
  - convert the big update to an INSERT-ON-DUPLICATE-KEY-UPDATE which creates/updates rows for every parent of attempt marked as `to_be_recomputed` which do not have a descendant marked as "to_be_recomputed". 
  - It means we can get rid of the `markAsProcessing` query and probably of the need for the "processing" state.
  - it means we should be able the write the loop as a single query with a CTE (INSERT... WITH.. SELECT..) (?)

zenovich:
Also
* modify triggers of `items_items` in order to delete attempts that possibly are outdated and to mark child attempts as changed so that actual parent attempts could be recreated
* update tests

Fixes #400

### DIFF
--- a/app/database/attempt_store_compute_all_attempts.go
+++ b/app/database/attempt_store_compute_all_attempts.go
@@ -26,7 +26,7 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
 
 	// Use a lock so that we don't execute the listener multiple times in parallel
 	mustNotBeError(s.WithNamedLock(computeAllAttemptsLockName, computeAllAttemptsLockTimeout, func(ds *DataStore) error {
-		// We mark as 'todo' all ancestors of objects marked as 'todo'
+		// We mark as 'to_be_recomputed' all ancestors of objects marked as 'to_be_recomputed'/'changed'
 		// (this query can take more than 50 seconds to run when executed for the first time after the db migration)
 		mustNotBeError(ds.db.Exec(
 			`UPDATE attempts AS ancestors
@@ -36,14 +36,39 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
 				descendants.group_id = ancestors.group_id
 			)
 			SET ancestors.result_propagation_state = 'to_be_recomputed'
-			WHERE descendants.result_propagation_state IN ('to_be_recomputed', 'changed')`).Error)
+			WHERE ancestors.result_propagation_state != 'to_be_recomputed' AND
+				(descendants.result_propagation_state = 'to_be_recomputed' OR
+				 descendants.result_propagation_state = 'changed')`).Error)
+
+		// Insert missing attempts for chapters & courses having descendants marked as 'to_be_recomputed'/'changed'
+		// (this query can take more than 25 seconds when executed for the first time after the db migration)
+		mustNotBeError(ds.RetryOnDuplicatePrimaryKeyError(func(retryStore *DataStore) error {
+			return retryStore.Exec(`
+				INSERT INTO attempts (id, group_id, item_id, ` + "`order`, " + `result_propagation_state)
+				SELECT
+					FLOOR(RAND() * 1000000000) + FLOOR(RAND() * 1000000000) * 1000000000,
+					descendants.group_id, items_ancestors.ancestor_item_id, 1, 'to_be_recomputed'
+				FROM attempts AS descendants
+				JOIN items_ancestors ON items_ancestors.child_item_id = descendants.item_id
+				LEFT JOIN attempts AS existing ON (
+					existing.group_id = descendants.group_id AND
+					existing.item_id = items_ancestors.ancestor_item_id
+				)
+				WHERE (
+					descendants.result_propagation_state = 'to_be_recomputed' OR
+					descendants.result_propagation_state = 'changed'
+				) AND existing.id IS NULL
+				GROUP BY descendants.group_id, items_ancestors.ancestor_item_id
+			`).Error()
+		}))
 
 		hasChanges := true
 
 		var markAsProcessingStatement, updateStatement *sql.Stmt
 
 		for hasChanges {
-			// We mark as "processing" all objects that were marked as 'todo' and that have no children not marked as 'done'
+			// We mark as "processing" all objects that were marked as 'to_be_recomputed' and
+			// that have no children marked as 'to_be_recomputed'.
 			// This way we prevent infinite looping as we never process items that are ancestors of themselves
 			if markAsProcessingStatement == nil {
 				const markAsProcessingQuery = `
--- a/app/database/attempt_store_compute_all_attempts.go
+++ b/app/database/attempt_store_compute_all_attempts.go
@@ -40,7 +40,7 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
 				(descendants.result_propagation_state = 'to_be_recomputed' OR
 				 descendants.result_propagation_state = 'changed')`).Error)
 
-		// Insert missing attempts for chapters & courses having descendants marked as 'to_be_recomputed'/'changed'
+		// Insert missing attempts for chapters having descendants marked as 'to_be_recomputed'/'changed'
 		// (this query can take more than 25 seconds when executed for the first time after the db migration)
 		mustNotBeError(ds.RetryOnDuplicatePrimaryKeyError(func(retryStore *DataStore) error {
 			return retryStore.Exec(`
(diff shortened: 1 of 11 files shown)


app/database/attempt_store_compute_all_attempts.go
```diff
@@ -36,14 +36,39 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
 				descendants.group_id = ancestors.group_id
 			)
 			SET ancestors.result_propagation_state = 'to_be_recomputed'
-			WHERE descendants.result_propagation_state IN ('to_be_recomputed', 'changed')`).Error)
+			WHERE ancestors.result_propagation_state != 'to_be_recomputed' AND
+				(descendants.result_propagation_state = 'to_be_recomputed' OR
+				 descendants.result_propagation_state = 'changed')`).Error)
```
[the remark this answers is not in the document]
zenovich: On my computer 'IN ('to_be_recomputed', 'changed')' didn't use an index at all.

app/database/attempt_store_compute_all_attempts.go
```diff
@@ -36,14 +36,39 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
...
+
+		// Insert missing attempts for chapters & courses having descendants marked as 'to_be_recomputed'/'changed'
```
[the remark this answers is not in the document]
zenovich: fixed

db/migrations/81_rename_groups_attempts_to_attempts.sql
```diff
@@ -70,13 +70,14 @@ CREATE TRIGGER `before_update_items_items` BEFORE UPDATE ON `items_items` FOR EA
...
 
-        IF (OLD.`parent_item_id` != NEW.`parent_item_id`) THEN
-            UPDATE `attempts` SET `result_propagation_state` = 'to_be_recomputed'
-            WHERE `item_id` = NEW.`parent_item_id`;
-        END IF;
+        DELETE FROM `attempts` WHERE (`item_id` = OLD.parent_item_id OR `item_id` = NEW.parent_item_id OR
+                                      `item_id` IN (
+                                          SELECT `ancestor_item_id` FROM `items_ancestors`
+                                          WHERE `child_item_id` = OLD.parent_item_id OR `child_item_id` = NEW.parent_item_id
+                                    )) AND `order` = 1 AND `creator_id` IS NULL;
```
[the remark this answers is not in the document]
zenovich: fixed

db/migrations/81_rename_groups_attempts_to_attempts.sql
```diff
@@ -121,7 +122,13 @@ CREATE TRIGGER `before_delete_items_items` BEFORE DELETE ON `items_items` FOR EA
...
     WHERE `permissions_generated`.`item_id` = OLD.`parent_item_id`;
 
-    UPDATE `attempts` SET `result_propagation_state` = 'to_be_recomputed'
+    DELETE FROM `attempts` WHERE (`item_id` = OLD.parent_item_id OR
+                                  `item_id` IN (
+                                      SELECT `ancestor_item_id` FROM `items_ancestors`
+                                      WHERE `child_item_id` = OLD.parent_item_id
+                                  )) AND `order` = 1 AND `creator_id` IS NULL;
+
+    UPDATE `attempts` SET `result_propagation_state` = 'changed'
```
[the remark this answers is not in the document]
zenovich: fixed

app/database/attempt_store_compute_all_attempts.go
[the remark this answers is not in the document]
smadbe: Waw, weird that it is not able to do such a basic optimization.

db/migrations/81_rename_groups_attempts_to_attempts.sql
```diff
@@ -73,10 +73,8 @@ CREATE TRIGGER `before_update_items_items` BEFORE UPDATE ON `items_items` FOR EA
         UPDATE `attempts` SET `result_propagation_state` = 'to_be_recomputed'
         WHERE `item_id` = OLD.`parent_item_id`;
 
-        IF (OLD.`parent_item_id` != NEW.`parent_item_id`) THEN
-            UPDATE `attempts` SET `result_propagation_state` = 'to_be_recomputed'
-            WHERE `item_id` = NEW.`parent_item_id`;
-        END IF;
+        UPDATE `attempts` SET `result_propagation_state` = 'changed'
+        WHERE `item_id` = OLD.`child_item_id` OR `item_id` = NEW.`child_item_id`;
```
smadbe: If you do so, the previous request on the parent may only be done if `(OLD.`parent_item_id` != NEW.`parent_item_id`)`.

smadbe: Why do you need to propagate the old child? There is nothing to be changed in its other parents.

zenovich:
> I didn't have the time finish the discussion yesterday, but so, maybe deleting in some cases does really make sense, but only for propagated ones which are not holding other information, so we could say we delete the ancestors of the delete relations with `started_at IS NULL AND score_edit_rule IS NULL`
> 
> (Other debate:)
> BTW, the change of the parent or child of an items_items is something which should not happen in practice, if we need to change a child, we should delete and recreate but not modify these attributes.
> I would be in favor of dropping the "before_update_items_items" for simplify thing, but the problem would be manual changes. For prevent change of parent or child, we could force fail if changed (https://stackoverflow.com/a/22489342/759432).

fixed

db/migrations/81_rename_groups_attempts_to_attempts.sql
```diff
@@ -33,48 +33,7 @@ END
...
-
-        INSERT IGNORE INTO `items_propagate` (id, ancestors_computation_state)
-        VALUES (NEW.child_item_id, 'todo') ON DUPLICATE KEY UPDATE `ancestors_computation_state` = 'todo';
-
-        UPDATE `attempts` SET `result_propagation_state` = 'to_be_recomputed'
-        WHERE `item_id` = OLD.`parent_item_id`;
-
-        UPDATE `attempts` SET `result_propagation_state` = 'changed'
-        WHERE `item_id` = OLD.`child_item_id` OR `item_id` = NEW.`child_item_id`;
+        SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = 'Unable to change immutable items_items.parent_item_id and/or items_items.child_item_id';
```
smadbe: 👍 
We have to add that if the `score_weight` or `category` has changed, we have to recompute the score of the parent anyway. Or we leave this duty to the service?

db/migrations/81_rename_groups_attempts_to_attempts.sql
```diff
@@ -119,6 +78,8 @@ CREATE TRIGGER `before_delete_items_items` BEFORE DELETE ON `items_items` FOR EA
     FROM `permissions_generated`
     WHERE `permissions_generated`.`item_id` = OLD.`parent_item_id`;
 
+    DELETE FROM `attempts` WHERE `item_id` = OLD.`parent_item_id` AND `started_at` IS NULL AND `score_edit_rule` IS NULL;
+
     UPDATE `attempts` SET `result_propagation_state` = 'to_be_recomputed'
     WHERE `item_id` = OLD.`parent_item_id`;
```
smadbe: I think that is still not correct, if the parent has other children, and has null `started_at` and `score_edit_rule`, it will be deleted and so its score will not be recomputed. 

As you had done before, ancestors may have to be deleted as well somehow... but it is maybe complicated to compute which ones can be removed for sure, it would be the ones with no started_at, no score_edit_rule, and without children after applying the same criteria on children... maybe that's safer not to delete anything actually. 😕

zenovich: > I think that is still not correct, if the parent has other children, and has null started_at and score_edit_rule, it will be deleted and so its score will not be recomputed.

db/migrations/81_rename_groups_attempts_to_attempts.sql
zenovich: We don't have services modifying items_items.

PR #428, opened 2020-01-30
Add `as_team_id` for some items/attempts services

### LINKED ISSUE #409: Add `as_team_id` for some items/attempts services
## Goal
When he is in "team mode", the user should only see what his team can see. Some services have to be adapted to do actions as the team in this case.

## Tasks
* Add `as_team_id` optional parameter query param in the following services. If given, 
  - `itemView`: verify visibility as the team and compute the `has_attempts` for the team  
  - `getNavData`: use the team visibility for items, and attempts for the team only (+ "max among all the user's attempts" -> "max among all the participant's attempts)
* In `contestEnter`, replace the url by `POST /contests/:item_id/enter` and use the `as_team_id` query param to distinguish if it is done as the user or a team. 
* In `contestGetQualificationState`, the same (using `GET /contests/{item_id}/qualification-state`)

zenovich:
Fixes #409

### DIFF
--- a/app/api/contests/get_qualification_state.go
+++ b/app/api/contests/get_qualification_state.go
@@ -138,7 +138,7 @@ func (srv *Service) getContestInfoAndQualificationStateFromRequest(r *http.Reque
 		ContestMaxTeamSize       int32
 		ContestEnteringCondition string
 	}
-	err = store.Items().VisibleByID(user, itemID).Where("items.duration IS NOT NULL").
+	err = store.Items().VisibleByID(user.GroupID, itemID).Where("items.duration IS NOT NULL").
 		Select("items.allows_multiple_attempts, items.contest_max_team_size, items.contest_entering_condition").
 		Take(&contestInfo).Error()
 	if gorm.IsRecordNotFoundError(err) {
--- a/app/api/contests/get_qualification_state.go
+++ b/app/api/contests/get_qualification_state.go
@@ -50,11 +50,11 @@ type contestGetQualificationStateResponse struct {
 // ---
 // summary: Get qualification state
 // description: >
-//                For the given contest and the given participant, returns the qualification state,
-//                i.e. whether he can enter the contest, and info on each team member.
+//                For the given contest and the given participant (the current user or his team if `as_team_id` is set),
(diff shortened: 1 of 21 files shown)


app/database/item_store.go
```diff
@@ -14,29 +14,29 @@ type ItemStore struct {
 	*DataStore
 }
 
-// Visible returns a view of the visible items for the given user
-func (s *ItemStore) Visible(user *User) *DB {
-	return s.WhereItemsAreVisible(user)
+// Visible returns a view of the visible items for the given group
```
smadbe: Converting all these functions from "for a given user" to " for a given participant" would probably have been enough.

app/api/contests/get_qualification_state.go
```diff
@@ -128,25 +125,28 @@ func (srv *Service) getContestInfoAndQualificationStateFromRequest(r *http.Reque
 		return nil, service.ErrInvalidRequest(err)
 	}
 
-	groupID, err := service.ResolveURLQueryPathInt64Field(r, "group_id")
-	if err != nil {
-		return nil, service.ErrInvalidRequest(err)
+	groupID := user.GroupID
```
smadbe: Ah you cannot use your new function here because it is another package. 😕
Cannot it be moved to the common service package? This is not something really related to items.

app/api/items/items.go
```diff
@@ -172,3 +174,24 @@ func createContestParticipantsGroup(store *database.DataStore, itemID int64) int
 	}))
 	return participantsGroupID
 }
+
+func (srv *Service) chooseBetweenUserAndAsTeamID(httpReq *http.Request, user *database.User) (int64, service.APIError) {
```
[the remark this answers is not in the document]
zenovich: fixed

app/database/item_store.go
zenovich: fixed

app/api/contests/get_qualification_state.go
zenovich: We perform different checks here in the 'contest' package, so we don't need that new function from the 'items' package here.

PR #455, opened 2020-02-07
Introduce "official sessions" and a new permission called "can_make_session_official"

### LINKED ISSUE #445: New permission "can_make_session_official"
## Goal
Sessions (sets of participants who participate to a same activity) are associated with an activity. In order to allow participants to freely register, sessions may be listed on the item itself. But not everybody can display his session on an item.

## Subtasks
* New permission in permissions_granted table (between `can_edit` and `is_owner`): `can_make_session_official` (bool) "Whether the group is allowed to associate official sessions to this item". This permission does not aggregate nor propagate.
* New attribute `groups.is_official_session`, bool default false, "Whether this session is shown on the activity page (require specific permissions)"
* On group update service:
  - allow setting `is_official_session` to true only if the user (or his group ancestors) has `can_make_session_official` on the activity item. 
  - on change of `activity_id`, if `is_official_session` is true, check that the permission is true for the authenticated user on the new item.

zenovich:
Fixes #445

### DIFF
--- a/app/api/groups/update_group.go
+++ b/app/api/groups/update_group.go
@@ -30,8 +30,11 @@ type groupUpdateInput struct {
 	// Nullable
 	CodeExpiresAt *time.Time `json:"code_expires_at"`
 	// Nullable
-	ActivityID  *int64 `json:"activity_id"`
-	OpenContest bool   `json:"open_contest"`
+	ActivityID *int64 `json:"activity_id"`
+	// Can be set only if activity_id is set and
+	// the current user has the 'can_make_session_official' permission on the activity item
+	IsOfficialSession bool `json:"is_official_session"`
+	OpenContest       bool `json:"open_contest"`
 
 	RequireMembersToJoinParent bool `json:"require_members_to_join_parent"`
 	// Nullable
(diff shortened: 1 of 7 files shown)


app/api/groups/update_group.go
```diff
@@ -141,6 +149,51 @@ func (srv *Service) updateGroup(w http.ResponseWriter, r *http.Request) service.
...
+			Where("permissions_granted.can_make_session_official").
+			Where("permissions_granted.item_id = ?", activityIDToCheck).
+			HasRows()
+		service.MustNotBeError(err)
+		if !found {
+			return service.ErrForbidden(errors.New("not enough permissions for attaching the group to the activity as an official session"))
+		}
+	}
+	return service.NoError
+}
```
smadbe: There is something that may be annoying, even if probably already like that in many other places:
If the activity_id is given but the same as before, possibly it is just the former data which is resubmitted as is (just changing the name for instance)... but so in this case the user is not setting a new activity, and so it should not be rechecked as he might not have the perm while his colleague who set it earlier had the perm. 
So IMO, `activityIDSet`/isOfficialSessionSet are not what should be used but rather, `activityIDChanged`/`isOfficialSessionChanged`

app/api/groups/update_group_test.go
```diff
@@ -47,7 +47,8 @@ func Test_validateUpdateGroupInput(t *testing.T) {
 func TestService_updateGroup_ErrorOnReadInTransaction(t *testing.T) {
 	assertUpdateGroupFailsOnDBErrorInTransaction(t, func(mock sqlmock.Sqlmock) {
 		mock.ExpectBegin()
-		mock.ExpectQuery(regexp.QuoteMeta("SELECT groups.is_public FROM `groups` "+
+		mock.ExpectQuery(regexp.QuoteMeta("SELECT groups.is_public, groups.activity_id, groups.is_official_session "+
```
smadbe: It is annoying that these unit tests have to change when you do such changes...

app/api/groups/update_group.go
zenovich: fixed

PR #467, opened 2020-02-12
Introduce new tables: user_batch_prefixes & user_batches

### LINKED ISSUE #458: Batch user creation: Preliminary changes
## Motivations 
Group managers may be allowed to create a set of users in batch (for a contest, ...). See [designs](https://france-ioi.github.io/algorea-designs/01.Groups_02.Composition_01.Dialogs_02.Create-batch-accounts) for the typical UI.

The created logins were currently prefixed by a prefix related to the user, it will be moved the group which "orders" this creation. `groups_login_prefixes` which stores the used prefixes can be changed. 

## Subtasks

* New table `user_batch_prefixes`: "Authorized login prefixes for user batch creation. A prefix cannot be deleted without deleting batches using it." with: 
   - **[PK]** `group_prefix` varchar(13) "Prefix used in front of all batches"
   - `group_id` **[FK on delete cascade]** - "Group and its subgroups in which managers can create users in batch"
   - `max_users` UNSIGNED MEDIUMINT: "Maximum number of users that can be created under this prefix"
   - `allow_new`, bool, default true, "Whether this prefix can be used for new creations"  

* New table `user_batches` "Batches of users which have been created." with:
   - **[PK]** `group_prefix` **[FK on DELETE RESTRICT]**  varchar(13) "Authorized (first) part of the full login prefix`
   - **[PK]** `custom_prefix` varchar(14) "Custom (second) part of the full login prefix." + CHECK that is matches `/^[a-z0-9-]+` (to make sure we have uniqueness of the full prefix) 
   - `size` UNSIGNED MEDIUMINT "Number of users created in this batch`
   - `creator_id`, nullable, FK to `users.id` ON CASCADE SET NULL
   - `created_at`, time, default now

* From `groups_login_prefixes`, extract from `prefix`: before/after the pre-last underscore for group_prefix/custom_prefix (so for  "nice_school_classA_", the group prefix is "nice_school" and the custom_prefix is "classA")  set `count` to the number of logins prefixed by the full prefix. 

* Deduce from that, the `group_prefix` table, so with one row per `group_prefix`, `max_users` to 1000 and group_id any of the group_id matching the group_prefix.

* Remove `groups_login_prefixes`

* Remove `users.login_module_prefix`

zenovich:
Fixes #458


db/migrations/99_create_tables_for_batches_of_users.sql
```diff
@@ -0,0 +1,79 @@
+-- +migrate Up
+UPDATE `groups_login_prefixes`
+    SET `prefix` = CONCAT(`prefix`, '_')
+WHERE `prefix` NOT LIKE '%\_';
+
+CREATE TABLE `user_batch_prefixes` (
+    `group_prefix` VARCHAR(13) NOT NULL PRIMARY KEY COMMENT 'Prefix used in front of all batches',
+    `group_id` BIGINT(20) NOT NULL COMMENT 'Group and its subgroups in which managers can create users in batch',
```
smadbe: I've changed my mind, instead of blocking the group deletion (which may be complicated to understand from the UI) if it has batches (as `user_batches`  is restricted so that they do not disappear without having the possibility to remove the users), let's allow the `group_id` to be NULL and make the FK "ON DELETE SET NULL". 
You can also change the comment to 'Group (and its descendants) in which managers can create users in batch. NULL if the group was deleted, in which case batches should be cleaned manually."

db/migrations/99_create_tables_for_batches_of_users.sql
```diff
@@ -0,0 +1,79 @@
...
+    `allow_new` TINYINT(1) NOT NULL DEFAULT 1 COMMENT 'Whether this prefix can be used for new creations',
+    CONSTRAINT `fk_user_batch_prefixes_group_id_groups_id`
+        FOREIGN KEY (`group_id`) REFERENCES `groups`(`id`) ON DELETE CASCADE
+) ENGINE=InnoDB CHARSET=utf8
+    COMMENT='Authorized login prefixes for user batch creation. A prefix cannot be deleted without deleting batches using it.';
+
+CREATE TABLE `user_batches` (
+    `group_prefix` VARCHAR(13) NOT NULL COMMENT 'Authorized (first) part of the full login prefix',
+    `custom_prefix` VARCHAR(14) NOT NULL
+        COMMENT 'Custom (second) part of the full login prefix',
```
smadbe: -> 'Second part of the full login prefix, given by the user creating the batch'

db/migrations/99_create_tables_for_batches_of_users.sql
```diff
@@ -0,0 +1,79 @@
+-- +migrate Up
+UPDATE `groups_login_prefixes`
+    SET `prefix` = CONCAT(`prefix`, '_')
+WHERE `prefix` NOT LIKE '%\_';
+
+CREATE TABLE `user_batch_prefixes` (
+    `group_prefix` VARCHAR(13) NOT NULL PRIMARY KEY COMMENT 'Prefix used in front of all batches',
+    `group_id` BIGINT(20) NOT NULL COMMENT 'Group and its subgroups in which managers can create users in batch',
+    `max_users` MEDIUMINT UNSIGNED DEFAULT NULL COMMENT 'Maximum number of users that can be created under this prefix',
+    `allow_new` TINYINT(1) NOT NULL DEFAULT 1 COMMENT 'Whether this prefix can be used for new creations',
```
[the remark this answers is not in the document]
zenovich: fixed

db/migrations/99_create_tables_for_batches_of_users.sql
zenovich: fixed

db/migrations/99_create_tables_for_batches_of_users.sql
zenovich: fixed

app/api/groups/remove_child.go
```diff
@@ -24,7 +24,7 @@ import (
 //
 //   If a group gets deleted, the service also deletes `groups_groups`, `attempts`,
 //   `group_membership_changes`, `group_pending_requests`,
-//   `permissions_granted`, `permissions_generated`, `groups_login_prefixes`, and `filters` linked to the group.
+//   `permissions_granted`, `permissions_generated`, and `filters` linked to the group.
```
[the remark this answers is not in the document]
zenovich: we do not delete them anymore (because of ON DELETE SET NULL in user_batch_prefixes)

PR #477, opened 2020-02-26
Convert can_view_* into permissions, introduce  items.entering_time_*

### LINKED ISSUE #447: can_grant_view:enter & can_enter_* in permissions
## Motivation
The *can_enter_\** permissions will not only be used for contests but as an alternative way (to can_view:content) to allow users to access content and create attempts. So `can_enter` can be moved to the other permission set.

## Subtasks
* Move *can_enter_\** permissions from `groups_contest_items` to `permissions_granted* (after `can_view` with description "Time from/until which the group can “enter” this item". These permissions do not aggregate nor propagate.
* Add "enter" in the can_grant_view permissions (cfr https://france-ioi.github.io/algorea-devdoc/design/access-rights/items/ for reference)
* Add to the `items` table the attributes `entering_time_min` and `entering_time_max`, not nullable time, mysql.min/max by default, "Lower/Upper bound on the entering time. Has the priority over given can_enter_from/until permissions."
* Update `get_qualification_state` service accordingly: so check that 'now' is inside the can_enter_from-until range of any granted permissions to the participant and its ancestors, and inside the entering_time_min-entering_time_max range.

zenovich:
* move groups_contestitems.can_enter_* to permissions_granted
* add the new value "enter" into permissions_granted/permissions.generated.can_grant_view
* introduce items.entering_time_* columns and support them in contestEnter & contestGetQualificationState

Fixes #447


db/migrations/102_rework_enter_permissions.sql
```diff
@@ -0,0 +1,56 @@
...
+    DROP COLUMN can_enter_from,
+    DROP COLUMN can_enter_until;
+
+ALTER TABLE `items`
+    ADD COLUMN `entering_time_min` DATETIME DEFAULT NULL
+        COMMENT 'Lower bound on the entering time. Has the priority over given can_enter_from/until permissions.'
+        AFTER `contest_entering_condition`,
+    ADD COLUMN `entering_time_max` DATETIME DEFAULT NULL
+        COMMENT 'Upper bound on the entering time. Has the priority over given can_enter_from/until permissions.'
+        AFTER `entering_time_min`;
```
smadbe: Contrary to what I said, let's consider these two attributes NOT NULL. It will be more consistent with `contest_entering_condition`, `contest_max_team_size`, ... and just use the default values when not applicable. Handling null values seems more complicated and less clear.

db/migrations/102_rework_enter_permissions.sql
```diff
@@ -0,0 +1,56 @@
+-- +migrate Up
+ALTER TABLE `permissions_granted`
+    ADD COLUMN `can_enter_from` DATETIME NOT NULL DEFAULT '9999-12-31 23:59:59'
+        COMMENT 'Time from which the group can “enter” this item' AFTER `can_view`,
+    ADD COLUMN `can_enter_until` DATETIME NOT NULL DEFAULT '9999-12-31 23:59:59'
+        COMMENT 'Time until which the group can “enter” this item' AFTER `can_enter_from`,
```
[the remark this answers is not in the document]
zenovich: fixed

db/migrations/102_rework_enter_permissions.sql
zenovich: fixed

PR #480, opened 2020-02-27
Implement the userBatchPrefixesView service

### LINKED ISSUE #459: Get all user-batch prefixes service
## Overview 
Group managers may be allowed to create a set of users in batch (for a contest, ...). See [designs](https://france-ioi.github.io/algorea-designs/01.Groups_02.Composition_01.Dialogs_02.Create-batch-accounts) for the typical UI.

The user should be able to choose among all the group's allowed prefixes, which one to use (not shown on the UI).

On `GET /groups/{group_id}/user_batch_prefixes/`

## Input
* group_id (in url)

## Output
* A list of the prefixes matching one the input group's ancestors and which are "allowed" and which match a group the the user can manage :
  - group_prefix
  - group_id
  - max_users  
  - total_size : sum of users in the batches already using this group_prefix

## Validations
* The authenticated user is manager of the input.group_id (or its ancestors) with at least can_manage:memberships.

zenovich:
Fixes #459


app/api/groups/get_user_batch_prefixes.feature
```diff
@@ -0,0 +1,76 @@
...
+      | test2        | 13       | 0         | 80        |
+      | test3        | 21       | 1         | 70        |
+      | test4        | 14       | 1         | 60        |
+    And the database has the following table 'user_batches':
+      | group_prefix | custom_prefix | size | creator_id |
+      | test         | custom        | 100  | null       |
+      | test         | custom1       | 200  | 13         |
+      | test1        | cust          | 300  | 21         |
+      | test1        | cust1         | 400  | null       |
+      | test2        | cus           | 300  | 21         |
```
smadbe:
```suggestion
      | test1        | cust1         | 400  | null       |
```
Maybe having some prefixes without batches would be better for the test.

app/api/groups/get_user_batch_prefixes.feature
```diff
@@ -0,0 +1,76 @@
...
+      | test2        | cus1          | 400  | null       |
+      | test3        | pref          | 500  | null       |
+      | test4        | prf           | 600  | null       |
+
+  Scenario: List user-batch prefixes (default sort)
+    Given I am the user with id "21"
+    When I send a GET request to "/groups/21/user-batch-prefixes"
+    Then the response code should be 200
```
smadbe: "21" is your user, so you cannot have the group you are asking for prefixes be "21" , otherwise we do not what you are testing (and it is not realistic at all, the frontend should never ask the prefix you can use on a single user)

app/api/groups/get_user_batch_prefixes.go
```diff
@@ -0,0 +1,108 @@
...
+	TotalSize int `json:"total_size"`
+}
+
+// swagger:operation GET /groups/{group_id}/user-batch-prefixes groups userBatchPrefixesView
+// ---
+// summary: List user-batch prefixes
+// description: >
+//
+//   Lists the user-batch prefixes  with `allow_new` = 1 matching the input group's ancestors
+//   (i.e., the `group_id` is a descendant of `user_batch_prefixes.group_id`).
```
smadbe: A only managers of the prefix group will be able to use this prefix, only list prefixes whose group has the current user as manager.

app/api/groups/get_user_batch_prefixes.feature
```diff
@@ -0,0 +1,76 @@
...
+
+  Scenario: List user-batch prefixes (default sort, another group_id)
+    Given I am the user with id "21"
+    When I send a GET request to "/groups/13/user-batch-prefixes"
+    Then the response code should be 200
+    And the response body should be, in JSON:
+    """
+    [
+      {"group_id": "13", "group_prefix": "test", "max_users": 90, "total_size": 2},
+      {"group_id": "13", "group_prefix": "test1", "max_users": null, "total_size": 2}
```
smadbe: Probably a test where there is no batch yet would be useful to spot a wrong "join" with batches or such.

app/api/groups/get_user_batch_prefixes.feature
zenovich: fixed in another way

app/api/groups/get_user_batch_prefixes.feature
zenovich: fixed

app/api/groups/get_user_batch_prefixes.go
zenovich: fixed

app/api/groups/get_user_batch_prefixes.feature
zenovich: fixed

app/api/groups/get_user_batch_prefixes.feature
```diff
@@ -0,0 +1,86 @@
...
+      | 16       | 21         | none        |
+    And the database has the following table 'groups_ancestors':
+      | ancestor_group_id | child_group_id |
+      | 13                | 13             |
+      | 13                | 15             |
+      | 13                | 21             |
+      | 14                | 14             |
+      | 15                | 15             |
+      | 16                | 13             |
+      | 16                | 15             |
```
smadbe: This does not match any groups_groups

zenovich: fixed

app/api/groups/get_user_batch_prefixes.feature
```diff
@@ -0,0 +1,85 @@
...
+      | login | group_id |
+      | owner | 21       |
+    And the database has the following table 'group_managers':
+      | group_id | manager_id | can_manage  |
+      | 13       | 21         | memberships |
+      | 16       | 21         | none        |
+    And the database has the following table 'groups_ancestors':
+      | ancestor_group_id | child_group_id |
+      | 13                | 13             |
+      | 13                | 15             |
```
smadbe: This ancestor is not matching `groups_groups` as well 😔
(I still think that for feature tests, we should have a step such as "and the group ancestors match groups_groups" simply building all combination from `groups_groups` definition, to make sure they are no errors (I am sure there are many) and to make feature tests shorter/simpler)

zenovich: Fixed. For the new step, I'll think about it.

app/api/groups/get_user_batch_prefixes.go
```diff
@@ -0,0 +1,114 @@
...
+
+// swagger:model userBatchPrefix
+type userBatchPrefix struct {
+	// required: true
+	GroupPrefix string `json:"group_prefix"`
+	// required: true
+	GroupID int64 `json:"group_id,string"`
+	// Nullable
+	// required: true
+	MaxUsers *int `json:"max_users"`
```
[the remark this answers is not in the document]
zenovich: In the initial request, It is not specified if `max_users` should be nullable or not, also it is not specified if it should have a default value.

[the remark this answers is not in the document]
smadbe: Ok, put it not null with default 1000 then.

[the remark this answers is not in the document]
zenovich: done

PR #490, opened 2020-03-05
Run GroupGroupStore.After() and ItemItemStore.After() after applying DB migrations + optimize AttemptStore.ComputeAllAttempts()

zenovich:
We discovered that some groups_ancestors rows are missing after the initial DB migration. The issues was caused by the fact that the db-migrate command didn't run GroupGroupStore.createNewAncestors(). This PR fixes the db-migrate command so that it will execute GroupGroupStore.createNewAncestors().
Also, it appeared that insertion of missing attempts in AttemptStore.ComputeAllAttempts() worked too slow (took more than 5 hours) after groups_ancestors recalculation. The PR fixes this issue too.

### DIFF
--- a/cmd/db_migrate.go
+++ b/cmd/db_migrate.go
@@ -61,7 +61,8 @@ func init() { // nolint:gochecknoinits
 				_, err = db.Exec("ANALYZE TABLE `attempts`")
 				assertNoError(err, "Cannot execute ANALYZE TABLE")
 				err = database.NewDataStore(gormDB).InTransaction(func(store *database.DataStore) error {
-					assertNoError(store.Attempts().ComputeAllAttempts(), "Cannot compute attempts")
+					assertNoError(store.GroupGroups().After(), "Cannot compute groups_groups")
+					assertNoError(store.ItemItems().After(), "Cannot compute items_items")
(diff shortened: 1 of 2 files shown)


cmd/db_migrate.go
```diff
@@ -61,7 +61,8 @@ func init() { // nolint:gochecknoinits
 				_, err = db.Exec("ANALYZE TABLE `attempts`")
 				assertNoError(err, "Cannot execute ANALYZE TABLE")
 				err = database.NewDataStore(gormDB).InTransaction(func(store *database.DataStore) error {
-					assertNoError(store.Attempts().ComputeAllAttempts(), "Cannot compute attempts")
+					assertNoError(store.GroupGroups().After(), "Cannot compute groups_groups")
+					assertNoError(store.ItemItems().After(), "Cannot compute items_items")
```
smadbe: maybe you should add as comment that `Attempts().ComputeAllAttempts()`  is called by `ItemItems().After(),`

zenovich:
> 👍
> Could you add to the PR some explanation of the problem... that there was missing ancestors, then request taking 5hours, ...
> As there is not related "issues", it may be useful to understand in 1y why you did that change.

done

cmd/db_migrate.go
zenovich: done

PR #504, opened 2020-03-10
Introduce "Skills"

### LINKED ISSUE #415: Introduce "Skills"
## Goal
Introduce "skills", a new type of 'items' which does not correspond to activity (chapter/tasks/course) but to skills with hierarchy of skills and dependencies between skills and other kind of items.

## Subtasks
* add a new type to `items`: "Skill"
* in `add_item`:
  - if type is not 'Skill', don't allow the children to be skills
  - if type is 'Skill', the parent item cannot be a non-skill 
  - never allow the parent item to be a 'Task' or 'Course'
  - if the type is  'Task' or 'Course', do not allow any children
* in `update_item`:
  - if type is not 'Skill', don't allow the children to be skills
  - if the type is  'Task' or 'Course', do not allow any children
  - remove the "type" input, a item cannot change type

zenovich:
Fixes #415

### DIFF
--- a/app/api/items/add_item.feature
+++ b/app/api/items/add_item.feature
@@ -237,3 +237,88 @@ Feature: Add item
       | 11       | 5577006791947779410 | solution           | solution_with_grant      | answer_with_grant   | all_with_grant     | 1                  |
     And the table "groups" should stay unchanged
     And the table "attempts" should stay unchanged
+
+
+  Scenario: Valid when type=Skill
+    Given I am the user with id "11"
+    And the database table 'items' has also the following rows:
+      | id | default_language_tag | type    |
+      | 12 | fr                   | Skill   |
+      | 34 | fr                   | Chapter |
+      | 50 | fr                   | Skill   |
+    And the database table 'permissions_generated' has also the following rows:
+      | group_id | item_id | can_view_generated       | can_grant_view_generated | can_watch_generated | can_edit_generated | is_owner_generated |
+      | 10       | 21      | none                     | content                  | none                | none               | 0                  |
+      | 11       | 12      | content_with_descendants | solution                 | answer              | all                | 0                  |
+      | 11       | 34      | solution                 | solution_with_grant      | answer_with_grant   | all_with_grant     | 0                  |
+      | 11       | 50      | solution                 | solution_with_grant      | answer_with_grant   | all_with_grant     | 0                  |
+    And the database table 'permissions_granted' has also the following rows:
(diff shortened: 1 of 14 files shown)


zenovich:
I'm becoming more and more convinced that the initial idea of considering skills as new kind of entities was correct. It looks like skills are only a bit more similar to items than, for instance, to groups. Implementing skills as items makes us add lots of conditions into the code and much more tests. Even after having all the conditions and validations specified in the parent task done, we still propagate attempts and permissions for skills, still require skills to be visible/editable by the current user on adding/updating, still allow setting values into fields that are not related to skills, still allow viewing skills as usual items and so on and so forth. Who knows how many unforeseen effects we will get afterwards and how many fixes we still have to apply to get skills blended with items finally... Anyway, this makes the code much more complex and harder to maintain.

smadbe:
> we still propagate attempts and permissions for skills, still require skills to be visible/editable by the current user on adding/updating,
Permissions have to be used for skills, and in the same ways as items. 

>  still allow setting values into fields that are not related to skills
This is the same between tasks and chapters. When checking with Mathias, actually we saw that there are more attributes in common between chapters and skills than between chapter and tasks. Of course, we could have 1<->0-1 relationship from `items` to new tables `chapters`, `tasks`, `course`, and `skills` but that is something we already discussed and you were not especially in favor.

> still allow viewing skills as usual items and so on and so forth.
They also need the items navigation (left nav, breadcrumb), unlocks, score for a participants, ... so main feature on items are there.

> Who knows how many unforeseen effects we will get afterwards and how many fixes we still have to apply to get skills blended with items finally... Anyway, this makes the code much more complex and harder to maintain.
If we have to rewrite the same service as items for skills for permissions (big piece!), navigation, unlocks, score, ... we'll have to rewrite 30 services for skills as well, with 95% of the code in common. It will not be a gain of effort and of complexity either.

With the current need on skills, I think it makes sense. Of course, it may depend on how it evolves.

zenovich:
> Looks ok. 👍
> 
> You should just add a word about the constraint in terms of parent-child (not only skills):
> 
> "adds new relations for the parent and (optionally) children items into `items_items` and propagates `permissions_generated`." -> add " The only authorized parent-child relationship are skills-*, chapter-task, chapter-course, chapter-chapter. Otherwise the error ... is returned."
> 
> And similar for update.

done

smadbe:
Sorry, I had forgotten "chapter-chapter" at first and added it after a few minutes.

zenovich:
> Sorry, I had forgotten "chapter-chapter" at first and added it after a few minutes.

fixed

PR #506, opened 2020-03-11
Result propagation: Do not create attempts for explicit-entry items, set latest_activity_at correctly

### LINKED ISSUE #405: Result propagation: Do not create attempt on explicit-entry items
## Motivation
Before entering into a contest or any explicit-entry item, we should not have any score on this item yet.

When the latest activity is set on a chapter, it should not be overridden by children.

## Subtasks
In result propagation: 
- do not insert (but still update of course) new attempt on explicit-entry items
- for attempt created just before propagation (to the attempt of visible items), do not let "latest_activity_at" empty as the default latest_activity_at is "current_timestamp". Set it to "1000-01...", which will be overridden at propagation.
- propagate latest_activity as the max between children_stats.latest_activity_at and target_attempts.latest_activity_at. So instead of `IFNULL(children_stats.latest_activity_at,target_attempts.latest_activity_at)`, we should have something like `GREATEST(IFNULL(children_stats.latest_activity_at,'1000-01...'), target_attempts.latest_activity_at)`

zenovich:
Fixes #405

### DIFF
--- a/app/database/attempt_store_compute_all_attempts.go
+++ b/app/database/attempt_store_compute_all_attempts.go
@@ -50,12 +50,14 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
 				WITH RECURSIVE ancestors (ancestor_item_id, child_item_id) AS (
 					SELECT items_items.parent_item_id, items_items.child_item_id
 					FROM items_items
-					JOIN items ON items.id = items_items.parent_item_id AND NOT items.requires_explicit_entry
+					JOIN items ON items.id = items_items.parent_item_id
+					WHERE NOT items.requires_explicit_entry
 					UNION
 					SELECT items_items.parent_item_id, ancestors.child_item_id
 					FROM items_items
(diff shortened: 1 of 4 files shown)


app/database/attempt_store_compute_all_attempts.go
```diff
@@ -42,16 +42,17 @@ func (s *AttemptStore) ComputeAllAttempts() (err error) {
 
 		// Insert missing attempts for chapters having descendants with attempts marked as 'to_be_recomputed'/'to_be_propagated'.
 		// We only create attempts for chapters which are (or have ancestors which are) visible to the group that attempted
-		// to solve the descendant items.
+		// to solve the descendant items. Chapters with explicit entry (items.entry_participant_type IS NOT NULL) are skipped).
```
smadbe: explicit entry is just `items.requires_explicit_entry` :-)

zenovich: fixed

PR #517, opened 2020-03-25
Implement attempt expiration

### LINKED ISSUE #488: Attempt expiration
## Motivations

* In case a participant attempts a contest second time (once his first trial is over), he will get can_view:content on the contest while it is not supposed to be able to modify the first attempt anymore. 
* When a contest is over, we may give view:content access to everybody to read the tasks again and the submitted answers. 

## Tasks

* Add a new attribute "allow_submissions_until" on attempts (new attempts, not results), datetime, default "9999-12-31...", "Time until which the participant can submit an answer on this attempt"
* When entering into a contest, the created attempt computes `allow_submissions_until` the same way the current group membership expiration is computed.
* When the additional time is changed, update the `allow_submissions_until` of the last attempt (greatest allow_submissions_until) unless the new allow_submissions_until value would be < now.
* On submit / get hint (`CheckSubmissionRights` &co functions), ensure the attempt allow submissions.

(the checks to be updated for multiple-attempts on entering are handled in a separate issue)

zenovich:
Fixes #488

### DIFF
--- a/app/api/contests/set_additional_time.feature
+++ b/app/api/contests/set_additional_time.feature
@@ -6,6 +6,8 @@ Feature: Set additional time in the contest for the group (contestSetAdditionalT
       | 11 | Group A | Class   |
       | 13 | Group B | Other   |
       | 14 | Group B | Friends |
+      | 15 | Club    | Club    |
+      | 16 | Team    | Team    |
       | 21 | owner   | User    |
       | 31 | john    | User    |
       | 33 | item10  | Other   |
@@ -20,6 +22,8 @@ Feature: Set additional time in the contest for the group (contestSetAdditionalT
       | group_id | manager_id |
       | 13       | 21         |
       | 14       | 21         |
+      | 15       | 21         |
+      | 16       | 21         |
       | 31       | 21         |
     And the database has the following table 'groups_groups':
       | parent_group_id | child_group_id | expires_at          |
@@ -30,6 +34,8 @@ Feature: Set additional time in the contest for the group (contestSetAdditionalT
       | 34              | 14             | 9999-12-31 23:59:59 |
       | 36              | 13             | 9999-12-31 23:59:59 |
       | 36              | 14             | 2018-12-31 23:59:59 |
+      | 36              | 15             | 2018-12-31 23:59:59 |
+      | 36              | 16             | 2018-12-31 23:59:59 |
       | 36              | 31             | 9999-12-31 23:59:59 |
     And the database has the following table 'groups_ancestors':
       | ancestor_group_id | child_group_id | expires_at          |
@@ -43,6 +49,8 @@ Feature: Set additional time in the contest for the group (contestSetAdditionalT
       | 13                | 13             | 9999-12-31 23:59:59 |
       | 13                | 14             | 9999-12-31 23:59:59 |
       | 14                | 14             | 9999-12-31 23:59:59 |
+      | 15                | 15             | 9999-12-31 23:59:59 |
+      | 16                | 16             | 9999-12-31 23:59:59 |
       | 21                | 21             | 9999-12-31 23:59:59 |
       | 31                | 31             | 9999-12-31 23:59:59 |
       | 33                | 33             | 9999-12-31 23:59:59 |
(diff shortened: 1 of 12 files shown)


app/api/contests/set_additional_time.go
```diff
@@ -183,7 +183,7 @@ func setAdditionalTimeForGroupInContest(
 		JOIN new_expires_at
 			ON new_expires_at.child_group_id = groups_groups.child_group_id
 		SET groups_groups.expires_at = new_expires_at.expires_at
-		WHERE NOW() < groups_groups.expires_at AND groups_groups.parent_group_id = ?`,
+		WHERE groups_groups.parent_group_id = ?`,
```
smadbe: As the previous request is on `ActiveGroupGroups`, is it enough?
And the access expiration from 2 years ago should not be changed, using ` NOW() < new_expires_at.expires_at` seems more appropriate... mmh actually that's not ok for time reductions, so it should be `(NOW() < new_expires_at.expires_at OR NOW() < groups_groups.expires_at)`. Correct?

app/api/contests/set_additional_time.go
```diff
@@ -199,8 +199,7 @@ func setAdditionalTimeForGroupInContest(
 			attempts.id =
 				(SELECT id FROM (
 					SELECT MAX(id) AS id FROM attempts WHERE participant_id = new_expires_at.child_group_id AND root_item_id = ? FOR UPDATE
-				) AS latest_attempt) AND
-			NOW() < allows_submissions_until
+				) AS latest_attempt)
```
smadbe: same remark

app/api/contests/set_additional_time.go
zenovich: > As the previous request is on ActiveGroupGroups, is it enough?

fixed

app/api/contests/set_additional_time.go
```diff
@@ -186,7 +186,23 @@ func setAdditionalTimeForGroupInContest(
 		WHERE NOW() < groups_groups.expires_at AND groups_groups.parent_group_id = ?`,
 		participantsGroupID)
 	service.MustNotBeError(result.Error())
-	if result.RowsAffected() > 0 {
+
+	groupsGroupsModified := result.RowsAffected() > 0
+
+	service.MustNotBeError(store.Exec(`
```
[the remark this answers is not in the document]
zenovich: done

app/api/contests/set_additional_time.go
zenovich: > And the access expiration from 2 years ago should not be changed, using ` NOW() < new_expires_at.expires_at` seems more appropriate... mmh actually that's not ok for time reductions, so it should be `(NOW() < new_expires_at.expires_at OR NOW() < groups_groups.expires_at)`. Correct?

IMO if we change the additional change, we should update all the affected relations/attempts, it doesn’t matter how old they are. I don’t see the reason to have some custom handling for old expirations: when expires_at changes from -2y to -2y+1h, is it a problem?

smadbe: It is a bit of a problem. First because if we want to understand some end of contest submissions, it is good to know when the user was actually over for the user, this time added afterwards may be misleading (mainly for the attempt, on the perm is less a concern). Second because of the reason of the previous comment, it may cause two attempts to become active while it should not be possible. :-/

zenovich: > As the previous request is on ActiveGroupGroups, is it enough?

No it wouldn't be enough. We want to update only groups_groups for the given (parent_group_id, child_group_id) pair, not all groups_groups with given child_group_id.

zenovich: Fixed both

app/api/contests/set_additional_time.go
zenovich: fixed

app/api/contests/set_additional_time.feature
```diff
@@ -221,6 +242,124 @@ Feature: Set additional time in the contest for the group (contestSetAdditionalT
     And the table "attempts" should stay unchanged
     And the table "results" should stay unchanged
 
+  Scenario: Doesn't update attempts.allows_multiple_submissions if both old and new values are in the past
```
smadbe: How `allows_multiple_submissions` is involved here?  You mean `allows_submissions_until`?

zenovich: fixed

PR #529, opened 2020-04-06
Do not allow user groups as containers

### LINKED ISSUE #508: Ensure group_id are not users when given for group operations
## Motivations
It seems that at multiple places, the type of the input group are not checked even while the operation are not compatible with "users", for instance for adding children to them. 

## Changes

_I'm not 100% sure here, it is just by quickly reading the code, I may have miss some checks._

### createUserBatch
We do not check that the group_id in subgroups are valid groups to put user in. 
Check that these groups are not of type "User".

### CreateRelationsWithoutChecking function
`CreateRelationsWithoutChecking` function comments should mention that the caller must ensure that the parent groups are not users. 

### getGroup
Should exclude users from dest. They may be returned as "ManagedByUser" is used in the query.

### createGroup
Prevent creation of team, but not users?!

### change code, discard code
Should not be applicable to users

### all request, invite, remove/add members services
Check the the container cannot be a user.

zenovich:
Fixes #508

### DIFF
--- a/app/api/groups/discard_code.go
+++ b/app/api/groups/discard_code.go
@@ -17,7 +17,7 @@ import (
 //
 //
 //   The authenticated user should be a manager of `group_id` with `can_manage` >= 'memberships',
-//   otherwise the 'forbidden' error is returned.
+//   otherwise the 'forbidden' error is returned. If the group is a user group, the 'forbidden' error is returned as well.
 // parameters:
 // - name: group_id
 //   in: path
--- a/app/api/groups/discard_code.go
+++ b/app/api/groups/discard_code.go
@@ -17,7 +17,7 @@ import (
 //
 //
 //   The authenticated user should be a manager of `group_id` with `can_manage` >= 'memberships',
-//   otherwise the 'forbidden' error is returned. If the group is a user group, the 'forbidden' error is returned as well.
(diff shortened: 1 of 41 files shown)


zenovich:
There are still some services that don't check that the group_id group is not a user group although they could:
1. Read-only services like groupGroupProgress, groupChildrenView, groupsMembersView, groupTeamDescendantView, groupTeamProgress, groupUserDescendantView, groupUserProgress return empty results when group_id is a user group.
2. groupLeaveRequestWithdraw, groupJoinRequestWithdraw, groupInvitationReject, groupLeave will do nothing because it is impossible to become a member of a user group or to create a pending request with a user group as a parent group.

app/api/groups/discard_code.go
```diff
@@ -17,7 +17,7 @@ import (
 //
 //
 //   The authenticated user should be a manager of `group_id` with `can_manage` >= 'memberships',
-//   otherwise the 'forbidden' error is returned.
+//   otherwise the 'forbidden' error is returned. If the group is a user group, the 'forbidden' error is returned as well.
```
smadbe, line 20: The term "user group" that you use everywhere is a bit misleading as a class is "user group" (in the sense "group of users"). I think it will be clearer if you just write "If the group is a user, ..."

zenovich: fixed

PR #539, opened 2020-04-14
Remove reverse proxy

smadbe:
Remove the reverse proxy which not useful anymore. Put a (tested) 404 when path not found.


app/api/not-found.feature
```diff
@@ -1,7 +1,12 @@
 Feature: Not found fallback
 
 Scenario: A request to a not found url is redirected to the fallback server
```
zenovich, line 3: The scenario name should be updated as well

PR #540, opened 2020-04-16
Implement frozen membership 

### LINKED ISSUE #491: Team frozen membership
## Motivations

When used for contests, teams may be allowed to change their composition while the contest has already started, or may not. (In case it may, a super admin may disallow it after some time manually). In any case, the team may add/remove members before the contest start.

So, team can "freeze their composition" (non revertible) when entering some contests that requires it.

## Tasks

* Create new attribute `groups.frozen_membership`, bool, default false, "Whether members can be added/removed to the group (intended for teams)"
* Rename `teams_editable` to (**inverted value!**) `items.entry_frozen_teams`, bool, default false, "Whether teams require to be have \`frozen_membership\` for entering"
* in entry checks: add if `items.entry_frozen_teams` and the entering group is a team, it must have `groups.frozen_membership`.
* in groupEdit, let the manager change `frozen_membership`, but only from 0 to 1. Should also auto-reject all pending join/leave requests and auto withdraw pending invites.
* in itemEdit, add entry_frozen_teams
* if `groups.frozen_membership`, disallow invitation creation/ acceptation, request creation/accept, join by code, leave request creation/accept, leave, direct add

zenovich:
Fixes #491


app/api/contests/get_qualification_state.go
```diff
@@ -151,12 +160,13 @@ func (srv *Service) getContestInfoAndQualificationStateFromRequest(r *http.Reque
 		AllowsMultipleAttempts   bool
 		ContestMaxTeamSize       int32
 		ContestEnteringCondition string
+		EntryFrozenTeams         bool
```
smadbe, line 163: Not really clear what a bool called "EntryFrozenTeams" means.

app/api/groups/update_group.feature
```diff
@@ -210,3 +216,38 @@ Feature: Update a group (groupEdit)
...
+    }
+    """
+    Then the response should be "updated"
+    And the table "groups" should stay unchanged but the row with id "14"
+    And the table "groups_groups" should stay unchanged
+    And the table "group_pending_requests" should be:
+      | group_id | member_id | type          |
+      | 14       | 31        | join_request  |
+    And the table "group_membership_changes" should be:
+      | group_id | member_id | action                | ABS(TIMESTAMPDIFF(SECOND, at, NOW())) < 3 |
```
smadbe, line 249: initiator column may worth checking maybe

app/api/contests/get_qualification_state.go
zenovich: Actually the column name has the same problem...

zenovich: The fields of this stucture have names of the related db columns, also they are not part of the output. Do we still want to rename 'EntryFrozenTeams'?

app/api/groups/update_group.feature
zenovich: fixed

PR #541, opened 2020-04-17
Implement updated login workflow

### LINKED ISSUE #518: Updated login workflow
## Motivations
The login workflow has changed a bit to make it more "standard" (without trying to stay close to what was done on AlgoreaPlatform). It will make other future workflows (LTI, temp user using group code,...) cleaner and more secure. In a nutshell, we let the frontend deal directly with the login-module for the pure authentication part. The backend is only involved for obtaining the token (the SPA may get the token directly as well, but it needs the backend to know this token anyway).

The new workflow is on: https://france-ioi.github.io/algorea-devdoc/auth/login-workflow/

## Subtasks
* Remove the login state handling and table
* Remove the `/login` service
* Remove  `/auth/login-callback` to be merged in `POST /auth/token` (see next line): as it is not really a callback and its role is now  (from the frontend point of view) only to provide a token from a code (even it has other side effect for the backend)
* Add this new case to `POST /auth/token`:
  - taking as input a mandatory `code` attribute and an optional `code_verifier`. These are passed (if given) to the login-module for getting the token (as before).  
  -  Then the user profile is fetched, as in it was done before `/auth/login-callback`
* So the new `POST /auth/token` has to refresh token as before if a token in auth header is given, and just generate a new one if `code` is given.  To prevent frontend bug, it is probably sane to fail if both the auth header and a code is given.

zenovich:
Fixes #518


app/api/auth/create_access_token.go
```diff
@@ -1,119 +1,169 @@
...
-// description: Creates a new access token (locally for temporary users or via the login module for normal users) and
-//              saves it in the DB keeping only the input token (from authorization headers) and the new token.
-//              Since the login module responds with both access and refresh tokens, the service updates the user's
-//              refresh token in this case as well.
+// summary: Create or refresh an access token
+// description:
+//     If the "Authorization" header is not given, the service converts the given code into tokens,
+//     creates or updates the authenticated user in the DB with the data returned by the login module,
+//     and saves new access & refresh tokens into the DB as well.
+//
```
smadbe, line 27: There should be a word about `code_verifier`:

> If the "Authorization" header is not given, the service converts the given OAuth2 authorization code into tokens,  creates or updates the authenticated user in the DB with the data returned by the login module,
> and saves new access & refresh tokens into the DB as well.
> If OAuth2 authentication has used the PKCE extension, the '{code_verifier}' should be provided so it can be sent together with the '{code}' to the authentication server.

app/api/auth/create_access_token.go
```diff
@@ -1,119 +1,169 @@
...
-		service.MustNotBeError(userIDsInProgress.withLock(user.GroupID, r, func() error {
-			newToken, expiresIn = srv.refreshTokens(r.Context(), user, oldAccessToken)
-			return nil
-		}))
+	if len(r.Header["Authorization"]) != 0 {
+		if len(r.URL.Query()["code"]) != 0 {
+			return service.ErrInvalidRequest(
+				errors.New("only one of the 'code' query parameter and the 'Authorization' header can be given"))
+		}
+		auth.UserMiddleware(srv.Store.Sessions())(service.AppHandler(srv.refreshAccessToken)).ServeHTTP(w, r)
```
smadbe, line 63: you should put a comment to summarize in which case you are at the end on this line:
> // "Authorization" header given, requesting a new token from the given token

app/api/auth/create_access_token.go
```diff
@@ -1,119 +1,169 @@
...
+		return service.NoError
+	}
+
```
smadbe, line 66: > // `code` given, requesting a token from `code` and optionally `code_verifier`, and create/update user.

app/api/auth/create_access_token.go
zenovich: fixed

app/api/auth/create_access_token.go
zenovich: fixed

app/api/auth/create_access_token.go
zenovich: fixed

PR #544, opened 2020-04-23
Reorganization of config

smadbe:
Make config agnostic of attributes of each package. Merge config pkg within the app pkg. 
Cascade the changes to the related packages. 

The way config are passed or not passed to each service has not been rethought yet, that's not the goal of this preliminary work. I would like to have that already merged before working further on it.


smadbe:
Tests are passing but CircleCI->GH fails. (GH has problems for the moment)

zenovich:
Looks like the tests are not isolated well:
```
$ for s in $(go list ./... | grep -v /vendor/); do if ! go test -gcflags=all=-l -failfast -v -p 1 $s; then break; fi; done;
2020/04/23 20:17:40 Loading environment: test
testing: warning: no tests to run
PASS
ok  	github.com/France-ioi/AlgoreaBackend	0.102s [no tests to run]
=== RUN   TestNew_Success
2020/04/23 20:17:46 Loading environment: test
[mysql] 2020/04/23 20:17:46 driver.go:123: could not use requested auth plugin 'mysql_native_password': this user requires mysql native password authentication.
ERRO this user requires mysql native password authentication.  module=database
--- FAIL: TestNew_Success (0.04s)
    app_test.go:31: 
        	Error Trace:	app_test.go:31
        	Error:      	Expected value not to be nil.
        	Test:       	TestNew_Success
    app_test.go:32: 
        	Error Trace:	app_test.go:32
        	Error:      	Received unexpected error:
        	            	this user requires mysql native password authentication.
        	Test:       	TestNew_Success
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x16400a4]

goroutine 7 [running]:
testing.tRunner.func1(0xc0001ebb00)
	/usr/local/opt/go/libexec/src/testing/testing.go:874 +0x3ae
panic(0x16fba20, 0x1d4c8c0)
	/usr/local/opt/go/libexec/src/runtime/panic.go:679 +0x1e0
github.com/France-ioi/AlgoreaBackend/app.TestNew_Success(0xc0001ebb00)
	/Users/dzenovich/gowork/src/github.com/France-ioi/AlgoreaBackend/app/app_test.go:33 +0xd4
testing.tRunner(0xc0001ebb00, 0x181a618)
	/usr/local/opt/go/libexec/src/testing/testing.go:909 +0xda
created by testing.(*T).Run
	/usr/local/opt/go/libexec/src/testing/testing.go:960 +0x331
FAIL	github.com/France-ioi/AlgoreaBackend/app	0.084s
FAIL
```
(we didn't have this issue before)

zenovich:
Great!

There is just a little issue inside the Viper which wasn't an issue before, but seems to be an issue after the reorganization as now we have those `Config.Get()` calls right in service methods. Viper's documentation says `One important thing to recognize when working with ENV variables is that the value will be read each time it is accessed.` and Viper really does. The problem here is that env variable reading is very slow  (https://dinolai.com/notes/golang/golang-benchmark-reading-config-from-os-env-and-file.html). Actually I don't know why they do this since it's impossible to change an env variable value of a running process without restarting... 

Also, there could be another issue if we used Viper's config files live reading. Since we now have multiple `Config.Get()` calls right in services, it could be a failure if Viper decided to reload the config somewhere in the middle of those calls during a request processing. It would be much more cleaner to load all the config values on start up without changing on the fly. If we would like to reload config we could handle a signal and reload the config values under lock.

smadbe:
Ok, it could be a further evolution (for a separate PR). Anyway, we also need to know how we'll deploy the app. The best solutino for config handling may be a bit different if it is using lambdas and a classic server.

PR #547, opened 2020-04-27
Allow participations ending

### LINKED ISSUE #493: Allow ending a participation
## Motivations

In multi-attempts contest in particular, a team may want to stop a contest to allow its members to participate as other teams. Also, in a explicit-entry item with no duration (but with a start date), a user may want to end a participation to stop the timer.

## Tasks

* New attribute: `attempts.ended_at`, datetime, nullable, "Time at which the attempt was (typically manually) ended. "
* New service `POST /attempts/{attempt_id}/leave" + optional query parameter `as_team_id` which:
   - checks that `ended_at` is null, `allows_submissions_until` > now for the given attempt
   - sets `results.ended_at` and `allow_submissions_until` to `min(now, current value)`  for the given attempt and all descendants attempts (those which have this attempt as parent_attempt_id, and children of children, so with a CTE?)
   - updates expiration of the participant group membership accordingly
* Make sure additional time change does no change submission time (and perm) when the attempt has ended

zenovich:
Fixes #493


app/api/items/items.go
```diff
@@ -45,6 +45,7 @@ func (srv *Service) SetRoutes(router chi.Router) {
 	router.Get("/items/{item_id}/entry-state",
 		service.AppHandler(srv.getEntryState).ServeHTTP)
 	router.Post("/attempts/{attempt_id}/items/{item_id}/enter", service.AppHandler(srv.enter).ServeHTTP)
+	router.Delete("/attempts/{attempt_id}", service.AppHandler(srv.endAttempt).ServeHTTP)
```
smadbe, line 48: It is not really deleting attempt, as we may allow to visit again the attempt but in read-only and allow teach to grade it. I think it would be misleading to use a delete here. I prefer to keep "POST actions" as others, so something like `POST /attempts/{attempt_id}/end` (or `leave`)

app/api/items/leave.feature
```diff
@@ -0,0 +1,149 @@
...
+      | 1  | 101            | 50           | null              | 9999-12-31 23:59:59      | null                |
+      | 1  | 102            | 50           | null              | 9999-12-31 23:59:59      | null                |
+      | 1  | 111            | 50           | null              | 9999-12-31 23:59:59      | null                |
+      | 2  | 101            | 60           | 0                 | 9999-12-31 23:59:59      | null                |
+      | 2  | 102            | 60           | 0                 | 2019-12-31 23:59:59      | 2019-12-31 23:59:59 |
+      | 2  | 111            | 60           | 0                 | 9999-12-31 23:59:59      | null                |
+
+  Scenario: User is able to end an attempt for his self group
+    Given I am the user with id "111"
+    When I send a DELETE request to "/attempts/0"
```
smadbe, line 58: Ah it is a good example of something we should probably not allow. It should not be allowed to end the default attempt (id 0), otherwise the user is completely stuck. Anyway it does not really make sense to end attempt 0 which was implicit.

app/api/items/leave.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/go-chi/render"
+
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+// swagger:operation DELETE /attempts/{attempt_id} items itemAttemptEnd
+// ---
+// summary: Leave the item
```
smadbe, line 15: `itemAttemptEnd`<> "Leave the item" , `ended_at` fct <> `leave.go`
We should decide if it is leave or end. I know I proposed "leave" in the issue, but intuitively when reading this service and as the attribute is `ended_at`, I find `end` clearer.

app/api/items/items.go
zenovich: fixed

app/api/items/leave.feature
zenovich: fixed

app/api/items/leave.go
zenovich: fixed

zenovich:
> Ah this is still written "DELETE /attempts/{attempt_id}" in the swagger header of the service, to be fixed!

fixed

PR #551, opened 2020-04-30
Remove server.domain from sample config

smadbe:
Remove server.domain from sample config which is not used (anymore)


zenovich:
> This "server.domain" (not the root "domain") is not used at all by code.

ah. okay then

PR #554, opened 2020-05-04
Config: load default

### LINKED ISSUE #63: Only config with a file default can be set by ENV
For the moment, only a configuration entry which has been defined in the config yml file (whatever its value) can be defined as an ENV variable. If not in the yml, the env var is not retrieved.

smadbe:
Fixes #63 
For DB config, the only one which is unmarshalled and which can be overriden by env var (and very typical for DB config), do not require a default value to be put in the config file to load the env variables.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -110,9 +110,10 @@ func subconfig(globalConfig *viper.Viper, subconfigKey string) *viper.Viper {
 
 // DBConfig returns the db connection fixed config from the global config.
 func DBConfig(globalConfig *viper.Viper) (config *mysql.Config, err error) {
-	// Env variables are not loaded if the key do not exist in the config file
-	// To fix this issue, all expected attributes are loaded in the config and
-	// are overridden by what is defined in the config (file or env)
+	// Env variables are not loaded if the keys do not exist in the config file
+	// To fix this issue, instead of loading config files and overriding with env vars,
+	// we load all possible keys (with their default value), override with config files,
+	// and then environmenent variables.
 	emptyConfig := &map[string]interface{}{}
 	if err = mapstructure.Decode(mysql.NewConfig(), &emptyConfig); err != nil {
 		return // unexpected
--- a/app/config.go
+++ b/app/config.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"strings"
 
+	"github.com/France-ioi/mapstructure"
 	"github.com/go-sql-driver/mysql"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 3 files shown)


app/config.go
```diff
@@ -108,10 +109,22 @@ func subconfig(globalConfig *viper.Viper, subconfigKey string) *viper.Viper {
 }
 
 // DBConfig returns the db connection fixed config from the global config.
-// Panic in case of unmarshaling error
 func DBConfig(globalConfig *viper.Viper) (config *mysql.Config, err error) {
-	sub := subconfig(globalConfig, databaseConfigKey)
-	err = sub.Unmarshal(&config)
+	// Env variables are not loaded if the key do not exist in the config file
```
zenovich, line 113: 'does not exist'?

app/config.go
```diff
@@ -108,10 +109,22 @@ func subconfig(globalConfig *viper.Viper, subconfigKey string) *viper.Viper {
...
+	// To fix this issue, all expected attributes are loaded in the config and
```
zenovich, line 114: > 'all expected attributes are loaded in the config'
can't get what it means

smadbe:
I tried to improve explanation. 
BTW, it has also an interesting benefits: before we extracting the mysql.Config struct raw from the config, so using the "Golang default" for attributes which were not configured. But now we start from `mysql.NewConfig()` and so from some default value which looks more consistent for MySQL.  (not 100% sure that it was not handled correctly somehow anyway, but cannot harm :-))

app/config.go
```diff
@@ -108,15 +109,27 @@ func subconfig(globalConfig *viper.Viper, subconfigKey string) *viper.Viper {
...
 func DBConfig(globalConfig *viper.Viper) (config *mysql.Config, err error) {
-	sub := subconfig(globalConfig, databaseConfigKey)
-	err = sub.Unmarshal(&config)
+	// Env variables are not loaded if the keys do not exist in the config file
+	// To fix this issue, instead of loading config files and overriding with env vars,
+	// we load all possible keys (with their default value), override with config files,
+	// and then environmenent variables.
+	emptyConfig := &map[string]interface{}{}
+	if err = mapstructure.Decode(mysql.NewConfig(), &emptyConfig); err != nil {
+		return // unexpected
```
zenovich, line 119: If it is really unexpected, it's better just to ignore the error (`_ = mapstructure.Decode(...)`)

app/config_test.go
```diff
@@ -110,23 +113,38 @@ func TestDBConfig_Success(t *testing.T) {
...
+	_, err := DBConfig(globalConfig)
+	assert.EqualError(err, "unmarshalling error")
+}
+
+func TestDBConfig_StructToMapError(t *testing.T) {
+	// unexpected error, must monkey patch it
+	assert := assertlib.New(t)
+	globalConfig := viper.New()
+	monkey.Patch(mapstructure.Decode, func(_ interface{}, _ interface{}) error {
+		return fmt.Errorf("struct2map error")
```
zenovich, line 143: errors.New(...)?

smadbe: Is it really better? Both are used in the code. I can change it.

app/config.go
smadbe: It is unexpected at runtime but if mysql.NewConfig() or mapstructure change their implem, it may fail. The "problem" if I ignore it is that the variable will keep its default value and may be undetected.

app/config_test.go
smadbe: I see that many other places in the code are doing this way, so I'll keep it here as well (it is a test anyway).

PR #556, opened 2020-05-05
Load configs from cwd/conf/ for normal builds and from project_root/conf for test builds, load tokens specified by relative path similarly

zenovich:
The 'project_root' is calculated by stripping the "/app/..." suffix from the current working directory for test builds.

Checked with 
* make test
* make && ./bin/Algoreabackend serve
* go test -gcflags=all=-l -failfast -v -p 1 -run TestGroupStore ./app/database
* godog --stop-on-failure ./app/api/currentuser/join_group_by_code.robustness.feature

### DIFF
--- a/app/api/answers/bdd_test.go
+++ b/app/api/answers/bdd_test.go
@@ -5,11 +5,9 @@ package answers_test
 import (
 	"testing"
 
-	"github.com/France-ioi/AlgoreaBackend/app"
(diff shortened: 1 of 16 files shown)


zenovich:
> And if somebody put the app in an "app" directory (or has app anywhere in the path) will it work correctly? You should match the last occurrence only, otherwise someone one day will loose 2days debugging it.

Nice catch! Fixed.

Issue #536, opened 2020-04-09
Unified naming


smadbe:
answers
- [ ] postGenerate an answer token
- [ ] getGet an answer
- [ ] getList answers
- [ ] postSave an answer
- [ ] putUpdate current answer
auth
- [ ] postUser login (redirect)
- [ ] getUser login callback
- [ ] postUser logout
- [ ] postCreate a temporary user
- [ ] postRequest a new access token
contests
- [ ] postEnter the contest
- [ ] putSet additional time for a contest
- [ ] getList additional times on a contest
- [ ] getGet a group by name
- [ ] getGet qualification state
- [ ] getList administered contests
users
- [ ] delDelete the current user
- [ ] getGet profile info for the current user
- [ ] getExport the short version of the current user's data
- [ ] getExport the current user's data
- [ ] putUpdate user's notification read date
- [ ] putUpdate the local user info cache
groups
- [ ] getSearch for available groups
- [ ] getGet current user's team for a contest
- [ ] postCreate a group
- [ ] getGet group info
- [ ] putUpdate group information
- [ ] delDiscard the group’s code
- [ ] postGenerate a new group code
- [ ] getGet group progress
- [ ] getList group managers
- [ ] delRemove a group manager
- [ ] postMake user a group manager
- [ ] putChange permissions of a group manager
- [ ] getGet recent activity of a group
- [ ] getGet group progress for teams
- [ ] getList user-batch prefixes
- [ ] getGet group progress for users
- [ ] putUpdate permissions
- [ ] postCreate a user batch
- [ ] delRemove a user batch
- [ ] getList user batches
group-memberships
- [ ] getList current invitations and requests to groups
- [ ] postAccept a group invitation
- [ ] postReject a group invitation
- [ ] delWithdraw a leave request
- [ ] postCreate a leave request
- [ ] getList current user's groups
- [ ] getGet invitations/requests for the current user
- [ ] delLeave a group
- [ ] postJoin a team using a code
- [ ] delWithdraw a request to join a group
- [ ] postCreate a request to join a group
- [ ] getList group's children
- [ ] delRemove members from a group
- [ ] getList group members
- [ ] getList pending requests and invitations for a group
- [ ] getList group's team descendants
- [ ] getList group's user descendants
- [ ] postInvite users to a group
- [ ] postWithdraw group invitations
- [ ] postAccept requests to join a group
- [ ] postReject requests to join a group
- [ ] postAccept leave request
[...]

PR #557, opened 2020-05-07
Unify naming groups part1

smadbe:
Few more services for cleaning the namings (#536).

### DIFF
--- a/app/api/currentuser/search_for_available_groups.go
+++ b/app/api/currentuser/search_for_available_groups.go
@@ -14,7 +14,7 @@ import (
 
 const minSearchStringLength = 3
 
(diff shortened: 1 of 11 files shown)


app/api/currentuser/search_for_available_groups.go
```diff
@@ -14,9 +14,9 @@ import (
 
 const minSearchStringLength = 3
 
-// swagger:operation GET /current-user/available-groups groups groupsAvailableSearch
+// swagger:operation GET /current-user/available-groups groups groupSearch
```
zenovich, line 17: I would prefer `groupsSearch` here since the service returns multiple groups

PR #560, opened 2020-05-12
Implement the resultStart service

### LINKED ISSUE #410: Service to "start" a result
## Overview

As described in the [UI-workflow](https://france-ioi.github.io/algorea-devdoc/items/ui-workflow/), results have to be existing and started so that a chapter can be expanded (in the nav menu) and to load an attempt on a item.

This service create a result if not existing, and start it.

On `POST /items/{i1_id}/{i2_id}/.../start-result?attempt_id=..`

## Actions
* if there is no results for attempt, item, and participant (current user or `as_team_id`), create one
* set `started_at` and `latest_activity_at` to now() (propagate it)
* if `started_at` was already set, leave it as it is (silently)

## Input
* attempt_id (in url)
* item_id (in url)
* as_team_id (query string, optional)

## Validations
  - that the first item is a root (`items.is_root`) or the item of a group (`groups.activity_id`) the current user is member of  
  - the path is ordered list on parent-child items with view>=content perm by the user
  - that the current participant has a started, allowing submission, not ended (if not implied by allowing-submissions) result on each item but last, so from the item before last to the first with, as attempt, the input attempt (or its parent attempt each time we reach a root of an attempt)
  - in case only one item is provided, these same rules applies except that the attempt must be "0"
  - check that items.requires_explicit_entry is false for the last item

zenovich:
Fixes #410

### DIFF
--- a/app/api/items/items.go
+++ b/app/api/items/items.go
@@ -46,6 +46,7 @@ func (srv *Service) SetRoutes(router chi.Router) {
 		service.AppHandler(srv.getEntryState).ServeHTTP)
 	router.Post("/items/{ids:(\\d+/)+}enter", service.AppHandler(srv.enter).ServeHTTP)
 	router.Post("/attempts/{attempt_id}/end", service.AppHandler(srv.endAttempt).ServeHTTP)
+	router.Post("/items/{ids:(\\d+/)+}start-result", service.AppHandler(srv.startResult).ServeHTTP)
 }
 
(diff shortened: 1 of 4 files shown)


zenovich:
Note that I added a requirement "the last item in `{ids}` should be either 'Task', 'Course', or 'Chapter'". Let me know if it is wrong.

smadbe:
> Note that I added a requirement "the last item in `{ids}` should be either 'Task', 'Course', or 'Chapter'". Let me know if it is wrong.

It will not be correct when we'll have skills. But as skills are not fully supported, we can leave it like this.

PR #566, opened 2020-05-25
 GetGroup: add level of management

### LINKED ISSUE #564: GetGroup: add level of management
## Motivations
Be able to display specific action depending on what is allowed for the user

## Subtasks
* Either create a new service or add in GetGroup the management info of the current user, so his `can_manage` level, `can_grant_group_access` and `can_watch_members`.
* I let you choose how you want to add it to make it clean.

zenovich:
Fixes #564


app/database/db_enums.go
```diff
@@ -0,0 +1,104 @@
...
+		"permissions_granted.can_watch",
+		"permissions_granted.can_edit",
+		"group_managers.can_manage",
+	}
+	enumName2Number     map[string]int
+	enumValueName2Index []map[string]int
+	enumValueIndex2Name []map[int]string
+)
+
+func loadDBEnum(db *DB, fullColumnName string) {
```
smadbe, line 23: why not `func (db *DB) loadEnum(fullColumnName string) ` ? (same for other functions below)

app/api/groups/get_group.feature
```diff
@@ -82,6 +87,9 @@ Feature: Get group by groupID (groupView)
       "code_expires_at": "2017-10-13T05:39:48Z",
       "open_activity_when_joining": true,
       "current_user_is_manager": true,
+      "can_manage": "memberships_and_group",
+      "can_grant_group_access": true,
+      "can_watch_members": true,
```
smadbe, line 92: actually, shouldn't we call these attributes `current_user_can_...` to be consistent with "current_user_is_manager" and as it is not a group attribute?

app/database/db_enums.go
zenovich: fixed

app/api/groups/get_group.feature
zenovich: fixed

PR #581, opened 2020-08-05
Rework root groups

### LINKED ISSUE #520: Rename Root*Group config param
## Motivations
RootGroup, RootSelfGroup, RootTempGroup are not clear. 

## Subtasks
* rename "RootGroup" -> AllUsersGroup
* rename "RootSelfGroup" -> NonTempUsersGroup, (or PersistentUsersGroup but it is less clear)
* rename "RootTempGroup" -> TempUsersGroup

(I wonder if we should really keep this RootGroup as it is just the parent group of the 2 others that the domain creator can organize as he wants)

zenovich:
Fixes #520

### DIFF
--- a/db/migrations/2008040105_rework_root_group.sql
+++ b/db/migrations/2008040105_rework_root_group.sql
@@ -1,0 +1,13 @@
+-- +migrate Up
+DELETE FROM `groups` WHERE `type`='Base' AND `text_id`='Root';
+UPDATE `groups` SET `name`='AllUsers', `text_id`='AllUsers', `description`='AllUsers' WHERE `type`='Base' AND `text_id`='RootSelf';
+UPDATE `groups` SET `name`='TempUsers', `text_id`='TempUsers' WHERE `type`='Base' AND `text_id`='RootTemp';
+
+-- +migrate Down
+SET @id = FLOOR(RAND(11) * 1000000000) + FLOOR(RAND(12) * 1000000000) * 1000000000;
+INSERT INTO `groups` (`id`, `name`, `type`, `text_id`, `description`) VALUES (@id, 'Root', 'Base', 'Root', 'Root');
+INSERT INTO `groups_groups` (parent_group_id, child_group_id)
+    SELECT @id, id FROM `groups` WHERE `type`='Base' AND `text_id`='RootSelf';
+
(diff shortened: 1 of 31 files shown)


db/migrations/2008040105_rework_root_group.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+DELETE FROM `groups` WHERE `type`='Base' AND `text_id`='Root';
```
smadbe, line 2: In the anonymized db, there were no permission granted to this group?

zenovich: There were some, but they disappeared completely after the DB migration

zenovich: > There were some, but they disappeared completely after the DB migration

Sorry. This was a wrong statement. Actually there are 80 rows in permissions_granted, all with the 'content' view permissions (probably, first time I was checking after removal of the root group). So, these 80 permissions should be copied for the 'AllUsers' group.

zenovich: Fixed

db/migrations/2008040105_rework_root_group.sql
```diff
@@ -1,4 +1,15 @@
 -- +migrate Up
+SELECT `id` INTO @root_id FROM `groups` WHERE `type`='Base' AND `text_id`='Root';
+SELECT `id` INTO @root_self_id FROM `groups` WHERE `type`='Base' AND `text_id`='RootSelf';
```
smadbe, line 3: What will happens if there are several ones?
As there may be several domains, we could have several roots, one per domain.

However, if there is only one in the anonymized we could consider it is ok this way (even if I would prefer it to fail if there are more than one instead of doing something completely wrong)

zenovich: There is only one group with (`type`='Base' AND `text_id`='Root') and only one group with (`type`='Base' AND `text_id`='RootSelf') in the anonymized DB. These SELECT statements would fail otherwise:
```
mysql> SELECT `id` INTO @root_id FROM `groups` WHERE `type`='Base';
ERROR 1172 (42000): Result consisted of more than one row
```

PR #583, opened 2020-08-06
Introduce a new cli command db-recompute

zenovich:


zenovich:
> Didn't you want to make them run even if there are no migration to apply?

I think there is need in that since now we have a separate command recalculating caches.

PR #595, opened 2020-08-31
Implement groupsCodeCheck service

### LINKED ISSUE #511: New service: group code check
## Overview 
When a temp user wants to join a group by code, it first has to create a non-temp user for it. 
To prevent creating a user if the group code is wrong,  allow checking a group code without applying it.

On `GET /groups/{group_id}/is-code-valid?code=...`

## Input
* code: required string

## Output
* valid: true|false

## Validations
* none

zenovich:
\+ Restore missing locking in SQL subquries
\+ Disallow SQL wildcards processing for 'code' parameter of groupsJoinByCode & 'name' parameter of contestGetGroupByName
\+ Slightly fix swagger docs

Fixes #511


app/api/contests/get_group_by_name.go
```diff
@@ -122,11 +122,11 @@ func (srv *Service) getGroupByName(w http.ResponseWriter, r *http.Request) servi
 			Joins(`
 				LEFT JOIN `+"`groups`"+` AS user_group
 					ON user_group.id = groups_groups_active.child_group_id AND user_group.type = 'User' AND
-						user_group.name LIKE ?`, groupName).
+						user_group.name = ? AND LENGTH(user_group.name) = LENGTH(?)`, groupName, groupName).
```
smadbe, line 125: I don't really get what you are trying to avoid. 
If you want exact matching, why do you check length in addition? Why having used 'like' in the first place?

app/api/groups/check_code.go
```diff
@@ -0,0 +1,92 @@
...
+//   The service returns false:
+//
+//   * if there is no team with `is_public` = 1, `code_expires_at` > NOW() (or NULL), and `code` = `{code}`;
+//
+//   * if the group is a team and the user is already on a team that has attempts for same contest
+//     while the contest doesn't allow multiple attempts or that has active attempts for the same contest,
+//     or if the group membership is frozen;
+//
+//   * if there is already a row in `groups_groups` with the found team as a parent
+//     and the user’s selfGroup’s id as a child;
```
smadbe, line 34:
```suggestion
//     and the user’s id as a child;
```

smadbe (commented):
A question.

app/api/contests/get_group_by_name.go
zenovich: `LIKE` was used by mistake here. The `LENGTH()` is needed because MySQL stips trailing spaces in comparison for non-binary strings (see https://dev.mysql.com/doc/refman/8.0/en/charset-binary-collations.html#charset-binary-collations-trailing-space-comparisons). We could use conversion to BINARY type here, but doing so would make the comparison case-sensitive.

smadbe (approved):

PR #596, opened 2020-09-02
Do not check member conditions if the team itself can enter in itemGetEntryState & itemEnter

### LINKED ISSUE #502: Item entry: do not check member conditions if the team has "can_enter"
## Motivations
In some contexts (Alkindi round 3), some teams may be qualified for the next round. In this case, we do not want to put conditions on the members of the team, but just allow the team.

## Task
In "enter" and "getQualificationState" services, if it is "team" item and if the team has "can_enter_*" condition valid from entering this item (so checking entering_time_min/max as well), do not check conditions on team members (contest_max_team_size and contest_entering_condition).

zenovich:
Fixes #502

### DIFF
--- a/app/api/items/get_entry_state.feature
+++ b/app/api/items/get_entry_state.feature
@@ -130,6 +130,57 @@ Feature: Get entry state (itemGetEntryState)
       | Half                             | not_ready      |
       | One                              | not_ready      |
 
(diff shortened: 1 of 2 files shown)


smadbe (approved):
👍

PR #597, opened 2020-09-02
update lambda archive to have a fully working version, at least for dev

smadbe:
Update the makefile "lambda-archive" command so that it fully work for deploying the current project. 
It is not fully ideal but it does the job.


zenovich (approved):
👍

PR #599, opened 2020-09-04
Get rid of items.is_root

### LINKED ISSUE #598: Replace is_root by root activities
## Motivations
"items.is_root" was the initial way of defining that an item has to be display in the user's navigation menu if the user has access to it.
Then "groups.root_activity_id" was introduced and plays a similar role, the user's navigation menu display all the root activities for all the group he belongs to and if it has access to the activity.

Having to ways to do the same things make things more complex in the backend and in the frontend. Let's get rid of "is_root".

## Analysis

In the anonymized db, there are 41 "is_root" items. For each of these items, only a few groups have been given access to it (via `permissions_granted`). In total there are 13 groups which have been given access to one of these items (mainly the `allUsersGroup` has, for obvious reasons), only 11 over these 13 groups are users.

So we will just put all these root items (items with is_root=t) as children of the root activity of the "all users" group. Only groups/users with enough permission will see them listed anyway.

## Subtasks

* create a chapter item with title "Public activities" in english, "Activités publiques" in French. 
* make this chapter visible to "all users" group (origin='group_membership',   can_view="content"
* set this chapter the root activity of "all users" group
* add the 41 "is_root" items as children of this chapter
* Get rid of "items.is_root" properties. Adapt all services.

zenovich:
Fixes #598

### DIFF
--- a/app/api/items/create_attempt.feature
+++ b/app/api/items/create_attempt.feature
@@ -1,24 +1,28 @@
 Feature: Create an attempt for an item
   Background:
     Given the database has the following table 'groups':
-      | id  | type | root_activity_id |
-      | 101 | User | null             |
-      | 102 | Team | 60               |
-      | 111 | User | null             |
+      | id  | type  | root_activity_id |
+      | 99  | Other | 50               |
+      | 100 | Class | 10               |
+      | 101 | User  | null             |
+      | 102 | Team  | 60               |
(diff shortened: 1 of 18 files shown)


smadbe (approved):
Seems correct but apparently the migration test failed

zenovich:
> Seems correct but apparently the migration test failed

Fixed

PR #604, opened 2020-10-13
Implement the resultStartPath service

### LINKED ISSUE #601: New service for starting results on a path
## Overview 

As discussed on 11/09/20 on Slack (Dmitry and DLe), we would need a service to find the result (so the attempt_id) to use for a given item path. If there is no such result, results should be created if possible. 

This service can be used in the context of link sharing (link given to another user) so with a missing or wrong attempt.  

(note that the UI probably would call this service with as path, the whole path but the last item, which do not necessarily have to be started when just displayed)

On `POST /items/{ids}/start-result-path`

## Input
* a path of items
* as_team_id

## Validations/Restrictions

* The usual as_team_id checks
* Valid path of visible items, starting from a root activity/skill

## Actions

Find the closest (starting from the end of the path) item which have started result and, if this result is allowing submission and is not ended, try to start result for the rest of the path (only if items do not requires_explicit_entry) until the last item of the path. 
If one step of the path cannot be started, do not start anything.

## Output
Generic output. On success, provide the attempt_id to use for the last item of the path.

zenovich:
Fixes #601

### DIFF
--- a/app/api/items/start_result_path.go
+++ b/app/api/items/start_result_path.go
@@ -1,0 +1,204 @@
+package items
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/go-chi/render"
+
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+// swagger:operation POST /items/{ids}/start-result-path items resultStartPath
+// ---
+// summary: Start results for an item path
+// description: >
+//   Creates new started results (or starts not started existing ones) for an item path if needed and returns the last attempt in the chain.
+//
+//   Of all possible chains of attempts the service chooses the one having missing/not-started results located closer
+//   to the end of the path, preferring chains having less missing/not-started results and having higher values of `attempt_id`.
+//   If there is no result for the first item, the service tries to create an attempt chain starting with the zero attempt.
+//   The chain of attempts cannot have missing results for items requiring explicit entry or require to start/create results
+//   within or below ended/not-allowing-submissions attempts.
+//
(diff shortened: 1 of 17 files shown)


app/api/items/start_result_path.go
```diff
@@ -0,0 +1,204 @@
...
+	subQuery := store.Table("visible_items as items0").WithWriteLock().Where("items0.id = ?", ids[0]).
+		Where("items0.id IN ? OR items0.id IN ?",
+			participantActivities.SubQuery(), participantSkills.SubQuery())
+
+	previousAttemptIDForCondition := "attempts0.id"
+	var score string
+	var columns string
+	attemptIsActiveCondition := "1"
+	var columnsForOrder string
+	for i := 0; i < len(ids); i++ {
```
smadbe, line 153: So you add more join for each depth? Does it really scale? I am surprised you do not limit (to a quite "low" value) the max number of elements in the path. I cannot see how it cannot cause a problem is the user comes up with a path of 1000 elements. 
Would it be safer (even if probably slower in total on average) to do that iteratively?
(maybe I didn't understood correctly what you did)

zenovich: Actually there is a limit:
```
func idsFromRequest(r *http.Request) ([]int64, error) {
	ids, err := service.ResolveURLQueryPathInt64SliceField(r, "ids")
	if err != nil {
		return nil, err
	}
	if len(ids) > 10 {
		return nil, errors.New("no more than 10 ids expected")
	}
	return ids, nil
}
```

zenovich: Added a test for the case

zenovich: (we never accept paths longer than 10 items anywhere)

smadbe: 👍

smadbe (approved):

PR #611, opened 2020-10-22
Implement the itemPathFromRootFind service

### LINKED ISSUE #602: New service to find the path to an item
## Overview 

In case the path to an item is not given or invalid (for instance, if a link is shared between users, and just provided), we try to find a path to the item which is valid for the user.

On `GET /items/:id/path_from_root`

## Input
* items
* as_team_id

## Output
* one path of items from one of the participant's root activities/skills to the input item. All items of the path must be visible (by the participant) and must not include item which requires_explicit_entry and not started, ended, or not allowing submission. 
* If there are multiple solution return one of them. 
* the path is an array of item_ids.
* 'forbidden' if there is no such path

## Validations
* usual as_team_id check

zenovich:
Fixes #602

### DIFF
--- a/app/api/items/path_from_root.feature
+++ b/app/api/items/path_from_root.feature
@@ -1,0 +1,109 @@
+Feature: Find an item path
+  Background:
+    Given the database has the following table 'groups':
+      | id  | type  | root_activity_id |
+      | 90  | Class | 10               |
+      | 91  | Other | 50               |
+      | 101 | User  | null             |
+      | 102 | Team  | 60               |
+      | 111 | User  | null             |
+    And the database has the following table 'users':
+      | login | group_id |
+      | john  | 101      |
+      | jane  | 111      |
+    And the database has the following table 'groups_groups':
+      | parent_group_id | child_group_id |
+      | 90              | 111            |
+      | 90              | 102            |
+      | 91              | 111            |
+      | 102             | 101            |
+    And the groups ancestors are computed
+    And the database has the following table 'items':
+      | id | url                                                                     | type    | allows_multiple_attempts | default_language_tag |
+      | 10 | null                                                                    | Chapter | 1                        | fr                   |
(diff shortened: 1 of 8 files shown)


app/api/items/path_from_root.feature
```diff
@@ -0,0 +1,109 @@
...
+      | 3  | 102            | 10           | 2019-05-30 11:00:00 | null                | 2019-05-30 11:00:00      |
+    And the database has the following table 'results':
+      | attempt_id | participant_id | item_id | started_at          | latest_activity_at  | result_propagation_state |
+      | 1          | 102            | 10      | 2019-05-30 11:00:00 | 2019-05-30 11:00:00 | done                     |
+      | 2          | 102            | 10      | 2019-05-30 11:00:00 | 2019-05-30 11:00:00 | done                     |
+      | 2          | 102            | 60      | 2019-05-30 11:00:00 | 2019-05-30 11:00:00 | done                     |
+      | 3          | 102            | 10      | 2019-05-30 11:00:00 | 2019-05-30 11:00:00 | done                     |
+      | 3          | 102            | 60      | 2019-05-30 11:00:00 | 2019-05-30 11:00:00 | done                     |
+
+  Scenario Outline: Find a path for the user's self group
```
smadbe, line 60: I don't get what you mean by "user's self group" here

zenovich: That means 'no as_team_id parameter'

smadbe: I don't get how "user's self group" mean "as a user", but ok ;-)

smadbe (approved):

app/api/items/path_from_root.feature
zenovich: fixed

smadbe (approved):
Sorry, I didn't see you re-asked for a review.

PR #631, opened 2020-11-25
Test undo only for new migrations 

smadbe:
- only try to undo the migrations which are new to this branch (not perfect but does the job)
- adapt new timeout to the new db dump

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,11 +117,11 @@ jobs:
       - run:
           name: Apply all migrations
           command: make db-migrate
-          no_output_timeout: 2h
+          no_output_timeout: 10m
       - run:
           name: Undo all migrations
           command: for m in `git diff --name-only HEAD..origin/master | grep "db/migrations/"`; do if ! make db-migrate-undo; then break; fi; done
-          no_output_timeout: 30m
+          no_output_timeout: 10m
       - run:
           name: Re-apply all migrations
           command: make db-migrate
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,9 @@ jobs:
           at: ./
       - restore_cache: *CACHEKEYMOD
       - run: sudo apt-get update
+      - run:
+          name: Skip if no new migrations added
+          command: git diff --name-only HEAD..origin/master | grep -q "db/migrations/" || circleci-agent step halt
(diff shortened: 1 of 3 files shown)


zenovich:
According on the previous experience, most migration bugs are found by undoing old migrations, so I'd prefer trying to undo all the migrations if at least one new migration is added.

smadbe:
It takes hours to run all undo migrations... and actually the only goal of testing the undo-migration is to check that in practice, if we deploy the new migration on prod, we can still just undo it... the goal is never to rollback 50 migrations.

smadbe:
The main cases where it would not really work (and why there is "not perfect but does the job") is when:
- a migration is added not at the last one of history or a previous migration is modified.. but that should not happen
- when the undo of 2 migrations are conflicting when merged together on master (and branch were not merged/rebased before merging)

zenovich:
> The main cases where it would not really work (and why there is "not perfect but does the job") is when:
> 
> * a migration is added not at the last one of history or a previous migration is modified.. but that should not happen
> * when the undo of 2 migrations are conflicting when merged together on master (and branch were not merged/rebased before merging)

The undo may be applied successfully without even doing anything. The only way to check it really changes the DB back is to undo all the previous migrations and run all the migrations again. There were cases where we got an error somewhere in very old migrations because the new one was not undone properly.

.circleci/config.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           no_output_timeout: 10m
       - run:
           name: Undo all migrations
-          command: for m in `git diff --name-only HEAD..origin/master | grep "db/migrations/"`; do if ! make db-migrate-undo; then break; fi; done
+          command: for m in db/migrations/*.sql; do if ! make db-migrate-undo; then exit 1; fi; done
```
smadbe, line 123: Fix one thing at a time.  Here you both fix a real bug, and revert my change. Useless to work in this PR is the commit revert the change.

conf/config.example2.yaml
```diff
@@ -0,0 +1,8 @@
+database:
+  dbname: algorea_example
+
+domains:
+  -
+    domains: [127.0.0.1]
+    allUsersGroup: 3
+    tempUsersGroup: 2
```
smadbe, line 8: :-/

zenovich: fixed

zenovich (approved):

PR #634, opened 2020-11-25
Correctly calculate number of migrations we want to undo

zenovich:


smadbe (approved):
I cannot say if it works, I just hope you tested it :-).
In the worth case, that is not a big deal, we'll fix it :-)

PR #647, opened 2020-12-27
Support watched_group_id in activitiesView & skillsView

### LINKED ISSUE #624: activitiesView/skillsView add `watched_group_id`
## Motivations
In "observation-mode", the manager of a group should see the website "as the group". So it should see the nav meny "as the group" 

## Subtasks
* add the optional `watched_group_id` arg
* if given, return the root skills/activities of all ancestor groups of the watched group which are also ancestors or descendants of at least one group that the current user manages.

So if the manager of E and G executes this request with watched_group_id=K, the service should return the root act/skill (if any) of A, E, G, K. 
```
    A
   /  \
  F    E  G  H
        \ | /
          K
          |
          Z
```

(wait for Mathias confirmation on 23Nov2020 15:40)

zenovich:
Fixes #624


smadbe (approved):
Looks good 👍

PR #653, opened 2021-01-12
Implement the permissionsView service

### LINKED ISSUE #643: Permission fetching service
## Overview 
In order to modify the permission in the UI. We need to fetch the existing permissions.

On `GET /groups/{source_group_id}/permissions/{group_id}/{item_id}`

## Input
source_group_id, group_id and item_id

## Output
For each permission (can_view, can_grant_view, ... but can_enter, see below*):
- the granted permission (its minimal value if no entry) on this item, for this group (only) as the given source_group_id and  "group_membership" origin.
- the effective computed permission (its minimal value if no entry) on this item for this group (including ancestors)
- the granted permission (the max of entries, with the minimal value if no entry) on this item, of "group_membership" origin, any source_group_id, on this group and its ancestors, **but excluding** the entry matching the given group + source_group_id 
- the granted permission (the max of entries, with the minimal value if no entry) on this item, of "item_unlocking" origin, any source_group_id, on this group and its ancestors
- same with "self" origin
- same with "other" origin

(\*) As can_enter_\* has not computed permission and works with ranges, return:
- the granted permission (can_enter_from and can_enter_until) on this item, for this group (only) as the given source_group_id and  "group_membership" origin.
- for the effective computed permission, granted permission x4: only return the next time it can enter (i.e. now if we are now in a range, "MIN(can_enter_from among can_enter_from > now)" otherwise)

## Validations
* current user must be manager of the source_group_id with can_grant_group_access
* group must be a descendant of source_group_id
* current user must have can_grant_view > none on item_id

zenovich:
Fixes #643

### DIFF
--- a/app/api/groups/get_permissions.feature
+++ b/app/api/groups/get_permissions.feature
@@ -1,0 +1,861 @@
+Feature: Get permissions for a group
+  Background:
+    Given the database has the following table 'groups':
+      | id | name       | type  |
+      | 10 | Other      | Other |
+      | 21 | owner      | User  |
+      | 23 | user       | User  |
+      | 25 | some class | Class |
+      | 31 | jane       | User  |
+    And the database has the following table 'users':
+      | login | group_id | first_name  | last_name |
+      | owner | 21       | Jean-Michel | Blanquer  |
+      | user  | 23       | John        | Doe       |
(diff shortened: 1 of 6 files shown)


smadbe (approved):
👍  From what I checked, that seems correct. I couldn't really review the code itself.

PR #654, opened 2021-01-17
Add manager permissions in groupChildrenView

### LINKED ISSUE #637: Add management permissions to group children list
## Motivations
In order to be able to navigate through managed groups in the left menu (in the sub-groups which are implicit managed groups), include the levels of management permissions in the children service.

## Subtasks
Update `groupChildrenView` service:
- add can_manage, can_grant_group_access and can_watch_members (if no match, return 'none', false, false instead of null).

zenovich:
Fixes #637


smadbe:
You assigned it to me, didn't you want to set me as a reviewer instead?

smadbe (approved):

PR #655, opened 2021-01-18
Fix results propagation

### LINKED ISSUE #640: Investigate not propagated scores
As discussed on 27 nov on slack:

* Why as user 670968966872011405, the item 1005481689658317378 has a score 0 while it has a child with a 100 score?
* This score becomes 25 if I mark the result as ‘to_be_recomputed’ and run the ‘db-recompute’ command
* isn't it done by a migration already!?

zenovich:
Fixes #640


smadbe (approved):
I cannot 100% guarantee that it is correct 🤔🤷‍♂️

PR #656, opened 2021-01-22
Make results propagation a bit faster

### LINKED ISSUE #623: itemUpdate service very slow for updating children
On the anonymized prod db, just for updating 2 children of an items (which already exists, so no changes required), the service takes >7sec:
```
curl 'https://dev.algorea.org/api/items/754307144702229123' -X PUT -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: en-US,en;q=0.5'-H 'Authorization: Bearer ***' -data-raw '{"children":[{"item_id":"426313127344282113","order":0},{"item_id":"178573120557837616","order":1}]}'
```

zenovich:
Do not mark results as 'processing' since this state is not really needed.
Fixes #623

### DIFF
--- a/app/api/items/items.go
+++ b/app/api/items/items.go
@@ -4,7 +4,6 @@ package items
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/go-chi/chi"
 	"github.com/go-chi/render"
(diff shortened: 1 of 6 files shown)


zenovich:
This change doesn't completely fix #623, but it makes things a bit better (10 seconds for 1980584647557587953, 8 seconds for 754307144702229123 on my computer)

smadbe (approved):

zenovich:
Fixed for the situation where relations are not being changed. Real relations creation will still take some though

zenovich:
Fixes #623

zenovich:
> When a new child is added latest_activity_at, tasks_tried, tasks_with_help are not changed. validated_at is only propagated if the task was validated. score_computed are only propagated when it was not 0.
A new child may have results for some groups already...

smadbe:
Right, but it may not

smadbe:
I mean: the trigger may not be the more appropriate way to deal with adding new children. Most of the time, there is no impact, the average case should be very fast.

zenovich:
> I mean: the trigger may not be the more appropriate way to deal with adding new children. Most of the time, there is no impact, the average case should be very fast.

The latest version of updateItem doesn't change anything in the DB when there is no real changes, so the trigger doesn't mark results as 'to_be_propagated' in this case. I have no idea how we can stop results propagation using the current approach where we mark all the ancestor results as 'to_be_recomputed' in the beginning: each propagation step changes the value of the result_propagation_state, so there always are affected rows. But even if we rework the results propagation completely, there always can be a situation where we explicitly mark an ancestor result together with its descendant result as 'to_be_recomputed', so we cannot stop until we process all of those.

zenovich:
Currently both requests from #623 take about 100ms (when they don't change anything)

smadbe:
When you do not change anything to the children? Yes right, that's good. 
But still children are changed, for instance when a child is added, the impact should not be that big. Only results with scores should be impacted basically.

smadbe:
> > I mean: the trigger may not be the more appropriate way to deal with adding new children. Most of the time, there is no impact, the average case should be very fast.
> 
> I have no idea how we can stop results propagation using the current approach where we mark all the ancestor results as 'to_be_recomputed' in the beginning: each propagation step changes the value of the result_propagation_state, so there always are affected rows.

Yes maybe that's the problem.

> But even if we rework the results propagation completely, there always can be a situation where we explicitly mark an ancestor result together with its descendant result as 'to_be_recomputed', so we cannot stop until we process all of those.

In case there is an actual propagation to done, right. But probably in the change we currently do in practice, only a low percentage is actual changes.

smadbe (approved):

zenovich:
> When you do not change anything to the children? Yes right, that's good.
> But still children are changed, for instance when a child is added, the impact should not be that big. Only results with scores should be impacted basically.

I suppose it is not that simple. If there are no results for the parent item, they are created and computed if there are results for the new child item. If there are results for the parent item, but no results for the new child item, the parent item results should be recomputed since average score changes and validated_at can become NULL if the parent item's validation_type != 'One'.

smadbe:
> I suppose it is not that simple. If there are no results for the parent item, they are created and computed if there are results for the new child item. If there are results for the parent item, but no results for the new child item, the parent item results should be recomputed since average score changes and validated_at can become NULL if the parent item's validation_type != 'One'.

Yes probably that's not that simple but I am pretty sure some heuristics may have big impacts, typically the change are on a few 2-3 children, but the the number of impacted `results` by a single item is quickly of thousand records and those impacted (by propagation) are probably quickly x100 this amount. Of course not changing non-impacted `items` is a first step, then not changing the non-impacted `results` may give great results as well.  

But no need to spend time on that now if we do not have a clear idea of what we should do. But it will probably have to be rethought.

PR #660, opened 2021-02-08
Add support for cookie-stored tokens

### LINKED ISSUE #576: Support for cookie-stored token
## Motivations

As storing access token in browser local or session storage is considered harmful in terms of security (as any JS code on the domain can access them, leaving some doors open for XSS), we can either:
- store the token in the running app only (subject to steal, auth lost each tab the app is refreshed, closed, ...)
- rely on httonly (non-available to JS) cookies for keeping the token (but cannot be both on different domain (`SameSite=none`) and on http (not `secure`), which makes local dev difficult)

Some ref [here](https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html#Local_Storage), [here](https://dev.to/rdegges/please-stop-using-local-storage-1i04), [here](https://medium.com/@benjamin.botto/secure-access-token-storage-with-single-page-applications-part-1-9536b0021321)

## Subtasks
* Add a `?use_cookies` (if you prefer to just infer it from other parameter, that's ok as well) in tempUserCreate and accessTokenCreate
  - if given, set the token in a cookie instead of in "data"
  - keep providing `expires_in` in the json response
  - if use_cookies is given, support the input parameters: 
    * `cookie_secure`. If given, add "Secure" attribute to cookie)
    * `cookie_same_site`. If given, add SameSite=Strict attribute to the cookie, otherwise SameSite=None
    * as a cookie cannot be both non-secure and none-same-site, fail if none of the arg if given (with an understandable error message)
  - store the `secure` and `same_site` cookie attribute in the db together with the token
  - infer cookie `domain`and `path` from backend config
  - set the Expires and Max-Age attributes to the cookie
  - always set the `HttpOnly` cookie attribute
* In authLogout, if this token was given as cookie, clear the cookie. 
* In the auth middleware, check for the cookie. If given, use it as a token. If both cookie and Authorization header is given, use the Authorization header.

## Implementation details
There are two ways of providing session token: via the 'Authorization' header and via the 'access_token' cookie.

The middleware we use for the services requiring authentication, prefers the 'Authorization' header when it is given. Also, when both the cookie and the header are given, the cookie gets deleted by the middleware using correct cookie attributes. When only 'access_token' cookie is given, it doesn't become deleted.

The accessTokenCreate service, when called with the 'code' parameter, ignores the 'access_token' cookie, but fails if the 'Authorization' header is given. If the service responds with a new token and an old cookie exists with different attributes (or the new token is returned in JSON), the old cookie gets deleted (otherwise, overwritten).

The tempUserCreate service acts similar way. It ignores the 'access_token' cookie, but fails if the 'Authorization' header is given. If the service responds with a new token and an old cookie exists with different attributes (or the new token is returned in JSON), the old cookie gets deleted (otherwise, overwritten).

The authLogout service works through the authentication middleware (see above) handling the 'Authorization' header and the 'access_token' cookie as any other service requiring the user to be authenticated. Then the service deletes the session cookie (if it is given).

In order to be able to delete session cookies explicitly with correct attributes, we store variable attributes' values inside the session cookie itself, together with the access token. The session cookie is encoded as 'security_kind!token!cookie_domain!cookie_path'. The 'security_kind' is one of 1 (SameSite=Strict), 2 (Secure, SameSite=None), 3 (Secure, SameSite=Strict).

zenovich:
Fixes #576


app/api/auth/create_access_token.go
```diff
@@ -58,6 +64,24 @@ import (
...
+//   default: 0
+// - name: cookie_secure
+//   in: query
+//   description: If 1, set the cookie with the `Secure` attribute
+//   type: integer
+//   enum: [0,1]
+//   default: 0
+// - name: cookie_same_site
+//   in: query
+//   description: If 1, set the cookie with the `SameSite`='Strict' attribute value (instead of `SameSite`='None')
```
smadbe, line 81: it is more "Otherwise set `SameSite`='None'". "instead of" seems to suppose that's the default, while "Lax" is the default.

app/api/auth/create_access_token.go
```diff
@@ -91,16 +124,24 @@ func (srv *Service) createAccessToken(w http.ResponseWriter, r *http.Request) se
...
-	// "Authorization" header is given, requesting a new token from the given token
-	if len(r.Header["Authorization"]) != 0 {
+	_, cookieErr := r.Cookie("access_token")
+
+	// "Authorization" header / "access_token" cookie is given, requesting a new token from the given token
+	if len(r.Header["Authorization"]) != 0 || cookieErr == nil {
 		if _, ok := requestData["code"]; ok {
 			return service.ErrInvalidRequest(
-				errors.New("only one of the 'code' parameter and the 'Authorization' header can be given"))
+				errors.New("only one of the 'code' parameter and the 'Authorization' header (or 'access_token' cookie) can be given"))
```
smadbe, line 133: Must be mentioned in the doc as well (probably with "use_cookie")

app/api/auth/create_access_token.go
```diff
@@ -73,6 +97,15 @@ import (
...
+//         description: If true, set a cookie instead of returning the OAuth2 code in the data
+//       cookie_secure:
+//         type: boolean
+//         description: If true, set the cookie with the `Secure` attribute
+//       cookie_same_site:
+//         type: boolean
+//         description: If true, set the cookie with the `SameSite`='Strict' attribute value (instead of `SameSite`='None')
 // responses:
 //   "201":
 //     description: "Created. Success response with the new access token"
```
smadbe, line 111: I supposed the struc of the resp must be updated as well to allow `access_token` not to be given.

app/api/auth/logout.feature
```diff
@@ -6,16 +6,55 @@ Feature: Sign the current user out
       | 3        | jane  |
     And the DB time now is "2019-07-16 22:02:28"
     And the database has the following table 'sessions':
+      | user_id | expires_at          | access_token              | use_cookie |
```
smadbe, line 9: Should we handle the case when a token is given through headers and cookies? 
In particular I can think of a version of the app which uses cookies, then the app use header token... but when login out, if the cookie is not removed, it will be served for next requests (instead of a 401), which will cause strange behaviors quite hard to debug.

smadbe (changes requested):
👍  and nice PR decomposition in commits which makes it easy-to-read.

db/migrations/2102051631_add_cookie_attributes_into_sessions.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+ALTER TABLE `sessions`
+    ADD COLUMN `use_cookie` TINYINT(1) NOT NULL DEFAULT 0 AFTER `issuer`,
+    ADD COLUMN `cookie_secure` TINYINT(1) NOT NULL DEFAULT 0 AFTER `use_cookie`,
+    ADD COLUMN `cookie_same_site` TINYINT(1) NOT NULL DEFAULT 0 AFTER `cookie_secure`,
+    ADD COLUMN `cookie_domain` VARCHAR(255) DEFAULT NULL AFTER `cookie_same_site`,
+    ADD COLUMN `cookie_path` VARCHAR(255) DEFAULT NULL AFTER `cookie_domain`;
```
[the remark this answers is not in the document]
zenovich: I think it is a kind of protection for the case where we change the app's config moving the api to some subdomain/subpath. In this case the cookie is still being sent, but it is impossible to delete it using the domain & path from the config.

app/api/auth/logout.feature
zenovich: What will happen to the header in that case? If the header remains, the cookie is not used (even when the header contains a wrong token)

app/api/auth/create_access_token.go
zenovich: fixed

app/api/auth/create_access_token.go
zenovich: fixed

app/api/auth/create_access_token.go
zenovich: fixed

db/migrations/2102051631_add_cookie_attributes_into_sessions.sql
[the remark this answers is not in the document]
smadbe: mmh right

app/api/auth/logout.feature
smadbe: When there is no token, no authorization header is sent by the frontend. In practice, it always tries to get a new one but nothing really guarantees that. Maybe that's not so harmful, indeed (especially if we use quite short validity for tokens)...
Do you have scenarios testing that? That we set a cookie, then use a valid token in headers (must use the token), then an invalid token in headers (must use the token).

app/api/auth/create_temp_user.go
```diff
@@ -19,7 +19,7 @@ import (
 // summary: Create a temporary user
 // description: Creates a temporary user and generates an access token valid for 2 hours
 //
-//   * No “Authorization” header should be present
+//   * None of “Authorization” header and "access_token" cookie should be present
```
smadbe, line 22: This breaks the scenario I discuss in the previous comment probably.
So: 
- the app use cookies,
- the app switch to tokens
- when starting, as it does not have token, it creates a temp user
- this service fails 

May be pretty hard to debug.

smadbe (changes requested):

app/api/auth/logout.feature
smadbe: As I wrote in the comment below, actually that only work if the create-temp-user and create-token-from-code work even if a cookie is provided (and ignored) tough.

app/api/auth/create_temp_user.go
zenovich: What will the app send in the Authorization header when there is no active session and the cookie is disabled? The middleware prefers the Authorization header over the cookie if both are present. So the app should send an empty bearer if it wants the backend to ignore the cookie.

zenovich: Should the backend prefer the Authorization header over the cookie in tempUserCreate & accessTokenCreate/accessTokenRefresh even when use_cookie=0?

smadbe: > What will the app send in the Authorization header when there is no active session and the cookie is disabled (especially in the requests where we do not have the use_cookie parameter)? The middleware prefers the Authorization header over the cookie if both are present. So the app should send a non-empty (but wrong, probably, short to be sure) bearer if it wants the backend to ignore the cookie.

Sending a non-empty wrong bearer when we do not have token ? Wow, very hacky. When the app has no token for a service which returns a token... it just does not send token, which makes sense to me. 

> Should the backend prefer the Authorization header over the cookie in tempUserCreate & accessTokenCreate/accessTokenRefresh even when use_cookie=1?

For this specific service, you could just ignore any authentication instead of failing when there is one given. That looks simple, is there something wrong with that?
For accessTokenCreate, as it requires input (either code/state or token), it is just a matter of priorities I think.

zenovich: So,
1) tempUserCreate & accessTokenCreate/accessTokenRefresh always ignore both the Authorization header and the cookie (no matter what use_cookie's value is) and set the new cookie (if use_cookie=1) without deleting the old one (this may lead to the situation where several different session cookies are set with different attributes).
2) All the services requiring the user to be authenticated prefer the Authorization header over the cookie, so, after logging out (when the UI doesn't use cookies), it is possible that the cookie still exists in the user's browser keeping the use authenticated.
3) authLogout deletes only one session at once (preferring the one from the header), so another session may still remain keeping the user logged in.

Is that really what we want?

smadbe: >     1. tempUserCreate & accessTokenCreate/accessTokenRefresh always ignore both the Authorization header and the cookie (no matter what use_cookie's value is) and set the new cookie (if use_cookie=1) without deleting the old one 

accessTokenCreate/accessTokenRefresh cannot ignore the given token as it may use it to get a new one.  So it can just use in this preference order: code/state, authorization header, cookie

> (this may lead to the situation where several different session cookies are set with different attributes).

So if a browser has two cookies with the same name and different parameters (that it does not send), it may send two cookies with the same name in its "Cookie" header to the backend ? Looks strange.

>     2. All the services requiring the user to be authenticated prefer the Authorization header over the cookie, so, after logging out (when the UI doesn't use cookies), it is possible that the cookie still exists in the user's browser keeping the use authenticated.

The backend would return authenticated-like response while the frontend thinks it is not connected.

>     3. authLogout deletes only one session at once (preferring the one from the header), so another session may still remain keeping the user logged in.

I don't see the difference with the previous one.

> Is that really what we want?

What are the other options?

1. The frontend always provides in header the type of authentication it uses. 

2. When a cookie is provided in addition to an authorization header or a code/state (for tokenCreate), the backend always return a Set-Cookie removing it. But it only works if it is a valid cookie (so it has parameters), otherwise it is unable to do so anyway. 

As a consequence of (2), the logout would logout both token if both an header and cookie would be given.

Does (2) looks good to you ? Otherwise do you have another proposal. Anyway we should re-write the full proposal as it is not straightforward.

zenovich: reworked, should be fine now

app/api/auth/create_access_token.go
```diff
@@ -21,26 +23,37 @@ import (
...
 //     saves it into the DB keeping only the input token (from authorization headers) and the new token.
 //     Since the login module responds with both access and refresh tokens, the service updates the user's
 //     refresh token in this case as well. If there is no refresh token for the user in the DB,
 //     the 'not found' error is returned.
 //
 //
-//   * One of the “Authorization” header and the `{code}` parameter should be present (not both at once).
+//   * The "Authorization" header is not allowed when the `{code}` is given.
+//
+//   * The "access_token" cookie is not allowed when the `{code}` is given and `{use_cookie}`=1.
```
smadbe, line 54: The client does not control the "access_token" cookie. If there is a dummy "access_token" cookie, the user cannot re-login with that rule and so is stuck forever?
Instead, the cookie could be ignored, and it would probably be overridden by the new one.

app/api/auth/create_temp_user.go
```diff
@@ -20,7 +19,30 @@ import (
 // summary: Create a temporary user
 // description: Creates a temporary user and generates an access token valid for 2 hours
 //
-//   * No “Authorization” header should be present
+//   * The "Authorization" header must not be given.
+//
+//   * The "access_token" cookie must not be given when `{use_cookie}`=1.
```
smadbe, line 24: same remark as above

smadbe (changes requested):
At the end, is the cookie invalidated (set to 0 validity) when invalid on every service as discussed? 
Anyway I need the expected behavior to be written (in comments of the issue for instance) so that I can validate it and for the records.

zenovich:
> At the end, is the cookie invalidated (set to 0 validity) when invalid on every service as discussed?

All the usual services requiring authentication only delete the session cookie if the Authorization header is given.
authLogout deletes the cookie only if the current session token is from the cookie.
refreshAccessToken() deletes the cookie before setting the new one only if the current session token is from the cookie.
Other services never delete the session cookie.

app/api/auth/create_access_token.go
zenovich: fixed

app/api/auth/create_temp_user.go
zenovich: fixed

smadbe (approved):
👍

PR #661, opened 2021-02-08
Implement the groupBreadcrumbsView service

### LINKED ISSUE #651: Get group breadcrumbs
## Overview 
Fetch a list of group info in order to display breadcrumbs. 

On `GET /groups/{ids)/breadcrumbs`

## Input
* slash-separated list of IDs

## Output
List of group info, in the same order as the id list, each with:
- group_id
- name
- type

## Validations
* No validation required on the hierarchy itself nor the root.
* Each group must be visible, so either 1) ancestors of group the current user joined, 2) ancestors of group he manages, 3) descendants of group he manages, 4) groups with is_public=1.

zenovich:
Fixes #651


app/api/groups/get_breadcrumbs.feature
```diff
@@ -0,0 +1,200 @@
...
+
+  Scenario: Get breadcrumbs
+    Given I am the user with id "41"
+    When I send a GET request to "/groups/9/11/7/16/4/19/27/22/13/14/breadcrumbs"
+    Then the response code should be 200
+    And the response body should be, in JSON:
+    """
+    [
+      {"id": "9", "name": "Managed Class", "type": "Class"},
+      {"id": "11", "name": "Ancestor Team", "type": "Team"},
```
smadbe, line 120: I don't know if that's important but this cannot happen as team cannot have a non-user as a child.

app/api/groups/get_breadcrumbs.feature
```diff
@@ -0,0 +1,200 @@
...
+      {"id": "7", "name": "Joined By Ancestor Team", "type": "Class"},
+      {"id": "16", "name": "Joined By Ancestor", "type": "Class"},
+      {"id": "4", "name": "Joined Class", "type": "Class"},
+      {"id": "19", "name": "Managed By Ancestor", "type": "Class"},
+      {"id": "27", "name": "Public", "type": "Base"},
+      {"id": "22", "name": "Root With Descendant Managed By Ancestor", "type": "Other"},
+      {"id": "13", "name": "Root With Managed Ancestor", "type": "Friends"},
```
smadbe, line 127: What's your definition of "root" if it is not a root?

app/api/groups/get_breadcrumbs.feature
```diff
@@ -0,0 +1,200 @@
...
+      | 10                | 41             | 2010-01-01 00:00:00 |
+      | 12                | 41             | 2010-01-01 00:00:00 |
+      | 13                | 41             | 2010-01-01 00:00:00 |
+      | 14                | 41             | 2010-01-01 00:00:00 |
+      | 15                | 41             | 2010-01-01 00:00:00 |
+      | 19                | 41             | 2010-01-01 00:00:00 |
+      | 22                | 41             | 2010-01-01 00:00:00 |
+      | 23                | 41             | 2010-01-01 00:00:00 |
+
+  Scenario: Get breadcrumbs
```
smadbe, line 112: Maybe you should indicate that this is a list of non-necessarily parent-child relations in case a reader wonder if they must or not.

smadbe (approved):

app/api/groups/get_breadcrumbs.feature
zenovich: This team has only one child - the user with id=41. So here we implicitly test that the service doesn't check the hierarchy.

app/api/groups/get_breadcrumbs.feature
zenovich: They all are expired

app/api/groups/get_breadcrumbs.feature
zenovich: fixed

Issue #662, opened 2021-02-10
Update item - bug in doc 


smadbe:
Update item doc (https://france-ioi.github.io/algorea-devdoc/api/#operation/itemUpdate) indicates most attributes have to be given under the "item" key, while in practice, that's not the case (as shown in the tests).

smadbe:
Fixed with #663

PR #663, opened 2021-02-11
Fix permissions checking in itemUpdate + fix swagger docs

zenovich:


smadbe:
Fixes #662 and a perm issue in update item service.

smadbe (approved):

PR #671, opened 2021-03-03
Implement groupPathFromRootFind service

### LINKED ISSUE #652: Find a group path service
## Overview 

On `GET /groups/:id/path-from-root`

## Input
* a group_id

## Output
A path as an array of group ids from a visible group root (a **visible** group without non-"base" parent) to the input group. Each group must be visible, so  either 1) ancestors of group he joined, 2) ancestors of group he manages, 3) descendants of group he manages, 4) groups with is_public=1.

Note that if the group_id is a public, none of its parents may be visible by the current user. In this case, this group is the root.

## Validations
* The given group id must be visible as well.

zenovich:
Fixes #652


smadbe (approved):
👍

PR #684, opened 2021-03-24
Display users' first name & last name only if there is an approval

### LINKED ISSUE #389: Display user's first name & last name only if there is an approval
There are still services (contestGetQualificationState, groupRequestsView, groupTeamDescendantView, groupUserDescendantView, itemAttemptsView and possibly invitationsView) where we display first_name and last_name without approvals checking. This should be fixed (we should display personal info only if the authorized user is a manager of a group for which the displayed user approved access to his personal info or when the authorized user views his own info).

zenovich:
Fixes #389


smadbe (approved):
👍

PR #685, opened 2021-03-25
Allow to keep orphans in groupRemoveChild

### LINKED ISSUE #682: groupRemoveChild: remove the orphan check
## Motivations
The workflow for removing subgroup has changed, adapt the service

## Subtasks
In [groupRemoveChild](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupRemoveChild) service:
- if the deletion would create an orphan and delete_orphans=0, do not fail with 422 but just delete it.

zenovich:
Fixes #682

### DIFF
--- a/app/api/groups/remove_child.feature
+++ b/app/api/groups/remove_child.feature
@@ -99,3 +99,32 @@ Feature: Remove a direct parent-child relation between two groups
       | 22                | 22             | 1       |
     And the table "groups" should stay unchanged but the row with id "13"
     And the table "groups" should not contain id "13"
+
+  Scenario: Delete a relation making a child group an orphan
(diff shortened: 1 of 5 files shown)


smadbe (approved):
There were no feature test on "refusing to delete orphans" before then 🤔

zenovich:
> There were no feature test on "refusing to delete orphans" before then 🤔

There was one, see https://github.com/France-ioi/AlgoreaBackend/pull/685/files#diff-27c7708bbea593c0b5da0f73e6739ec29acf10edf3536ad6d45594ff37ba3c0cL39

smadbe:
Ah sorry, I didn't see it.

PR #686, opened 2021-03-27
Implement groupDelete service

### LINKED ISSUE #683: groupDelete
## Overview 
Group deletion service

On `DELETE /groups/:id`

## Input
* the group id

## Validations/Restrictions
* The current user must be manager with `memberships_and_group` level (implicitly or explicitly)
* The group must be empty (no active subgroups of any type)
* The group must not be a user

## Actions
* delete the group (with all cascading as done in groupRemoveChild with delete_orphans)

## Output
Generic output

zenovich:
Fixes #683

### DIFF
--- a/app/api/groups/delete_group.go
+++ b/app/api/groups/delete_group.go
@@ -1,6 +1,7 @@
 package groups
 
 import (
+	"errors"
 	"net/http"
 
 	"github.com/go-chi/render"
@@ -24,8 +25,9 @@ import (
 //
 //   Restrictions (otherwise the 'forbidden' error is returned):
 //     * the authenticated user should be a manager with `can_manage` = 'memberships_and_group' on the `{group_id}`,
-//     * the group should not be of type "User",
-//     * the group must be empty (no active subgroups of any type).
+//     * the group should not be of type "User".
+//
+//   Also, the group must be empty (no active subgroups of any type), otherwise the 'not found' error is returned.
 // parameters:
 // - name: group_id
 //   in: path
@@ -41,6 +43,8 @@ import (
 //     "$ref": "#/responses/unauthorizedResponse"
 //   "403":
 //     "$ref": "#/responses/forbiddenResponse"
+//   "404":
+//     "$ref": "#/responses/notFoundResponse"
 //   "500":
 //     "$ref": "#/responses/internalErrorResponse"
 func (srv *Service) deleteGroup(w http.ResponseWriter, r *http.Request) service.APIError {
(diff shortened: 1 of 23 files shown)


app/api/groups/delete_group.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/service"
+)
+
+// swagger:operation DELETE /groups/{group_id} groups groupDelete
+// ---
+// summary: Delete a group
+// description: >
+//   Removes a group, its parent-child relations, and recursively deletes each
+//   new orphaned group.
```
smadbe, line 17: The spec was "The group must be empty (no active subgroups of any type)" so I don't get why you would delete any orphaned group. 🤔

zenovich: Fixed. I was confused by the phrase 'with all cascading as done in groupRemoveChild with delete_orphans'.

app/api/groups/delete_group.go
```diff
@@ -0,0 +1,84 @@
...
+//   The service deletes `groups_groups`, `attempts`, `results`,
+//   `group_membership_changes`, `group_pending_requests`,
+//   `permissions_granted`, `permissions_generated`, and `filters` linked to the group.
+//   Access rights are updated accordingly too.
+//
+//
+//   Restrictions (otherwise the 'forbidden' error is returned):
+//     * the authenticated user should be a manager with `can_manage` = 'memberships_and_group' on the `{group_id}`,
+//     * the group should not be of type "User",
+//     * the group must be empty (no active subgroups of any type).
```
smadbe, line 28: The UI needs to be able to distinguish between these two errors.  
(Probably the group not been empty is more a 400?)

zenovich: fixed (with 404)

smadbe (changes requested):
And if the id is not existing, what is the error? 404 as well I assume (Not documented?)
So we may show "the group you try to delete is not empty" error while actually another admin has deleted the group a few seconds ago. 😕

zenovich:
> And if the id is not existing, what is the error? 404 as well I assume (Not documented?)
> So we may show "the group you try to delete is not empty" error while actually another admin has deleted the group a few seconds ago. 😕

If the group_id doesn't exist, the service returns 403 (because it should be impossible to distinguish invisible groups from absent ones)

smadbe:
So we cannot distinguish between not-visible and not-authorized-to-delete. 🤷‍♂️

smadbe (approved):

PR #689, opened 2021-03-29
Allow setting propagation rules to the same or lower values in itemUpdate

### LINKED ISSUE #532: Update item: allow not changing or decreasing propagation rules
## Motivations
Fix a mistake in the update of the item perm propagation rules:
The doc states "To decrease the propagation level (whatever the level), you do not need any specific permissions on the child item.". It means that a user can submit the same or a lower propagation rule than the current one even if he would not be allowed to give this propagation rule in the first place. 

## Subtasks
* Fix `updateItem` service

zenovich:
Fixes #532


smadbe (approved):
👍

PR #692, opened 2021-03-29
Improve groupManagersView

### LINKED ISSUE #411: Improve `List group managers` service
## Changes

**Note:** Some of these may already be implemented but, if so, are not documented. Doc should include this info.

- If the manager is a user, add his first_name and last_name (no specific permission required)
- Allow manager of ancestor groups to use this service
- If "include_ancestor_groups" is given and set to 1:
  * include the managers of all ancestor groups (so who can manage this group as well)
  * add the inherited level of managership for each perm (can_manage, can_watch_members, can_grant_group_access), take the max among ancestors (including the group itself) (so keep `can_manage` but introduce `ancestors_can_manage` (or `implicit_can_manage` or `inherited_can_manage`,...))

zenovich:
Fixes #411

### DIFF
--- a/app/api/groups/get_managers.feature
+++ b/app/api/groups/get_managers.feature
@@ -4,19 +4,20 @@ Feature: Get managers of group_id
       | login | group_id | first_name  | last_name  | grade |
       | owner | 21       | Jean-Michel | Blanquer   | 3     |
     And the database has the following table 'groups':
-      | id | name        |
-      | 11 | user        |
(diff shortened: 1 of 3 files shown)


smadbe (approved):

smadbe (changes requested):
You should add in the doc that the group is part of its ancestor

smadbe (approved):

PR #710, opened 2021-04-22
groupGet: add more info about membership and managership

### LINKED ISSUE #702: get group: add more info about membership and managership
## Motivations
Be able to display, on group page, info banner such as the blue one in https://france-ioi.github.io/algorea-designs/01.Groups_00.Header_07.Member-of-parent (same for manager, ...)

## Subtasks

Proposal (feel free to suggest something better):
- current_user_is_member -> current_user_is_member and current_user_is_direct_member
- "True when there is an active group->user relation in groups_groups" in the doc -> I prefer we keep something clearer (not db related) for UI devs: "Whether the current user is an (active) direct member of this group"
- descendants_current_user_is_member_of: list of descendant (excluding self) groups (id, name) that the current user is member of 
- current_user_is_manager -> current_user_is_direct_manager and current_user_is_manager (explicit+implicit): whether the current user (or one of his ancestor) is a manager of this group / of this group or one its ancestors
- ancestors_current_user_is_manager_of: list of ancestor (excluding self) groups (id, name) that the current user (or his ancestor groups) is manager of 
- descendants_current_user_is_manager_of: list of descendants (excluding self) groups (id, name) that the current user (or his ancestor groups) is manager of

zenovich:
Fixes #702

### DIFF
--- a/app/api/groups/get_group.feature
+++ b/app/api/groups/get_group.feature
@@ -65,11 +65,14 @@ Feature: Get group by groupID (groupView)
       "code_lifetime": "01:00:00",
       "code_expires_at": "2017-10-14T05:39:48Z",
       "open_activity_when_joining": true,
-      "current_user_is_manager": true,
+      "current_user_managership": "direct",
       "current_user_can_manage": "none",
       "current_user_can_grant_group_access": false,
       "current_user_can_watch_members": false,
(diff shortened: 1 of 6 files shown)


smadbe (approved):
👍

PR #731, opened 2021-05-14
Speed up propagations

### LINKED ISSUE #715: Make the service updating item children faster
Investigate how we can make faster the service updating an item's children.
It is not even possible to reorder children among this (quite top level) chapter: 
```
curl 'https://dev.algorea.org/api/items/1980584647557587953' -X PUT -H 'Accept: application/json, text/plain, */*'-H 'Authorization: Bearer ******' -H 'Content-Type: application/json --data-raw '{"url":null,"uses_api":false,"children":[{"item_id":"39530140456452546","order":0},{"item_id":"1005481689658317378","order":1},{"item_id":"272785095226927012","order":2},{"item_id":"38417679686525447","order":3},{"item_id":"468199841181004922","order":4},{"item_id":"299402844547209142","order":5},{"item_id":"1053382493721676","order":6},{"item_id":"1794094566378556690","order":7},{"item_id":"1236671322844621710","order":8},{"item_id":"184321270892061553","order":9},{"item_id":"812762718770641925","order":10},{"item_id":"624843409166980007","order":11},{"item_id":"972755911692173036","order":12},{"item_id":"619766063193798036","order":13},{"item_id":"15961864736924119","order":14}]}'
```

zenovich:
Fixes #715

This PR make the request mentioned in #715 twice faster on my compute (7 seconds vs 15 seconds on my computer). Note that the request takes as long only for the first, on next runs it completes immediately since the items relations stay unchanged even if we change the ordering.

### DIFF
--- a/app/database/permission_granted_store_compute_all_access.go
+++ b/app/database/permission_granted_store_compute_all_access.go
@@ -99,9 +99,12 @@ func (s *PermissionGrantedStore) computeAllAccess() {
 	defer func() { mustNotBeError(stmtUpdatePermissionsGenerated.Close()) }()
 
 	// marking 'self' permissions_propagate (so all of them) as 'children'
+	// (although all existing rows in permissions_propagate have propagate_to='self' at this moment,
+	//  we still need to use WHERE clause in order for MySQL to use indexes,
+	//  otherwise the query can take minutes to execute)
 	const queryMarkSelfAsChildren = `
 		UPDATE permissions_propagate
-		SET propagate_to = 'children'`
+		SET propagate_to = 'children' WHERE propagate_to='self'`
 	stmtMarkSelfAsChildren, err = s.db.CommonDB().Prepare(queryMarkSelfAsChildren)
 	mustNotBeError(err)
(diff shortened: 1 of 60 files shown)


app/database/permission_granted_store_compute_all_access.go
```diff
@@ -34,14 +34,14 @@ func (s *PermissionGrantedStore) computeAllAccess() {
 		JOIN permissions_generated AS parents
 			ON parents.item_id = items_items.parent_item_id
 		JOIN permissions_propagate AS parents_propagate
-			ON parents_propagate.group_id = parents.group_id AND parents_propagate.item_id = parents.item_id AND
-				 parents_propagate.propagate_to = 'children'
```
smadbe, line 38: how do you find this kind of improvements?

smadbe (approved):
👍 👍

app/database/permission_granted_store_compute_all_access.go
zenovich: I usually go through different alternatives with EXPLAIN and EXPLAIN ANALYZE

PR #733, opened 2021-05-16
Unify hiding of user personal info in all the services (omit first_name and last_name instead of setting them to null)

### LINKED ISSUE #732: Unify hiding of user personal info in all the services (omit first_name and last_name instead of setting them to null)

zenovich:
Fixes #732


smadbe (approved):
I haven't checked everything but I suppose that's correct.

PR #742, opened 2021-05-25
Do not require watched groups to have view permission on items in activitiesView & skillsView

### LINKED ISSUE #738: Do not require the watched group to view the item to list it in activitiesView
## Motivations
In the activitiesView service with `watched_group_id`, most groups does not any activities returned as it requires both the current user and the watched group to have access to the item. The need for the watched group to have access is not that important as group members may have access by other means (for instance, "all users" group).

## Subtasks
Remove the requirement for the watched group to be allowed to view an item for it to be listed in the `activitiesView` service with `watched_group_id` parameter.

Same for skillsView

Same for itemNavigationView

zenovich:
Fixes #738

### DIFF
--- a/app/api/currentuser/get_group_activities.go
+++ b/app/api/currentuser/get_group_activities.go
@@ -261,7 +261,7 @@ func (srv *Service) getRootItemsFromDB(
 		itemsWithResultsQuery = itemsWithResultsQuery.Joins("JOIN items ON items.id = groups.root_skill_id")
 	}
 	itemsWithResultsSubquery := itemsWithResultsQuery.
-		JoinsPermissionsForGroupToItemsWherePermissionAtLeast(groupID, "view", "info").
+		JoinsPermissionsForGroupToItemsWherePermissionAtLeast(watcherID, "view", "info").
 		Joins("LEFT JOIN results ON results.participant_id = ? AND results.item_id = items.id", groupID).
 		Joins("LEFT JOIN attempts ON attempts.participant_id = results.participant_id AND attempts.id = results.attempt_id").
 		Select(`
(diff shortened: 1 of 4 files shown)


smadbe (changes requested):
I would expect a spec/doc change to explain the two changes (perm required, perm returned)

smadbe (approved):
actually I can approve, but please do the doc changes.

zenovich:
> I would expect a spec/doc change to explain the two changes (perm required, perm returned)

fixed

app/api/currentuser/get_group_activities.go
```diff
@@ -92,8 +92,10 @@ type rootItem struct {
 // description:
 //   If `{watched_group_id}` is not given, the service returns the list of root activities of the groups the current user
 //   (or `{as_team_id}`) belongs to.
-//   Otherwise, the service returns the list of root activities of all ancestor groups of the watched group which are also
+//   Otherwise, the service returns the list of root activities (visible to the current user or `{as_team_id}`)
+//   of all ancestor groups of the watched group which are also
 //   ancestors or descendants of at least one group that the current user manages.
```
smadbe, line 97: Reading  "of all ancestor groups of the watched group which are also ancestors or descendants of at least one group that the current user manages." -> but as the watched group is manage, all ancestors of the watched group are "ancestors or descendants of at least one group that the current user manages". Am I wrong?

zenovich: Not really.
```
A   B
  \ /
   C
```
If the current user is a manager of B, and the watched group is C, then the current user can manage C, but the root activity of A will not be displayed although it is an ancestor of the watched group.

smadbe: You should use "explicitly manage" and "implicitly manage". 
So your "at least one group that the current user manages" means "explicitly manage" while "the user should ba a manager of the group" two sentences below mean "implicitly manage"?

zenovich: Fixed in #744

PR #747, opened 2021-05-28
Implement the itemDependencyApply service

### LINKED ISSUE #695: item-dependencies: apply to existing
## Overview 
(re)Apply a specific existing item-dependencies rule on existing results

On `POST /items/{dependent_item_id}/prerequisites/{prerequisite_item_id}/grant_content_view`

## Input
* dependent_item_id, prerequisite_item_id

## Validations/Restrictions
- The dependent_item_id, prerequisite_item_id must exist, with grant_content_view=1 
- current-user must have:
  - can_edit=all on dependent_item_id
  - can_grant_view >= content on the dependent_item_id

## Actions
Apply the rule, i.e. grant the content access, for all existing participants which meet the condition defined by this dependency.
Do not delete entries for those who would not meet the condition anymore.

## Output
Generic output

zenovich:
Fixes #695

### DIFF
--- a/app/database/result_store_propagate.go
+++ b/app/database/result_store_propagate.go
@@ -300,8 +300,7 @@ func (s *ResultStore) Propagate() (err error) {
 				can_view = IF(VALUES(can_view) = 'content' AND can_view_value < ?, 'content', can_view),
 				can_enter_from = IF(
 					VALUES(can_enter_from) <> '9999-12-31 23:59:59' AND can_enter_from > VALUES(can_enter_from),
-					VALUES(can_enter_from), can_enter_from),
-				can_enter_until = IF(VALUES(can_enter_from) <> '9999-12-31 23:59:59', '9999-12-31 23:59:59', can_enter_until)`,
+					VALUES(can_enter_from), can_enter_from)`,
 			canViewContentIndex, canViewContentIndex)
 
 		mustNotBeError(result.Error)
--- a/app/database/result_store_propagate.go
+++ b/app/database/result_store_propagate.go
@@ -278,9 +278,9 @@ func (s *ResultStore) Propagate() (err error) {
 			INSERT INTO permissions_granted
 				(group_id, item_id, source_group_id, origin, can_view, can_enter_from, latest_update_at)
 				SELECT
-					groups.id AS group_id,
+					results.participant_id,
 					item_dependencies.dependent_item_id AS item_id,
-					groups.id,
+					results.participant_id,
 					'item_unlocking',
 					IF(items.requires_explicit_entry, 'none', 'content'),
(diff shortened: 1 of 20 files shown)


app/database/result_store_propagate.go
```diff
@@ -302,9 +301,8 @@ func (s *ResultStore) Propagate() (err error) {
 				can_enter_from = IF(
 					VALUES(can_enter_from) <> '9999-12-31 23:59:59' AND can_enter_from > VALUES(can_enter_from),
 					VALUES(can_enter_from), can_enter_from),
-				can_enter_until = IF(
-					VALUES(can_enter_from) <> '9999-12-31 23:59:59' AND can_enter_until <> '9999-12-31 23:59:59',
-					'9999-12-31 23:59:59', can_enter_until)`, canViewContentIndex, canViewContentIndex)
+				can_enter_until = IF(VALUES(can_enter_from) <> '9999-12-31 23:59:59', '9999-12-31 23:59:59', can_enter_until)`,
```
smadbe, line 304: But so, if the new value of `from`is +inf, you set `until` to its existing value? It looks like the case where you want to put `until` to `+inf` anyway. (`until` < `from` makes no sense).

The old rule meant: we set `until` to `+inf` EXCEPT if it previous value was not `+inf`and the new `from` is `+inf`... in which case we leave it to its previous value. (so `until` < `from` )

But as this `until` < `from`  does not make sense and "item_unlocking" rules are only set by code that we control, I think, we can just ignore `can_enter_until` and leave it to its db default value "9999-12-31 23:59:59".

smadbe (changes requested):

app/database/result_store_propagate.go
zenovich: In this query we never want to set `can_enter_until` to +inf explicitly.

Actually, the old rule meant we set until to +inf EXCEPT if it previous value was not +inf OR the new from is +inf... in which case we leave it to its previous value.

The only difference between the old rule and the new rule is that the old rule explicitly kept the old value of `can_enter_until` if it was +inf. The new rule overwrites +inf by +inf. So there is no difference between the rules.

smadbe: I just mean, that the new like the new rule does not really sense IMO. So we may not set `can_enter_until` at all.

zenovich: But if the can_enter_until is in the past, the item will not be unlocked (no matter how we set can_enter_from), so we should update can_enter_until here.

smadbe: Why would it be in the past for a "item_unlocking" row? But so, put it to "9999-12-31 23:59:59" without conditions.

zenovich: fixed

smadbe (approved):

PR #760, opened 2021-06-16
Implement currentAnswerGet service

### LINKED ISSUE #758: GET current answer service
## Overview 
When arriving on a task, by default, the "current answer" (latest auto-saved answer) should load. So we need a way to retrieve it (other than listing all answers and getting it by id)

On `GET /items/:item_id/current-answer` (?)

## Input
* item_id
* attempt_id
* as_team_id (optional)

## Output
same as https://france-ioi.github.io/algorea-devdoc/api/#operation/answerGet for the answer for the current participant, given item, given attempt and type=="Current"

## Validations
same as https://france-ioi.github.io/algorea-devdoc/api/#operation/answerGet

zenovich:
Fixes #758

### DIFF
--- a/app/api/answers/answers.go
+++ b/app/api/answers/answers.go
@@ -33,6 +33,7 @@ func (srv *Service) SetRoutes(router chi.Router) {
 	router.Post("/answers", service.AppHandler(srv.submit).ServeHTTP)
 
 	routerWithParticipant := router.With(service.ParticipantMiddleware(srv.Store))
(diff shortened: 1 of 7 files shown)


smadbe (changes requested):
My spec indicates a `attempt_id` parameter. I don't see it in your service.

zenovich:
> `attempt_id` parameter

Sorry. Now it is fixed.

smadbe (approved):

PR #763, opened 2021-06-21
Keep old values of items_items fields when they are not given in itemUpdate + make items_items.child_order required there

### LINKED ISSUE #762: Keep old values of items_items fields when they are not given in itemUpdate

zenovich:
Fixes #762


smadbe (approved):
thanks!

PR #770, opened 2021-07-16
Change type of groups.code_lifetime to int

### LINKED ISSUE #768: Various improvements to group code lifetime
## Overview
When testing the platform with actual user, we notice some aspects of the group code lifetimes which need to be improved.

## Acceptance criteria
- [ ] The current data structure for storing the code lifetime is a mysql "time" whose max value is '838:59:59' which is only <35days. It should be possible to set it to 1 year. Strictly negative value is not relevant, nor sub-second values. 
- [ ] It should be possible to express "infinite".
- [ ] "0" is valid, it means that the code is usable once (expires at its first usage)
- [ ] Infinite should be the default value
- [ ] In the legacy data, all "0" are actually 'infinite'.

zenovich:
Fixes #768


smadbe (approved):

smadbe:
Don't merge it now. Let's merge and deploy the other ongoing PRs first and I'll deploy this one a bit later so that the functionality is not broken for too long in the UI.

smadbe (changes requested):
I think you haven't handled to last request in the list: 

> In the legacy data, all "0" are actually 'infinite'.

Which means you need to put "NULL" everywhere the value is "0" in the existing data.

smadbe:
Ah I didn't see your

> There are no rows with code_lifetime = '00:00:00' in the anonymized DB

smadbe (approved):
Ah I didn't see you addressed this in the question. That's because I have 0 in many groups in the UI... but checking more precisely, that's because I was receiving "NULL" for many groups while having a code... which was kind of... invalid until now.

PR #775, opened 2021-07-20
Eliminate nil pointer dereference in itemChildrenView when show_invisible_items=1

### LINKED ISSUE #774: Bug: Get a 500 on itemChildrenView with show_invisible_items
As any http code `500` is unexpected, I suppose there is an issue here:
```
~ dle $ curl 'https://dev.algorea.org/api/items/561969155071897034/children?attempt_id=0&show_invisible_items=1' -H 'Authorization: Bearer ***'
{"success":false,"message":"Internal Server Error","error_text":"Runtime error: invalid memory address or nil pointer dereference"}
```

zenovich:
Fix #774


smadbe (approved):
Thanks!

PR #788, opened 2021-08-29
Really require the current user to have can_watch > "none" when watched_group_id is given

### LINKED ISSUE #782: Bug: Permission not properly verified on activity  log service
## Overview
The [activity log service](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemActivityLogForItem) seems not to implement the expected permissions: in observation mode (so if watched_group_id is set), the current user needs to have "can_watch > 'none'" on the item (in addition to the perm on the group).

The usual user can fetch activity of user 6381276653871190562 on 4702 while the current user has can_watch='none' on 4702.

```
~ dle $ curl 'https://dev.algorea.org/api/items/4702/log?limit=20&watched_group_id=6381276653871190562' -H 'Authorization: Bearer token123' 
[{"at":"2021-07-06T07:35:22Z","activity_type":"result_started","attempt_id":"0","from_answer_id":"-1","participant":{"id":"10071646858248478","name":"yann...]
```

## Acceptance criteria
- [ ] Implement correctly the permission: WHEN watched_group_id is set (and only in this case), the current user needs to have "can_watch > 'none'" on the item to run the request.

zenovich:
Fixes #782

### DIFF
--- a/app/api/items/get_activity_log.feature
+++ b/app/api/items/get_activity_log.feature
@@ -70,6 +70,7 @@ Feature: Get activity log
       | 36 | 31        | 11             | 1          | 203     | Submission | Current | 2017-05-29 06:38:38 |
       | 37 | 31        | 11             | 0          | 203     | Submission | Current | 2017-05-29 06:38:38 |
       | 38 | 31        | 11             | 1          | 203     | Submission | Current | 2017-05-30 06:38:38 |
+      | 39 | 31        | 11             | 1          | 204     | Submission | Current | 2017-05-30 06:38:38 |
     And the database has the following table 'gradings':
       | answer_id | graded_at           | score |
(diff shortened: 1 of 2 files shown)


smadbe (commented):
I would have expected a new test on `itemActivityLogForItem` on a non-authorized item  and which returns a 403.

zenovich:
> I would have expected a new test on `itemActivityLogForItem` on a non-authorized item and which returns a 403.

The service doesn't return 403 when can_watch='none'. Instead it hides the item from the list similarly to invisible items.

smadbe (approved):
Not sure it will be very clear for the frontend dev, but anyway if they do the pre-checks correctly they should not have the issue.

PR #792, opened 2021-09-10
Make itemActivityLogForItem work much faster for cases when the result is small or empty

### LINKED ISSUE #766: Slow "get log for group"
Investigate why this request is so slow and if there is no solution to make it run faster (do not implement them without validation if not trivial):
```
time curl 'https://dev.algorea.org/api/items/log?limit=20&watched_group_id=672913018859223173' -H 'Authorization: Bearer ***'
[]

real	0m5.040s
```

zenovich:
Fixes #766

### DIFF
--- a/app/api/items/get_activity_log.feature
+++ b/app/api/items/get_activity_log.feature
@@ -120,13 +120,14 @@ Feature: Get activity log
       | tag |
       | fr  |
 
-  Scenario: User is a manager of the group and there are visible descendants of the item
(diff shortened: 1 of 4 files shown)


zenovich:
> I don't really get how `forceStraightJoinInItemActivityLog` is set out of the tests.

It is here https://github.com/France-ioi/AlgoreaBackend/pull/792/files#diff-a0c52a4111813fd1008955c6ae5e626a42f53d46fe4512e497c412a9ed147b71R173

PR #804, opened 2021-11-14
Rework pagination

zenovich:
Require only values of tie-breaker fields as from.* parameters,
Remove unnecessary features and unused configuration fields
Rework configuration structures


smadbe (approved):
👍

PR #805, opened 2021-11-14
Implement grantedPermissionsView service

### LINKED ISSUE #783: List granted permissions
## Overview 
In order to have an overview of the permissions given to group/user, list all granted permissions.

On `GET /groups/{group_id}/granted_permissions`

Sort by item title ? (beware at the items without title)

## Input
* group_id
* descendants: boolean (default 0)

## Output
[if descendants = 0] For every granted permissions given by (= with as a source group) the group or one of its ancestors on which the current-user has can_grant_group_access, to the group or its ancestors, ...

[if descendants = 1] For every granted permissions given by (= with as a source group) the group, its ancestors+descendants`*` on which the current-user has can_grant_group_access, to its descendants (not including the group itself), ...

... , AND only if the current user can grant permissions to the item ('can_grant_view' >= 'enter' or can*='..with_grant'), return:

* group: id, name 
* source_group_id: id, name
* item: id, title, language tag, require_explicit_entry
* permissions

## Validations
* The current user (or one its ancestors) must be allowed to grant permissions (so at least 'can_grant_group_access'=true) to the given group or one of its ancestors.
* (`*`) Reminder: the managing permission propagates from group to its descendants, but not to participants, so a school see permissions with, as source group, a class of this school, but not with, as source group, a student.

zenovich:
Fixes #783

### DIFF
--- a/app/api/groups/get_granted_permissions.feature
+++ b/app/api/groups/get_granted_permissions.feature
@@ -1,0 +1,392 @@
+Feature: Get permissions granted to group
+  Background:
+    Given the database has the following table 'groups':
+      | id | name        | type  |
+      | 8  | Club        | Club  |
+      | 9  | Class       | Class |
+      | 10 | Other       | Other |
+      | 21 | owner       | User  |
+      | 23 | user        | User  |
+      | 25 | some class  | Class |
+      | 26 | other class | Class |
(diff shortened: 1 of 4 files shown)


smadbe (approved):
👍

PR #809, opened 2021-11-29
Allow observers to use getAnswer service

### LINKED ISSUE #806: Get answer: let observers call the service
## Overview

Users observing a group and having the appropriate permissions should be able to call this service. The [current spec](https://france-ioi.github.io/algorea-devdoc/api/#operation/answerGet) indicates only participants should be able to call this service.

## Acceptance criteria
- [ ] If current-user has [`can_watch>=answer`](https://france-ioi.github.io/algorea-devdoc/design/access-rights/items/#can_watch) on the item and watch permission on the participant (so manager with `can_watch_members` of an ancestor of the participant), he should be able to get an answer
- [ ] the current case should still be supported (a user should still be able to see his own answer)

zenovich:
Fixes #806

### DIFF
--- a/app/api/answers/answers.go
+++ b/app/api/answers/answers.go
@@ -6,6 +6,7 @@ import (
 	"github.com/go-chi/render"
 
 	"github.com/France-ioi/AlgoreaBackend/app/auth"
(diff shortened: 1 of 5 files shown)


smadbe (approved):
👍

PR #833, opened 2022-02-18
Release process

smadbe:
Make change to the app so we can have proper releases. 

* What has to be done for a release in document in the README file.
* API docs is generated by version so that the developers can retrieve the doc of a specific version
* Create a specific github release which can be used by external deployers
* The `version` is injected into the app (better usage that the dummy one will have to be done by you as a separate task)

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,11 @@ import (
 	"github.com/France-ioi/AlgoreaBackend/cmd"
 )
 
+var (
+	version = "unknown"
+	_       = version
+)
+
 func main() {
 	cmd.Execute()
 }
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -1,9 +1,16 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/France-ioi/AlgoreaBackend/cmd"
 )
 
+var version = "unknown"
```
zenovich, line 9: It might be cleaner to write
```
var (
   version = "unknown"
  _ = version
)
```
here instead of adding a dummy code into the main function

zenovich (changes requested):

main.go
smadbe: fixed

zenovich (approved):

PR #846, opened 2022-08-08
Process badges in accessTokenCreate & userDataRefresh

### LINKED ISSUE #842: Process badges in profile
## Motivations

Badges are a way to grant permissions across the platform managed by France-IOI.  Other platforms may offer badges as an achievement to a user profile, or just give a code which can be traded to a badge. At the end, the user can have badges assigned to his profile. These badges have to be interpreted, each time the user profile is synced, as a group membership, which is an easy way in algorea to grant permission.

## Subtasks
- Implement what is described in the [badge description](https://france-ioi.github.io/algorea-devdoc/task_platforms/badges/#badge-processing-by-the-backend), this sync has to be done each time the profile is updated.
- require probably to add a new attribute "latest_profile_sync" on the user table.
- return this new attribute in the get-profile service so that the frontend can trigger a new sync if the last update is not recent enough

zenovich:
Fixes the first part of #842

### DIFF
--- a/app/api/answers/answers.go
+++ b/app/api/answers/answers.go
@@ -26,12 +26,12 @@ type answerData struct {
 // SetRoutes defines the routes for this package in a route answers
 func (srv *Service) SetRoutes(router chi.Router) {
 	router.Use(render.SetContentType(render.ContentTypeJSON))
-	router.Use(auth.UserMiddleware(srv.Store.Sessions()))
+	router.Use(auth.UserMiddleware(srv.Base))
 	router.Get("/items/{item_id}/answers", service.AppHandler(srv.listAnswers).ServeHTTP)
 	router.Get("/answers/{answer_id}", service.AppHandler(srv.getAnswer).ServeHTTP)
 	router.Post("/answers", service.AppHandler(srv.submit).ServeHTTP)
 
-	routerWithParticipant := router.With(service.ParticipantMiddleware(srv.Store))
+	routerWithParticipant := router.With(service.ParticipantMiddleware(srv.Base))
 	routerWithParticipant.Get("/items/{item_id}/current-answer", service.AppHandler(srv.getCurrentAnswer).ServeHTTP)
 	routerWithParticipant.Post("/items/{item_id}/attempts/{attempt_id}/answers", service.AppHandler(srv.answerCreate).ServeHTTP)
 	routerWithParticipant.Put("/items/{item_id}/attempts/{attempt_id}/answers/current", service.AppHandler(srv.updateCurrentAnswer).ServeHTTP)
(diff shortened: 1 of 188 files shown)


zenovich:
Note 1: UserJoinsGroupByCode is used for adding users into badge groups. If that's not what we want, it can be changed.

zenovich:
Note 2:
Consider the situation where there are 2 badges in the login module: B1 and B2. B1 is a parent of B2. Also, there are two users: U1 and U2. Both users are managers of both badges in the login module. AlgoreaBackend isn't aware of those badges and users for now. Then U1 logs in to AlgoreaBackend for the first time. AlgoreaBackend creates user U1, badges B1 and B2, and makes U1 a manager of B1 and B2 as the badges are returned as
```
"badges": [{"url":"B2", "manager":true, "badge_info":{"group_path":[{"url":"b1", "manager": true}]}}]
```
 Then U2 logs in to AlgoreaBackend for the first time. With the same badges info, as both badges are already known to AlgoreaBackend, it creates U2 and makes him a manager of B2, but it doesn't make U2 a manager of B1 (as B1 is already known).

A similar behaviour is true for non-managerial users (with "manager":false in "badges"). User U2 will not become a member of group B1 in the scenario above.

zenovich:
Note 3:
When we create groups for badges and their group paths, the resulting hierarchy depends on the order of badges on the list. For
```
"badges": [
  {"url":"B3", "badge_info":{"group_path":[{"url":"B1"},{"url":"B2"}]}},
  {"url":"B4", "badge_info":{"group_path":[{"url":"B0"},{"url":"B2"}]}}
]
```
the result is B1 -> B2 -> (B3, B4) while for
```
"badges": [
  {"url":"B4", "badge_info":{"group_path":[{"url":"B0"},{"url":"B2"}]}},
  {"url":"B3", "badge_info":{"group_path":[{"url":"B1"},{"url":"B2"}]}}
]
```
it is B0 -> B2 -> (B3, B4)

smadbe:
> Note 1: UserJoinsGroupByCode is used for adding users into badge groups. If that's not what we want, it can be changed.

`UserJoinsGroupByCode` is not related with badges.  This is used to join (through Algorea) a group by entering the `group.code` (managed by algorea). 







> Note 3: When we create groups for badges and their group paths, the resulting hierarchy depends on the order of badges on the list. For
> 
> ```
> "badges": [
>   {"url":"B3", "badge_info":{"group_path":[{"url":"B1"},{"url":"B2"}]}},
>   {"url":"B4", "badge_info":{"group_path":[{"url":"B0"},{"url":"B2"}]}}
> ]
> ```
> 
> the result is B1 -> B2 -> (B3, B4) while for
> 
> ```
> "badges": [
>   {"url":"B4", "badge_info":{"group_path":[{"url":"B0"},{"url":"B2"}]}},
>   {"url":"B3", "badge_info":{"group_path":[{"url":"B1"},{"url":"B2"}]}}
> ]
> ```
> 
> it is B0 -> B2 -> (B3, B4)

Fine!

smadbe (approved):
Except my comment above, that looks good, thanks

zenovich:
> Except my comment above, that looks good, thanks

fixed

smadbe (approved):

PR #849, opened 2022-09-09
Store latest profile sync time and return it in userData

### LINKED ISSUE #842: Process badges in profile

zenovich:
Fixes #842


smadbe (approved):
👍

PR #853, opened 2022-11-04
Fix spec of updatePermissions

### LINKED ISSUE #835: Fix updatePermissions: allow giving perm to a root activity/skill of a group
## Motivations
To avoid the chicken-egg problem, allow group to receive access to their root activities/skills

## Subtasks
In updatePermissions:
- allow an authorized user (given current rules) to give a group permissions to an item, even the group does not already have access to one of the parents of the item or the item itself, if the item is a root activity/skill of the group on one of its ancestors

zenovich:
Fixed #835 
Add a note about that it is allowed to set permissions on items invisible for the user if they are root items for the user

### DIFF
--- a/app/api/groups/update_permissions.go
+++ b/app/api/groups/update_permissions.go
@@ -75,7 +75,8 @@ type managerGeneratedPermissions struct {
 //   * The user giving the access must have `permissions_generated.can_grant_view` >= given `can_view`
 //     for the item.
 //
-//   * The group must already have access to one of the parents of the item or the item itself.
+//   * The group must already have access to one of the parents of the item or the item itself. If it does not,
+//     the item must be a root activity/skill for an ancestor of the group.
 // parameters:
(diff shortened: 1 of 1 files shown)


app/api/groups/update_permissions.go
```diff
@@ -73,7 +73,7 @@ type managerGeneratedPermissions struct {
 //     of `{source_group_id}` which should be an ancestor of the `{group_id}`.
 //
 //   * The user giving the access must have `permissions_generated.can_grant_view` >= given `can_view`
-//     for the item.
+//     for the item. If he does not, the item must be a root activity/skill for an ancestor group of the user.
```
smadbe, line 76: 🤨 Isn't the remark about the root related with the next line instead?

zenovich: Right. Fixed :(

smadbe (approved):

PR #863, opened 2023-01-06
Provide "login" in TaskToken

### LINKED ISSUE #857: Provide "login" in TaskToken
The task token has a attribute for login. It is [already implemented in the struct](https://github.com/France-ioi/AlgoreaBackend/blob/master/app/payloads/task_token.go#L28) but [it is currently not set](https://github.com/France-ioi/AlgoreaBackend/blob/master/app/api/items/generate_task_token.go#L169).

So just add the current user's `login` to the token.

GeoffreyHuck:
fixes #857


smadbe (approved):
Looks good 👍

PR #867, opened 2023-01-09
Merge item type Course into Task

### LINKED ISSUE #820: Merge `task` and `course`
## Motivations
Items of type "courses" behave as the "tasks". Differences they had in the past has now disappeared or are expressed through other attributes.

## Subtasks
* Set the `no_score` bit to `1` for all the existing courses
* Change all items of type `course` to `task`
* Remove the `course` type in the db schema
* Update every services so that the `course` type does not appear anymore

GeoffreyHuck:
fixes #820


app/api/items/search_for_items.feature
```diff
@@ -36,12 +36,12 @@ Feature: Search for items
...
       | 1       | fr           | (the) Our Chapter                       |
       | 2       | fr           | (the) Our Task ___                      |
-      | 3       | en           | (the) Our Course                        |
+      | 3       | en           | (the) Our Task                          |
       | 4       | fr           | (the) \|\|\|Our Skill \\\\\\%\\\\%\\ :) |
       | 6       | en           | Another Chapter                         |
       | 6       | fr           | Un autre chapitre                       |
       | 7       | en           | Another %%%Task                         |
-      | 8       | en           | Another %%%Course                       |
+      | 8       | en           | Another %%%Task 2                       |
```
smadbe, line 44: I suppose that was the kind of data which was useful only to test both the task and the course case... so testing twice the tasks do not make sense anymore.

app/database/result_store_propagate_validated_at_integration_test.go
```diff
@@ -195,7 +195,7 @@ func TestResultStore_Propagate_Categories_SetsValidatedAtToMaxOfValidatedAtsOfCh
 		UpdateColumn("category", "Validation").Error())
 
 	assert.NoError(t, itemStore.Where("id=1").Updates(map[string]interface{}{
-		"type": "Course",
+		"type": "Task",
```
smadbe, line 198: Have you checked it was not a Task already? Otherwise your loose a bit a part of the goal of the test.

smadbe (approved):
I think many data rows in test cases using courses were just there to test the "course" case... while the "task" case is already tested. So converting them to "task" is quite useless (and so making the test more complex (to read (mainly) and run) for nothing. 
But I don't know how these cases are easy to identify... so it is up to you to see if it is worth the effort.

smadbe:
Please wait before merging anyway, I've just found another change that might be needed, but I need Michel's validation on on that first.

smadbe:
Small update: for existing courses (the migration), set the `no_score` bit to `1`.

app/api/items/search_for_items.feature
```diff
@@ -1,3 +1,4 @@
+@wip
```
smadbe, line 1: !!

smadbe (changes requested):

app/api/items/search_for_items.feature
GeoffreyHuck: oops !

app/database/result_store_propagate_validated_at_integration_test.go
GeoffreyHuck: I'm not sure what this is for ? Where does the item with id=1 come from ?

app/api/items/search_for_items.feature
GeoffreyHuck: This one was indeed only used for Courses. For the other ones it's hard to tell.

app/database/result_store_propagate_validated_at_integration_test.go
smadbe: For unit tests, data is loaded from fixture files (cfr `db := testhelpers.SetupDBWithFixture("results_propagation/_common", "results_propagation/validated_at")` on top of the test)... so loaded from what is in the `testdata` dir from the test directory.

smadbe (approved):

app/database/result_store_propagate_validated_at_integration_test.go
GeoffreyHuck: Ok. I'm not sure what to do with this line. The type is not defined within the fixtures files, but the test still works without this line.

smadbe: Ok I thought it was a test for the item get/update... but it is a test for propagation, so the courses behave like tasks, so it is better to keep as you did. 
Sorry, my initial remark was useless here.

PR #868, opened 2023-01-10
.editorconfig specified LF endlines but the whole project uses CRLF. Updated to CRLF.

GeoffreyHuck:
We could also keep LF in the config and update all the files


smadbe (approved):
This is the kind of thing which should be included in the lint checks.

PR #871, opened 2023-01-16
Fix ancestorsOfManagedGroups query that returned users

### LINKED ISSUE #858: Bug: root groups return users
The "List root groups" (`/groups/roots`) service spec: "Returns groups which are ancestors of a joined groups or managed non-user groups and do not have parents, not considering "type='Base'" groups". 

Here I have a case where a user gets several users in his results. (for the record: user 5353829006594239718 on opentezos)
Quickly checking the code, I do not understand how it is possible... but so there should be a case in which it happens.

So basically this task is: investigate what the problem might be and fix the problem if you find it (and create a tests reproducing it)

GeoffreyHuck:
fixes #858


smadbe (approved):
👍

PR #872, opened 2023-01-17
/groups/roots documentation update

GeoffreyHuck:


GeoffreyHuck:
Is the doc for ancestorsOfManagedGroups correct ?

smadbe (approved):

app/api/groups/get_roots.go
```diff
@@ -151,6 +151,7 @@ func ancestorsOfJoinedGroups(store *database.DataStore, user *database.User) *da
 		Select("groups_ancestors_active.ancestor_group_id")
 }
 
+// This function will also return entries for users who are in a group managed by the user.
```
smadbe, line 154: I think the Go doc convention is to start with the function name.
Ok, I am going to improve the explanation anyway: (I let you fix the line breaks)

```suggestion
// ancestorsOfManagedGroups returns all groups which are ancestors of managed groups, and all users who are descendants from managed groups.
```

I prefer to consider that users are not among the "explicitly and implicitly managed group". This is how it was defined initially with Dmitry. Maybe it should be explained in devdoc as well.

But so, maybe the function name should be changed to something like `managedUsersAndAncestorsOfManagedGroups`?

smadbe (changes requested):

PR #885, opened 2023-01-30
New service: get a task token for observation

### LINKED ISSUE #878: New service: get a task token for observation
## Motivations

The current `itemTaskTokenGenerate` service generates a task token to be provided to the task platform for solving a task. But another usage for a task token is observation: we need to generate a token for requesting a task to load the answer of another user. The current service does not handle the second case at all.

## Overview 

New "Generate a read-only task token from an answer" service, inspired by the current "Generate task token" service.

On `POST /answers/{answer_id}/generate-task-token`.

## Input
* `answer_id` - the `id` of an `answer` 

## Validations/Restrictions

- the answer should exist
- the item of the answer should be a "task"

- the current user must have a started result on the item (whatever the attempt)

- if the participant of the answer is either the current-user or a team which the current-user is member of
  - the current user must be allowed to "view >= 'content'" the item

- otherwise: 
  - the current user must be allowed to "watch" the participant of the answer
  - the current user must be allowed to "watch answer" for the item

## Actions

Unlike `itemTaskTokenGenerate`, do not update `latest_activity_at`.

Generate a task token with: 
- `AccessSolutions`: same as `itemTaskTokenGenerate`
- `SubmissionPossible`: false 
- `HintsAllowed`: false
- `HintsRequested`: those corresponding to the "result" of the answer
-  `HintsGivenCount`: corresponding to the "result" of the answer
- `IsAdmin`: false
- `ReadAnswers`: true
- `UserId`: the `author_id` of the answer
- `LocalItemID`  : same as `itemTaskTokenGenerate`
- `AttemptID`:  same as `itemTaskTokenGenerate`, using the attempt id corresponding to the answer
- `ItemURL` same as `itemTaskTokenGenerate`
- `SupportedLangProg`: same as `itemTaskTokenGenerate`
- `RandomSeed`:  same as `itemTaskTokenGenerate`, using the attempt id corresponding to the answer
- `PlatformName`:same as `itemTaskTokenGenerate`
- `Login`: the `login` of the `author_id` of the answer

## Output
Same as `itemTaskTokenGenerate`

GeoffreyHuck:
fixes #878

### DIFF
--- a/app/api/answers/generate_task_token.feature
+++ b/app/api/answers/generate_task_token.feature
@@ -1,0 +1,181 @@
+Feature: Generate a read-only task token for an item from an answer
+  Background:
+    Given the database has the following table 'groups':
+      | id  | name     | type  |
+      | 101 | john     | User  |
+      | 102 | team     | Team  |
+      | 103 | manager  | User  |
+      | 104 | Groupe A | Class |
+    And the database has the following table 'users':
+      | login | group_id |
+      | john    | 101    |
+      | manager | 103    |
+    And the database has the following table 'groups_groups':
+      | parent_group_id | child_group_id |
+      | 102             | 101            |
+      | 104             | 101            |
+    And the database has the following table 'group_managers':
+      | group_id | manager_id | can_watch_members |
+      | 104      | 103        | 1                 |
+    And the groups ancestors are computed
+    And the database has the following table 'items':
+      | id | url                                                                     | type | entry_participant_type | default_language_tag | text_id |
+      | 10 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | User                   | fr                   | task10  |
+      | 20 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | Team                   | fr                   | task20  |
+      | 30 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | User                   | fr                   | task30  |
+    And the database has the following table 'permissions_generated':
+      | group_id | item_id | can_view_generated | can_watch_generated |
+      | 101      | 10      | content            | none                |
+      | 102      | 20      | content            | none                |
+      | 103      | 10      | content            | answer              |
+      | 103      | 20      | content            | answer              |
+      | 103      | 30      | content            | answer              |
+      | 104      | 10      | content            | answer              |
+      | 104      | 20      | content            | answer              |
+      | 104      | 30      | content            | answer              |
+    And the database has the following table 'attempts':
+      | id | participant_id |
+      | 0  | 101            |
+      | 0  | 102            |
+      | 0  | 103            |
+      | 1  | 103            |
+    And the database has the following table 'results':
+      | attempt_id | participant_id | item_id | started_at          |
+      | 0          | 101            | 10      | 2020-01-01 01:01:01 |
+      | 0          | 101            | 20      | 2020-01-01 01:01:01 |
+      | 0          | 101            | 30      | 2020-01-01 01:01:01 |
+      | 0          | 102            | 20      | 2020-01-01 01:01:01 |
+      | 0          | 103            | 10      | 2020-01-01 01:01:01 |
+      | 0          | 103            | 20      | 2020-01-01 01:01:01 |
+      | 0          | 103            | 30      | null                |
+      | 1          | 103            | 30      | 2020-01-01 01:01:01 |
+    And the database has the following table 'answers':
+      | id | participant_id | attempt_id | item_id | author_id  | created_at          |
+      | 1  | 101            | 0          | 10      | 101        | 2020-01-01 01:01:01 |
+      | 2  | 102            | 0          | 20      | 101        | 2020-01-01 01:01:01 |
+      | 3  | 101            | 0          | 30      | 101        | 2020-01-01 01:01:01 |
+    And time is frozen
+
+  Scenario: User is able to fetch a task token when participant is the current user
+    Given I am the user with id "101"
+    When I send a POST request to "/answers/1/generate-task-token"
+    Then the response code should be 200
+    And the response body decoded as "GenerateTaskTokenResponse" should be, in JSON:
+      """
+      {
+        "data": {
+          "task_token": {
+            "date": "{{currentTimeInFormat("02-01-2006")}}",
+            "bAccessSolutions": false,
+            "bHintsAllowed": false,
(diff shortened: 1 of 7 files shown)


app/api/answers/generate_task_token.feature
```diff
@@ -0,0 +1,251 @@
...
+      | group_id | manager_id | can_watch_members |
+      | 104      | 103        | 1                 |
+    And the groups ancestors are computed
+    And the database has the following table 'items':
+      | id | url                                                                     | type | entry_participant_type | default_language_tag | text_id |
+      | 10 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | User                   | fr                   | task10  |
+      | 20 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | Team                   | fr                   | task20  |
+      | 30 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | User                   | fr                   | task30  |
+      | 40 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | User                   | fr                   | task40  |
+      | 50 | http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936 | Task | User                   | fr                   | task50  |
```
smadbe, line 27: if all urls are the same, we cannot check they are correct. BTW, use `http://taskplatform/10`, that's easier to read :-)
(beware that the `url` may be related to the `platform` table which provides the keys for signing token ??? to be checked)

app/api/answers/generate_task_token.feature
```diff
@@ -0,0 +1,251 @@
...
+            "platformName": "{{app().Config.GetString("token.platformName")}}",
+            "sLogin": "john"
+          }
+        },
+        "message": "created",
+        "success": true
+      }
+      """
+
+  Scenario: bAccessSolutions is true when item `can_view` >= 'solution'
```
smadbe, line 191: For `bAccessSolutions` (both tests): as the user of the answer is the same as the user making the request, I am not able to spot an error is perm checking (typically checking the perm of the wrong user)

It is strange to write "when item `can_view` >= 'solution'" ... it is the user who has a permission... and here it should even be written that it is the requester.
Same remark for the next test "when item has been validated" -> we have to know that it is by the requester... (it should not be the result related to the answer)

app/api/answers/generate_task_token.feature
```diff
@@ -0,0 +1,251 @@
...
+    And the database has the following table 'results':
+      | attempt_id | participant_id | item_id | started_at          | validated_at        |
+      | 0          | 101            | 10      | 2020-01-01 01:01:01 | null                |
+      | 0          | 101            | 20      | 2020-01-01 01:01:01 | null                |
+      | 0          | 101            | 30      | 2020-01-01 01:01:01 | null                |
+      | 0          | 102            | 20      | 2020-01-01 01:01:01 | null                |
+      | 0          | 103            | 10      | 2020-01-01 01:01:01 | null                |
+      | 0          | 103            | 20      | 2020-01-01 01:01:01 | null                |
+      | 0          | 103            | 30      | null                | null                |
+      | 1          | 103            | 30      | 2020-01-01 01:01:01 | null                |
```
smadbe, line 55: it is a good idea to have an attempt != 0 to test that it is correct attempt_id which is used, but I don't know how you test it in this case. Maybe you could only have only one attempt "1" for this user so that we can check that it is the observed-user attempt_id which is used.

app/api/answers/generate_task_token.feature
```diff
@@ -0,0 +1,251 @@
...
+            "bHintsAllowed": false,
+            "bIsAdmin": false,
+            "bReadAnswers": true,
+            "bSubmissionPossible": false,
+            "idAttempt": "101/0",
+            "idUser": "101",
+            "idItemLocal": "10",
+            "idItem": "task10",
+            "itemUrl": "http://taskplatform.mblockelet.info/task.html?taskId=403449543672183936",
+            "nbHintsGiven": "0",
```
smadbe, line 87: `nbHintsGiven` should be tested as well, it is always at "0" in the tests

smadbe (changes requested):
I have only read the tests for now. I'll review the code after these changes.

app/api/answers/generate_task_token.feature
GeoffreyHuck: fixed

app/api/answers/generate_task_token.feature
GeoffreyHuck: fixed, the permissions checks in the code where wrong indeed :)

app/api/answers/generate_task_token.feature
GeoffreyHuck: fixed

app/api/answers/generate_task_token.feature
GeoffreyHuck: fixed

app/api/answers/generate_task_token.go
```diff
@@ -0,0 +1,202 @@
...
+		Select("1").
+		Limit(1)
+
+	err = store.Answers().WithItems().WithResults().
+		Joins("JOIN groups_ancestors_active ON groups_ancestors_active.child_group_id = ?", user.GroupID).
+		Joins(`
+				JOIN permissions_generated
+					ON permissions_generated.item_id = items.id AND
+						 permissions_generated.group_id = groups_ancestors_active.ancestor_group_id`).
+		Joins("JOIN attempts ON attempts.participant_id = results.participant_id AND attempts.id = results.attempt_id").
```
smadbe, line 129: do you use the `attempts` table for something?

app/api/answers/generate_task_token.go
```diff
@@ -0,0 +1,202 @@
...
+		WhereUserHaveStartedResultOnItem(user).
+		Limit(1).
+		Take(&answerInfos).Error()
+
+	if gorm.IsRecordNotFoundError(err) {
+		return service.InsufficientAccessRightsError
+	}
+	service.MustNotBeError(err)
+
+	fmt.Println(answerInfos)
```
smadbe, line 167: debugging?

smadbe (approved):
Looks good 👍 , except my remark.

app/api/answers/generate_task_token.go
GeoffreyHuck: fixed

app/api/answers/generate_task_token.go
GeoffreyHuck: fixed

PR #886, opened 2023-01-30
Add pre-commit hook to prevent commit with tags in *.feature files

GeoffreyHuck:


GeoffreyHuck:
It's possible to make a process to install the hooks by asking the user to run a script that creates symlinks. It might be overkill for only one hook though, so I've just explained what to do it the doc.

smadbe (approved):

PR #889, opened 2023-01-31
Forum: new permission and propagation

### LINKED ISSUE #883: Forum: new permission and propagation
## Motivations

To support the new "forum" feature (i.e. discussion thread related to participant-item pair), we need a few new tables and services This issue is a prerequisites for the next ones. Technical aspects are discussed in the [dev doc](https://france-ioi.github.io/algorea-devdoc/forum/).

This issue only focus on permissions.

## Subtasks

- [x] new permission (in `permissions_granted` only?): `can_request_help_to` which takes a `group` (FK) as argument
  * NULL can probably be interpreted as "to nobody" (= no perm)
- [x] add a new propagation rule `help_request_propagation` on `items_items`
  - [x] the `PUT /item/:id` service should apply a similar rule as the edit/watch/grant_view propagation for the default value set to `help_request_propagation`, i.e. set the highest possible value depending on user's perm
- the propagation mechanism should not be changed
- [x] Make sure no problem occurs with FK if a group if deleted: `permissions_granted.can_request_help_to` should be set back to NULL if the related group is deleted

- Adding the new permissions to the services returning/updated them will be done later.

GeoffreyHuck:
fixes #883

### DIFF
--- a/app/api/items/items.go
+++ b/app/api/items/items.go
@@ -144,6 +144,9 @@ type itemChild struct {
 	// Defaults to true  if `can_edit` >= 'all_with_grant', false otherwise.
 	// It is always possible to set this field to the same or a lower value, `can_edit` doesn't matter in this case.
 	EditPropagation bool `json:"edit_propagation"`
+	// Can be set to true if `can_request_help_to` can be set (i.e. if user have “can_grant_view ≥ content” on item)
+	// Defaults to true if `can_request_help_to` can be set, false otherwise.
+	RequestHelpPropagation bool `json:"request_help_propagation"`
 }
 
 type insertItemItemsSpec struct {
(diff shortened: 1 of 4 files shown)


app/api/items/items.go
```diff
@@ -144,6 +144,9 @@ type itemChild struct {
 	// Defaults to true  if `can_edit` >= 'all_with_grant', false otherwise.
 	// It is always possible to set this field to the same or a lower value, `can_edit` doesn't matter in this case.
 	EditPropagation bool `json:"edit_propagation"`
+	// Can be set to true if `can_request_help_to` can be set (i.e. if user have “can_grant_view ≥ content” on item)
+	// Defaults to true if `can_request_help_to` can be set, false otherwise.
+	RequestHelpPropagation bool `json:"request_help_propagation"`
```
smadbe, line 149: The same `// It is always possible to set this field to the same or a lower value, `can_edit` doesn't matter in this case.` remark should apply here as well. -> If it is already "true", you can leave send "true" even if you do not have the perm to set it to true otherwise.

smadbe (approved):
Looks good 👍 , just fix that comment.

app/api/items/items.go
GeoffreyHuck: fixed

Issue #769, opened 2021-07-16
Migrate to Gorm v2


zenovich:
Migrate to Gorm v2
Since Gorm v2 supports contexts it will allow us to set timeouts on SQL queries. Also, it is easier to add AddSelect() method to the DB class with Gorm v2 so that columns are appended to the list which is helpful enough.
There are some problems though:
https://github.com/go-gorm/gorm/issues/4525, https://github.com/go-gorm/gorm/issues/4533, https://github.com/go-gorm/gorm/issues/4534

The task is blocked until the issues are fixed.

zenovich:
The branch with changes related to the task: https://github.com/France-ioi/AlgoreaBackend/tree/gormv2

GeoffreyHuck:
It looks like only the second issue wasn't fixed (and won't be) : https://github.com/go-gorm/gorm/issues/4533

Have to check how much implications this have. The issue mentions the case when a table is empty, so hopefully it's possible to work it around.

PR #893, opened 2023-02-07
Forum: initial service

### LINKED ISSUE #884: Forum: initial service
## Motivations

To support the new "forum" feature (i.e. discussion thread related to participant-item pair), we need a few new tables and services This issue is a prerequisites for the next ones. Technical aspects are discussed in the [dev doc](https://france-ioi.github.io/algorea-devdoc/forum/).

## Subtasks

- [x] drop the existing `messages`, `threads` and `user_thread` tables, they will not be used or a bit differently
- [x] create a new `threads` table with as attributes:
  * participant_id, item_id (PK) 
  * status (enum : `waiting_for_participant`, `waiting_for_trainer`, `closed`. Any name fancier than "waiting_for_" is welcome... for instance including "open" (to be discussed first))
  * ~~latest_status_update_at (datetime, not null, default to now)~~
  * helper_group_id (FK to group.id, not null)
  * latest_update_at (datetime, not null, default to now). Comment "Last time a message was posted or the status was updated"
  * message_count (not null number, default to '0'). Comment "Approximation of the number of message sent on the thread."

- [x] Make sure no problem occurs with FK if a group if deleted: `threads.helper_group` should  probably be set to the all users group (it is a config parameter) if the group is deleted

- [x] initial service on `GET /items/:item_id/participant/:participant_id/thread` to retrieve thread information
  - outputs
     - participant_id
     - item_id
     - status, including a status "not_started" is the thread does not exist (no entry)
  
  - Validations
    * one of these conditions matches:
       * the current-user is the thread participant and allowed to `can_view>=content` the item
       * have `can_watch>=answer` permission on the item
       * match all rules:
          * the current-user is descendant of thread `helper_group`
          * the thread is either open (=`waiting_for_participant` or `waiting_for_trainer`), or closed for less than 2 weeks
          * the current-user has validated the item

GeoffreyHuck:
fixes #884 

Almost done, a few questions/observations:

- One big limitation of the actual query system is that we cannot make complex "OR" queries like here: https://gorm.io/docs/advanced_query.html#Group-Conditions . It looks like the GORM version we're using is too old and doesn't support it. For this reason, the current code inefficiently uses multiple queries until we'll update it.

After a few checks:
- We need to investigate why we rely on a fork of an old version of gorm (https://github.com/France-ioi/gorm/commits?author=zenovich), and if we still need this change if we update
- Using a recent version of gorm requires us to upgrade go to a later version

- Questions for the restriction number 3:
```
- the following rules all matches:
  * the current-user is descendant of the thread helper_group
  * the thread is either open (=waiting_for_participant or =waiting_for_trainer), or closed for less than 2 weeks
  * the current-user has validated the item
```

1). If the current-user didn't validate the item but is in a team who did validate it, do we get the thread?
2). If there is a thread and it's closed for more than 2 weeks, do we return "Access denied", or status=not_started?
3). Do the 2 weeks rule also apply with the other restrictions which are:
```
- the current-user is the thread participant and allowed to "can_view >= content" the item
- the current-user has the "can_watch >= answer" permission on the item
```

- Questions for the tests:
Some tests are failing because they check the complete status of the table after a deletion

eg.
```
    And the table "groups_ancestors" should be:
      | ancestor_group_id | child_group_id | is_self |
      | 13                | 13             | 1       |
      | 14                | 14             | 1       |
      | 14                | 21             | 0       |
      | 15                | 13             | 0       |
      | 15                | 15             | 1       |
      | 21                | 21             | 1       |
      | 22                | 13             | 0       |
      | 22                | 14             | 0       |
      | 22                | 21             | 0       |
      | 22                | 22             | 1       |
```

 It is verified that the table contains all the rows that were not deleted, and those are copy-pasted. If we add new rows in the tables to check new functionnalities, those tests will always fail and have to be update.

A solution could be to add a new Gherkin feature `And the table "groups_groups" should stay unchanged but the rows with child_group_id "11" should be deleted`. What do you think ?


smadbe:
>     One big limitation of the actual query system is that we cannot make complex "OR" queries like here: https://gorm.io/docs/advanced_query.html#Group-Conditions . It looks like the GORM version we're using is too old and doesn't support it. For this reason, the current code inefficiently uses multiple queries until we'll update it.

For info, Dmitry started working on updating gorm but faced several issues, see #769 . 

> We need to investigate why we rely on a fork of an old version of gorm (https://github.com/France-ioi/gorm/commits?author=zenovich), and if we still need this change if we update

The fork was required for advanced features (or bug-fixes?) which were missing... to be checked. I suspect it may be a lot of work.

> Using a recent version of gorm requires us to upgrade go to a later version

In itself, there is no problem in upgrading Go to a later version.

> 1). If the current-user didn't validate the item but is in a team who did validate it, do we get the thread?

I omitted the team aspect on purpose for now for simplifying the problem as team may require to use the service with specific args (apply to most existing services).

> 2). If there is a thread and it's closed for more than 2 weeks, do we return "Access denied", or status=not_started?

denied

> 3). Do the 2 weeks rule also apply with the other restrictions which are:
> - the current-user is the thread participant and allowed to "can_view >= content" the item
> - the current-user has the "can_watch >= answer" permission on the item

no

> A solution could be to add a new Gherkin feature And the table "groups_groups" should stay unchanged but the rows with child_group_id "11" should be deleted. What do you think ?

You can add new Gherkin rules, just keep in mind that the goal of these tests is that they should be readable without re-reading their spec each time. In your case, if there are no strange case I miss, that looks good, it improves both readability and writability.

GeoffreyHuck:
I checked the problem with lint. When connecting though ssh to reproduce I faced a "go not found" error, and indeed the path to go was not in PATH. I don't know why it sometimes worked before. It seems to work for now at least.

I added -coverpkg option for go test, which makes the cover go from one package to another to avoid testing the same things multiple times. I then faced a problem with the *test packages not being 100% covered. So in the end I use "go list" to get list of the packages of the project, and remove those ending with "test" from the list.

I also added a Gherkin feature to test for deletion of rows and refactored the database-related Gherkin features because the code was a little messy and it didn't pass the lint.

Finally, I added a pre-commit hook to check the validity of the circleci config if it is modified.

app/database/thread_store.go
```diff
@@ -0,0 +1,99 @@
...
+	// the current-user has validated the item
+
+	now := time.Now()
+	twoWeeksAgo := now.AddDate(0, 0, -14)
+	currentUserCanHelp, err := s.Threads().
+		Joins("JOIN results ON results.item_id = threads.item_id").
+		Joins("JOIN groups_ancestors_active ON groups_ancestors_active.child_group_id = ?", user.GroupID).
+		Where("threads.helper_group_id = groups_ancestors_active.ancestor_group_id").
+		Where("threads.item_id = ?", itemID).
+		Where("threads.status != 'closed' OR (threads.status = 'closed' AND threads.latest_update_at > ?)", twoWeeksAgo).
```
smadbe, line 73: Probably does not really matter in terms of perf but equivalent to (the simpler):
```suggestion
		Where("threads.status != 'closed' OR threads.latest_update_at > ?", twoWeeksAgo).
```

app/database/thread_store.go
```diff
@@ -0,0 +1,99 @@
...
+	var status string
+
+	err := s.Threads().
+		Select("threads.status AS status").
+		Where("threads.participant_id = ?", participantID).
+		Where("threads.item_id = ?", itemID).
+		Limit(1).
+		PluckFirst("status", &status).
+		Error()
+	if err != nil {
```
smadbe, line 94: don't you have the risk here to interpret other errors (e.g., db connection error) to a `not_started` status?
An (ugly?) alternative might be to "union" your query with `select 'not_started' as status`.

db/migrations/2342234112_delete_old_forum_tables.sql
```diff
@@ -0,0 +1,56 @@
+-- +migrate Up
```
smadbe, line 1: The migration file name were corresponding to YYMMDDhhmm ... I suppose "42" is not the month, what is the logic?

smadbe (commented):
Great! Look good 👍 
Just a few remarks/questions.

app/api/items/get_thread.robustness.feature
```diff
@@ -0,0 +1,152 @@
...
+    Given I am the user with id "1"
+    When I send a GET request to "/items/10/participant/1/thread"
+    Then the response code should be 403
+    And the response error message should contain "Insufficient access rights"
+
+  Scenario: >
+      Should be forbidden when
+      the current-user is descendant of the thread helper group
+      and user has validated the item
+      but the thread is closed for 2 weeks
```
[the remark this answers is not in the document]
GeoffreyHuck: Yes, it's two weeks exactly (==). And the next one is to test that it works for more (>). I also wondered if we really need to test this.

app/database/thread_store.go
GeoffreyHuck: fixed. Added check for IsRecordNotFoundError

db/migrations/2342234112_delete_old_forum_tables.sql
GeoffreyHuck: The logic was random ^^'. I edited the names and will change the few others I made in previous changes for consistency.

smadbe (approved):
Ok, good for me! 

(just https://github.com/France-ioi/AlgoreaBackend/pull/893#discussion_r1121429392 maybe)

PR #896, opened 2023-02-07
Make items.text_id unique

### LINKED ISSUE #881: Make items.text_id unique
## Motivations
The `text_id` of `items` is used as "identifier" internal to tasks themselves. It can be used for tasks to request to navigate to another tasks (tasks do not know the `id` of the platform but may know the `text_id` of another task)

## Subtasks
* On existing data: 
   *  set `text_id` to null when it is just a empty string
   * if  several items have the same `text_id` append "2", "3", "4" to duplicates (1 occurrence in former prod copy)
* Add a unique index on `text_id`(the index will be soon used for querying)
* Update the create and update item service doc to inform about the constraint
* On create and update item services: there should be a way for the caller (frontend) to know that an error is a problem of `text_id` uniqueness. -> extract the info from the sql error and pass the info in 403 response

GeoffreyHuck:
fixes #881

### DIFF
--- a/app/api/items/create_item.go
+++ b/app/api/items/create_item.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/go-chi/render"
+	"github.com/go-sql-driver/mysql"
 	"github.com/jinzhu/gorm"
 
 	"github.com/France-ioi/validator"
@@ -27,6 +28,7 @@ type Item struct {
 	EntryFrozenTeams bool `json:"entry_frozen_teams"`
 	// default: false
 	NoScore bool `json:"no_score"`
+	// Unique identifier to reference the task. An error is returned if another task has the same `text_id`
 	// Nullable
 	TextID                 *string `json:"text_id"`
 	DisplayDetailsInParent bool    `json:"display_details_in_parent"`
@@ -235,7 +237,11 @@ func (srv *Service) createItem(w http.ResponseWriter, r *http.Request) service.A
 			}
 
 			// insertion
-			itemID = srv.insertItem(lockedStore, user, formData, &input)
+			itemID, apiError = srv.insertItem(lockedStore, user, formData, &input)
+			if apiError != service.NoError {
+				return apiError.Error
+			}
+
 			return nil
 		})
 
@@ -521,11 +527,11 @@ func registerChildrenValidator(formData *formdata.FormData, store *database.Data
 }
 
 func (srv *Service) insertItem(store *database.DataStore, user *database.User, formData *formdata.FormData,
-	newItemRequest *NewItemRequest) (itemID int64) {
+	newItemRequest *NewItemRequest) (itemID int64, apiError service.APIError) {
 	itemMap := formData.ConstructPartialMapForDB("ItemWithRequiredType")
 	stringMap := formData.ConstructPartialMapForDB("newItemString")
 
-	service.MustNotBeError(store.WithForeignKeyChecksDisabled(func(fkStore *database.DataStore) error {
+	err := store.WithForeignKeyChecksDisabled(func(fkStore *database.DataStore) error {
 		return fkStore.RetryOnDuplicatePrimaryKeyError(func(s *database.DataStore) error {
 			itemID = s.NewID()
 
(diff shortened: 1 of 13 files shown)


app/api/items/create_item.go
```diff
@@ -521,19 +530,26 @@ func registerChildrenValidator(formData *formdata.FormData, store *database.Data
...
 			itemID = s.NewID()
 
 			itemMap["id"] = itemID
 			itemMap["default_language_tag"] = newItemRequest.LanguageTag
 			return s.Items().InsertMap(itemMap)
 		})
-	}))
+	})
+	e, ok := err.(*mysql.MySQLError)
+	if ok && e.Number == 1062 {
```
smadbe, line 547: Looks like a magic number, shouldn't it be a reusable constant in a dedicated file?

app/api/items/create_item.go
```diff
@@ -27,6 +28,7 @@ type Item struct {
 	EntryFrozenTeams bool `json:"entry_frozen_teams"`
 	// default: false
 	NoScore bool `json:"no_score"`
+	// Unique identifier to reference the task. An error is returned if another task has the same `text_id`
```
smadbe, line 31: Shouldn't the doc "explain" how to spot this error in the output and what will be the error code in such a case (400? 403?) ? Or do we consider the way as standard ? 
Could be either in a specific 403 output (more "logic" place) or the input text should state that the error will be a 403 with `text_id` key in `errors` (more convenient).

app/api/items/update_item.go
```diff
@@ -206,13 +206,25 @@ func updateItemInDB(itemData map[string]interface{}, participantsGroupID *int64,
 	}
 
 	err := store.Items().Where("id = ?", itemID).UpdateColumn(itemData).Error()
-	// ERROR 1452 (23000): Cannot add or update a child row: a foreign key constraint fails
-	// (no items_strings for the new default_language_tag)
-	if e, ok := err.(*mysql.MySQLError); ok && e.Number == 1452 {
-		return service.ErrInvalidRequest(formdata.FieldErrors{"default_language_tag": []string{
-			"default language should exist and there should be item's strings in this language",
-		}})
+
```
smadbe, line 209: no doc update in this service?

smadbe (changes requested):

db/migrations/2373317322_items_text_id_unique.sql
```diff
@@ -0,0 +1,50 @@
...
+# Then, we update text_id_unique for each row, and use a TRIGGER to add a number if the value already exists
+# Finally, we rename text_id_unique into text_id_unique and REMOVE the TRIGGER
+
+UPDATE `items` SET `text_id` = NULL WHERE `text_id` = '';
+
+ALTER TABLE `items`
+  ADD COLUMN `text_id_unique` VARCHAR(200) DEFAULT NULL COMMENT 'Unique string identifying the item, independently of where it is hosted'
+  AFTER `text_id`;
+ALTER TABLE `items`
+  ADD UNIQUE `unique_text_id_unique`(`text_id_unique`);
```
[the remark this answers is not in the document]
GeoffreyHuck: Yes, so it'll be at the same place after the migration

db/migrations/2373317322_items_text_id_unique.sql
```diff
@@ -0,0 +1,50 @@
...
+
+-- +migrate StatementBegin
+CREATE TRIGGER itemTextIdUniqueUpdate BEFORE UPDATE ON `items` FOR EACH ROW BEGIN
+  SET @counter = 1;
+  SET NEW.`text_id_unique` = OLD.`text_id`;
+  WHILE exists (SELECT 1 FROM `items` WHERE `items`.`text_id_unique` = NEW.`text_id_unique`) DO
+    SET NEW.`text_id_unique` = CONCAT(OLD.`text_id`, @counter);
+    SET @counter = @counter + 1;
+  END WHILE;
+END
```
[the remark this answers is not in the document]
GeoffreyHuck: Indeed, I didn't know about ROW_NUMBER !

app/api/items/create_item.go
GeoffreyHuck: fixed everywhere

app/api/items/create_item.go
GeoffreyHuck: It's the same as any other field related error: 403 with a key for the field, and the error, along. I changed the doc to just mark the field as "Unique" instead of mentionning the error. To make it the same as when "Required" is mentionned for example.

app/api/items/create_item.go
smadbe: 👍

smadbe (approved):
👍  looks good

PR #897, opened 2023-02-08
Service list all possible breadcrumbs for a started item from a text_id

### LINKED ISSUE #882: New service "List all possible breadcrumbs for a started item" from a text_id
## Overview 

The `text_id` of `items` is used as "identifier" internal to tasks themselves. It can be used for tasks to request to navigate to another tasks (tasks do not know the `id` of the platform but may know the `text_id` of another task)

In order to navigate between items using the "text_id", we need a way to resolve a text_id to an id and list all the ways to get to this item, actually a version of `itemBreadcrumbsFromRootsGet` using a `text_id` instead of an `id`. (another option would be to have a service just for mapping text_id->id, but as it would need to always be used with `itemBreadcrumbsFromRootsGet` for each link, the overhead may be noticeable)

On `GET /items/by-text-id/:text-id/breadcrumbs-from-roots`  if it is possible in the backend router

## Input
* text-id

## Output
Same as itemBreadcrumbsFromRootsGet

## Validations
Same as itemBreadcrumbsFromRootsGet

GeoffreyHuck:
fixes #882


app/api/items/get_breadcrumbs_from_roots.go
```diff
@@ -76,24 +78,73 @@ type breadcrumbElement struct {
...
 
-	user := srv.GetUser(r)
-	participantID := user.GroupID
+	return srv.getBreadcrumbsFromRoots(w, r, itemID)
+}
+
+// swagger:operation GET /items/by-text-id/{item_id}/breadcrumbs-from-roots items itemBreadcrumbsFromRootsByTextIdGet
+// ---
+// summary: List all possible breadcrumbs for a started item using `text_id`
+// description: Same as "List all possible breadcrumbs for a started item using `item_id`"
```
smadbe, line 93: Cannot you do something which would be recognized as a "link" by swagger? (don't lose 2 hours with that)

app/api/items/get_breadcrumbs_from_roots.feature
```diff
@@ -131,3 +138,7 @@ Feature: Find all breadcrumbs to an item
         [{"id": "60", "title": "Reduce Graph", "language_tag": "en", "type": "Task"}]
       ]
       """
+    Examples:
+      | service_url                                   |
+      | /items/60/breadcrumbs-from-roots              |
+      | /items/by-text-id/id60/breadcrumbs-from-roots |
```
smadbe, line 144: Probably it is also worth testing that url decoding work as expected (encoding of spaces, slashes, ...) as `text_id` is a string

smadbe (changes requested):

app/api/items/get_breadcrumbs_from_roots.go
GeoffreyHuck: Yes! updated

app/api/items/get_breadcrumbs_from_roots.feature
GeoffreyHuck: fixed. It only allows alphanum characters now to avoid issues. MySQL migration added to remove other characters. Doc updated too.

The migration strips all non-alphanum characters including accented characters. "créée" will be updated as "cre".

smadbe: To be sure: So you require the text_id to no non-alphanum char? 
 (so sure that's ok, as it is an external id)

GeoffreyHuck: Yes, only a-z, A-Z and 0-9. We could add more but if possible it would be better if we whitelist explicitly the characters, as many characters can potentially make the url parsing fail, like #, ?, /, and we'll most likely forget to test them all.

smadbe: If you do classic url encoding/decoding, you'll most characters we need to support without having to care about #, ?, /, ...

GeoffreyHuck: fixed. using url-encoding/decoding

app/api/items/get_breadcrumbs_from_roots.feature
```diff
@@ -28,13 +28,13 @@ Feature: Find all breadcrumbs to an item
...
-      | id | url                    | type    | default_language_tag | text_id |
-      | 10 | null                   | Chapter | en                   | id10    |
-      | 50 | http://taskplatform/50 | Task    | en                   | id50    |
-      | 60 | http://taskplatform/60 | Task    | en                   | id60    |
-      | 70 | http://taskplatform/70 | Task    | fr                   | id70    |
-      | 80 | null                   | Chapter | en                   | id80    |
-      | 90 | null                   | Chapter | en                   | id90    |
+      | id | url                    | type    | default_language_tag | text_id               |
+      | 10 | null                   | Chapter | en                   | id10                  |
+      | 50 | http://taskplatform/50 | Task    | en                   | -_ '#&?:=/\.,+%¤€aéàd |
```
smadbe, line 33: nice text_id 😁

smadbe (approved):

PR #901, opened 2023-02-27
Rename last migration files with YYMMDDHHII

GeoffreyHuck:
After an error with the migration numbers.

Before making the migration, this SQL has to be run to make sure that the renamed migrations will not be executed twice:

    UPDATE `gorp_migrations` SET id='2301091000_item_type_course_becomes_task.sql' WHERE id='2223181211_item_type_course_becomes_task.sql';
    UPDATE `gorp_migrations` SET id='2301171000_items_strings_image_url_length.sql' WHERE id='2341534429_items_strings_image_url_length.sql';
    UPDATE `gorp_migrations` SET id='2301310900_forum_permissions.sql' WHERE id='2352424510_forum_permissions.sql';
    UPDATE `gorp_migrations` SET id='2302071000_items_text_id_unique.sql' WHERE id='2373317322_items_text_id_unique.sql';


smadbe:
It will break production (the already-deployed migrations) if I do that. I assume that we could do something simple like inserting a migration on 1/01/23 which rename migrations in the db migration table, but it has to be tested locally first.

smadbe (approved):

PR #902, opened 2023-03-07
Forum: update-thread service

GeoffreyHuck:
I started to add higher level Gherkin features for the tests because:
- Creating the database content for every test is very tedious
- I couldn't understand my own tests the day after I wrote them
- The robustness tests are not very robust: if a requirement in the database changes but the tests are not updated, they may continue to pass (ie. returning Insufficient access rights), but without covering the reason they were written for initially

I had a problem when I tried to upgrade go, the fmt would reformat comments and mess up with swagger declarations. That's why there's a lot of spaces adding/removal in swagger definitions in the modifications.


smadbe:
> I started to add higher level Gherkin features for the tests 

Please create a PR for each set of changes  (git cherry-picking is good for that) so that each PR has a specific clear goal and can be reviewed properly.

PR #905, opened 2023-03-08
0) Add verbose to golangci-lint because it keeps crashing

GeoffreyHuck:


smadbe (approved):

app/api/items/create_item.go
```diff
@@ -69,7 +69,7 @@ type Item struct {
 
 // ItemWithRequiredType represents common item fields plus the required type field
 type ItemWithRequiredType struct {
-	Item `json:"item,squash"`
+	Item `json:"item,squash"` // nolint:staticcheck SA5008: unknown JSON option "squash"
```
smadbe, line 72: Cannot you configure the linter so that he does not fail for "squash"? These `nolint` everywhere are degrading readability.

app/structures/structures.go
```diff
@@ -1,3 +1,4 @@
+// Package structures provides item related structures
```
smadbe, line 1: I would say it is not very clear as I don't understand what it stands for when reading it. 
Actually the word structure at the end is not clear.
"item related" -> "item-related"

smadbe (approved):

app/api/items/create_item.go
GeoffreyHuck: In this case, the "squash" option comes from the module https://github.com/France-ioi/mapstructure , which is a fork of https://github.com/mitchellh/mapstructure .

The latest version of mitchellh/mapstructure puts the "squash" option into `mapstructure:",squash"` instead of json, which would remove the error.

We could try to update the fork, or better, to remove the fork if it's possible. The number of modifications doesn't seem high : https://github.com/mitchellh/mapstructure/compare/main...France-ioi:mapstructure:master

app/structures/structures.go
GeoffreyHuck: maybe like this ? // Package structures provides item-related structs.

PR #907, opened 2023-03-08
3. Refactor group queries

GeoffreyHuck:


smadbe:
Putting a description is very important, especially with the title is " Refactor group queries"... I don't know what is the goal / motivation and what I have to check.

app/logging/logger.go
```diff
@@ -1,3 +1,4 @@
+// Package logging provides utilities for logging.
```
smadbe, line 1: isn't related with the lint fix?  anyway..

smadbe (approved):

PR #911, opened 2023-03-08
6) Update thread

GeoffreyHuck:
Implements the update thread service.

This PR also introduces higher level Gherkin features, closer to the language of the specifications for:
- Making the writing of tests easier
- Making the reading of tests easier
- Making the robustness tests more robust. If a new requirement is added, existing robustness tests might return an AccessDenied error for the wrong reason (because the new requirements are not added in the test database). If that happens, they won't fail anymore if what they initially tested is broken. Writing them in a higher level ensures that if some requirements change, those will be reflected in those higher level features, because they same are used in the acceptance tests.
- Making the tests less brittle. If there is a change in the module groups, we shouldn't need to update the tests for the threads
- The closer the code is to the language we express our requirements, the more confident we can be that we don't introduce unnecessary complexity
- Something to focus on: when working on one module (eg. threads), we shouldn't need a deep knowledge of how other modules work (eg. groups, items).

Also:
- insertOrUpdateMaps now updates all the columns provided if the column list is nil

### DIFF
--- a/app/api/threads/update_thread.feature
+++ b/app/api/threads/update_thread.feature
@@ -60,8 +60,8 @@ Feature: Update thread
 
   Scenario: Create a thread if it doesn't exist
     Given I am the user with id "3"
-    And There is no thread with "item_id=1000,participant_id=3"
-    And I am in the group with id "100"
+    And there is no thread with "item_id=1000,participant_id=3"
+    And I am a member of the group with id "100"
     And I can request help to the group with id "100" on the item with id "1000"
     When I send a PUT request to "/items/1000/participant/3/thread" with the following body:
       """
@@ -84,7 +84,7 @@ Feature: Update thread
   #    OR have validated the item.
   Scenario: Can write to thread condition (1) when status is not set
     Given I am the user with id "1"
-    And There is a thread with "item_id=10,participant_id=1"
+    And there is a thread with "item_id=10,participant_id=1"
     When I send a PUT request to "/items/10/participant/1/thread" with the following body:
       """
       {
@@ -99,7 +99,7 @@ Feature: Update thread
 
   Scenario: Can write to thread condition (2) when status is not set
     Given I am the user with id "2"
-    And There is a thread with "item_id=20,participant_id=3"
+    And there is a thread with "item_id=20,participant_id=3"
     And I can watch answer on item with id "20"
     And I can watch the participant with id "3"
     When I send a PUT request to "/items/20/participant/3/thread" with the following body:
@@ -116,7 +116,7 @@ Feature: Update thread
 
   Scenario: Can write to thread condition (3) with can_watch>=answer on the item, when status is not set
     Given I am the user with id "2"
-    And There is a thread with "item_id=30,participant_id=3"
+    And there is a thread with "item_id=30,participant_id=3"
     And I am part of the helper group of the thread
     And I can watch answer on item with id "30"
     When I send a PUT request to "/items/30/participant/3/thread" with the following body:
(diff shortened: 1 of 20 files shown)


app/api/threads/update_thread.go
```diff
@@ -0,0 +1,304 @@
...
+//   Validations and restrictions:
+//     * if `status` is given:
+//       - The participant of a thread can always switch the thread from open to any another other status.
+//         He can only switch it from non-open to an open status if he is allowed to request help on this item.
+//       - A user who has `can_watch>=answer` on the item AND `can_watch_members` on the participant: can always switch
+//         a thread to any open status (i.e. he can always open it but not close it)
+//       - A user who `can write` on the thread can switch from an open status to another open status.
+//     * if `status` is already "closed" and not changing status OR if switching to status "closed":
+//       `helper_group_id` must not be given
+//     * if switching to an open status from a non-closed status: `helper_group_id` must be given
```
smadbe, line 62: The rule is " if switching to an open status from **a non-open status**: helper_group_id must be given" otherwise, this does not make sense.
The implementation seems to implement the correct rule.

app/api/threads/update_thread.robustness.feature
```diff
@@ -0,0 +1,681 @@
...
+      | 100      | 100             | 240     | 100                 |
+      | 100      | 100             | 250     | 100                 |
+      | 100      | 100             | 260     | 100                 |
+      | 100      | 100             | 270     | 100                 |
+      | 12       | 3               | 3000    | 10                  |
+      | 12       | 3               | 3001    | 12                  |
+    And the database has the following table 'threads':
+      | item_id | participant_id | status | helper_group_id | latest_update_at |
+
+  Scenario: Should be logged
```
smadbe, line 76:
```suggestion
  Scenario: Should be logged in
```

app/api/threads/update_thread.feature
```diff
@@ -0,0 +1,386 @@
...
+      """
+    Then the response should be "updated"
+    And the table "threads" should stay unchanged but the row with item_id "20"
+    And the table "threads" at item_id "20" should be:
+      | latest_update_at    | message_count |
+      | 2022-01-01 00:00:00 | 2             |
+
+  Scenario: Can write to thread condition (3) with can_watch>=answer on the item, when status is not set
+    Given I am the user with id "2"
+    And There is a thread with "item_id=30,participant_id=3"
```
smadbe, line 119: there should be an upper case to "There" ... it should be "And there is ... ". Probably you have been mislead by the "I am ..."

smadbe (approved):
Looks good after fixed the few minor comments. 
I haven't checked the implementation and tests of each rules into details. (if would have taken me >the full day)

testhelpers/feature_context.go
```diff
@@ -29,6 +30,20 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^the application config is:$`, ctx.TheApplicationConfigIs)
 	s.Step(`^the context variable "([^"]*)" is "([^"]*)"$`, ctx.TheContextVariableIs)
 
+	s.Step(`^There is a group with "([^"]*)"$`, ctx.ThereIsAGroupWith)
+	s.Step(`^I am a manager of the group with id "([^"]*)"$`, ctx.IAmTheManagerOfTheGroup)
+	s.Step(`^I can watch the participant with id "([^"]*)"$`, ctx.ICanWatchParticipantWithID)
+	s.Step(`^I can view (none|info|content|content_with_descendants|solution) on item with id "([^"]*)"$`,
+		ctx.ICanViewOnItemWithID)
+	s.Step(`^I can watch (none|result|answer|answer_with_grant) on item with id "([^"]*)"$`, ctx.ICanWatchOnItemWithID)
+	s.Step(`^I am in the group with id "([^"]*)"$`, ctx.IAmInTheGroupWithID)
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed

app/api/threads/update_thread.go
GeoffreyHuck: fixed

PR #919, opened 2023-03-16
Fixes for MySQL 8.0.26

GeoffreyHuck:
Some SQL adaptations to fix the bugs introduced by MySQL 8.0.26

-> One is the replacement of a subquery by doing the subquery first and using the results as parameters
-> The other one is the replacement of a WITH (...) by standard joins

Both problems are fixed in Mysql 8.0.28, though I couldn't find the source of the issues in the changelog https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-28.html

It could be related to, but isn't exaclty:
A query using aggregation on a [BIT](https://dev.mysql.com/doc/refman/8.0/en/bit-type.html) type could return different results depending on the indexes or join type employed. This was due to the fact that a DML statement using such an aggregation caches the BIT values using an integer type, and later looks up and converts to a string format for output. The current issue arose because this lookup treated the BIT value as an integer, resulting in an incorrect string value.

The only thing I could do was to use both mysql version side-by-side and see the differences returned by the requests.

Note: Uses PR https://github.com/France-ioi/AlgoreaBackend/pull/918

### DIFF
--- a/app/api/groups/get_granted_permissions.go
+++ b/app/api/groups/get_granted_permissions.go
@@ -155,7 +155,7 @@ func (srv *Service) getGrantedPermissions(w http.ResponseWriter, r *http.Request
 
 	// Used to be a subquery, but it failed with MySQL-8.0.26 due to obscure bugs introduced in this version.
 	// It works when doing the query first and using the result in the second query.
-	// See commit 1abc337a81f83462d4361b9876abee2a82c0e23a
+	// See commit 5a25fbded8134c93c72dc853f72071943a1bd24c
 	managedGroupsWithCanGrantGroupAccessIds := user.GetManagedGroupsWithCanGrantGroupAccessIds(store)
 
 	var sourceGroupsQuery, groupsQuery *database.DB
--- a/app/api/groups/get_granted_permissions.go
+++ b/app/api/groups/get_granted_permissions.go
@@ -153,10 +153,10 @@ func (srv *Service) getGrantedPermissions(w http.ResponseWriter, r *http.Request
 			store.PermissionsGranted().PermissionIsAtLeastSQLExpr("grant_view", "enter")).
 		Select("DISTINCT item_id AS id")
 
-	managedGroupsQuery := store.ActiveGroupAncestors().ManagedByUser(user).
-		Group("groups_ancestors_active.child_group_id").
-		Having("MAX(can_grant_group_access)").
-		Select("groups_ancestors_active.child_group_id AS id")
+	// Used to be a subquery, but it failed with MySQL-8.0.26 due to obscure bugs introduced in this version.
+	// It works when doing the query first and using the result in the second query.
+	// See commit 1abc337a81f83462d4361b9876abee2a82c0e23a
+	managedGroupsWithCanGrantGroupAccessIds := user.GetManagedGroupsWithCanGrantGroupAccessIds(store)
 
 	var sourceGroupsQuery, groupsQuery *database.DB
 	if forDescendants {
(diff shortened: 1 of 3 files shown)


app/api/groups/get_granted_permissions.go
```diff
@@ -170,7 +170,7 @@ func (srv *Service) getGrantedPermissions(w http.ResponseWriter, r *http.Request
 
 		sourceGroupsQuery = store.Groups().
 			Where("groups.type != 'User'").
-			Where("id IN ?", managedGroupsQuery.SubQuery()).
+			Where("id IN (?)", managedGroupsWithCanGrantGroupAccessIds).
```
smadbe, line 173: Mmh ok, that's definitely something we want to get rid of as soon as mysql is fixed. `id IN (list of ids)` may have very bad perf if there are many of ids (unlike the subquery version).

app/api/groups/get_granted_permissions.go
```diff
@@ -153,10 +153,10 @@ func (srv *Service) getGrantedPermissions(w http.ResponseWriter, r *http.Request
 			store.PermissionsGranted().PermissionIsAtLeastSQLExpr("grant_view", "enter")).
 		Select("DISTINCT item_id AS id")
 
-	managedGroupsQuery := store.ActiveGroupAncestors().ManagedByUser(user).
-		Group("groups_ancestors_active.child_group_id").
-		Having("MAX(can_grant_group_access)").
-		Select("groups_ancestors_active.child_group_id AS id")
+	// Used to be a subquery, but it failed with MySQL-8.0.26 due to obscure bugs introduced in this version.
+	// It works when doing the query first and using the result in the second query.
+	// See commit 1abc337a81f83462d4361b9876abee2a82c0e23a
```
smadbe, line 158: Isn't commit change its id with a rebase? I don't see it in this PR.

app/api/items/navigation_data_mapper.go
```diff
@@ -106,54 +106,50 @@ func constructItemListWithoutResultsQuery(dataStore *database.DataStore, groupID
...
-		watchedGroupParticipantsQuery = dataStore.ActiveGroupAncestors().
-			Select("participant.id").
-			Joins(`
-				JOIN `+"`groups`"+` AS participant
-					ON participant.id = groups_ancestors_active.child_group_id AND participant.type IN ('User', 'Team')`).
-			Where("groups_ancestors_active.ancestor_group_id = ?", watchedGroupID).SubQuery()
+
+		// Used to be made with a WITH(), but it failed with MySQL-8.0.26 due to obscure bugs introduced in this version.
+		// It works when we get the groups directly with joins.
+		// See commit 1abc337a81f83462d4361b9876abee2a82c0e23a
```
smadbe, line 119: same here

app/database/user.go
```diff
@@ -40,3 +40,17 @@ func (u *User) CanWatchGroupMembers(store *DataStore, groupID int64) bool {
 
 	return found
 }
+
+// GetManagedGroupsWithCanGrantGroupAccessIds retrieve all group ids that the user manager and for which
```
smadbe, line 44:
```suggestion
// GetManagedGroupsWithCanGrantGroupAccessIds retrieves all group ids that the user manages and for which
```

smadbe (approved):
Please write an issue to revert this PR once MySQL >x is released on our infrastructure.

app/api/groups/get_granted_permissions.go
GeoffreyHuck: fixed

app/api/items/navigation_data_mapper.go
GeoffreyHuck: fixed

app/api/groups/get_granted_permissions.go
GeoffreyHuck: fixed. issue created

PR #923, opened 2023-03-17
Add package comments on all packages

GeoffreyHuck:
So the linter won't come anymore with those errors. And also because it's nice to have package comments.

### DIFF
--- a/app/token/token.go
+++ b/app/token/token.go
@@ -1,3 +1,4 @@
+// Package token provides a way to serialize/un-serialize data structures in an encrypted token.
 package token
 
 import (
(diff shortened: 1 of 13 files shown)


app/token/token.go
```diff
@@ -1,3 +1,4 @@
+// Package token
```
smadbe, line 1: what the difference with `app/token/abstract.go` which is `// Package token provides a way to serialize/un-serialize data structures in an encrypted token.`?

smadbe (approved):

app/token/token.go
GeoffreyHuck: fixed

PR #927, opened 2023-03-21
Add references in Gherkin definitions

GeoffreyHuck:
It is now possible to use references in Gherkin definition with the @ReferenceName notation.

This allows the use of meaningful names instead of writing ids everywhere. It also makes it possible to combine definitions with names and the use of their id.

    Given I am RichardFeynman
    When I send a PUT request to "/items/aaa/participant/@RichardFeynman/thread"`

### DIFF
--- a/app/api/threads/update_thread.robustness.feature
+++ b/app/api/threads/update_thread.robustness.feature
@@ -74,26 +74,26 @@ Feature: Update thread - robustness
       | item_id | participant_id | status | helper_group_id | latest_update_at |
 
   Scenario: Should be logged in
-    When I send a PUT request to "/items/10/participant/1/thread"
+    When I send a PUT request to "/items/@SomeItem/participant/@RichardFeynman/thread"
     Then the response code should be 401
     And the response error message should contain "No access token provided"
 
   Scenario: The item_id parameter should be an int64
-    Given I am the user with id "1"
-    When I send a PUT request to "/items/aaa/participant/1/thread"
+    Given I am RichardFeynman
+    When I send a PUT request to "/items/aaa/participant/@RichardFeynman/thread"
     Then the response code should be 400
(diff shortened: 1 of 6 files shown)


app/api/threads/update_thread.robustness.feature
```diff
@@ -74,35 +74,35 @@ Feature: Update thread - robustness
       | item_id | participant_id | status | helper_group_id | latest_update_at |
 
   Scenario: Should be logged
-    When I send a PUT request to "/items/10/participant/1/thread"
+    When I send a PUT request to "/items/@SomeItem/participant/@RichardFeynman/thread"
```
smadbe, line 77: I think I understand the initial idea but in practice, if I read that, I don't know what ids are used for the requests and so I cannot really check if the behavior is correct. There is no even RichardFeynman in the db on top, what is given as value used in practice in this test for `@SomeItem`, are we sure the rule was not interpret as the string "@SomeItem" here which would explain the 404 ?...  it makes things a bit obscure and so do not really help.
That's just an opinion, we can discuss this.

app/api/threads/update_thread.robustness.feature
```diff
@@ -73,36 +73,40 @@ Feature: Update thread - robustness
...
+    When I send a PUT request to "/items/@Exercice/participant/aaa/thread"
     Then the response code should be 400
     And the response error message should contain "Wrong value for participant_id (should be int64)"
 
   Scenario: Either status, helper_group_id, message_count or message_count_increment must be given
-    Given I am the user with id "1"
-    And there is a thread with "item_id=10,participant_id=1"
-    When I send a PUT request to "/items/10/participant/1/thread" with the following body:
+    Given I am RichardFeynman
+    And there is an item Exercice
```
smadbe, line 98: "Exercice" naming sounds more like a type than a name of a specific item.

PR #929, opened 2023-03-22
Update golangci-lint to 1.27.0

GeoffreyHuck:
I kept having unexpected lint errors on circleci. It was because the version of golandci-lint used was too old for my editor (GoLand) and I wasn't alerted locally.

I updated to 1.27, which is the oldest supported version.

Some new linters have been disabled because they require further work. Details in comments in .golangci.yml.

Also, gofmt requires that the line endings are LF, but we currently had CRLF. So the files have been updated.

The biggest changes in the code are the warnings introduced by the linter https://github.com/jingyugao/rowserrcheck which requires checking rows.Err() after all rows.Next() have been performed. It is also not very flexible in the way it detect those changes, so sometimes the checks had to be put in defer.


.golangci.yml
```diff
@@ -46,14 +46,20 @@ linters-settings:
...
     - maligned
     - prealloc
     - gochecknoglobals
-    - wsl # requires many changes, enable it later
+    - testpackage # requires that go test files are in a different package, but current tests are on private functions
+    - wsl # requires whitelines between blocks. requires many changes, enable it later
+    - nolintlint # requires explanation for each "nolint" directive. requires many changes
+    - gomnd # flags all "magic numbers" in the code. requires many changes
+    - goerr113 # don't allow dynamic errors, requires wrapped static errors. requires many changes
+    - nestif # Doesn't allow too long nested if. requires complex changes
```
smadbe, line 62: yes, indeed, that's a lot of disable... you should plan to start fixing them next week

smadbe (approved):

PR #930, opened 2023-03-23
Fix swagger gen

smadbe:
Quick fix broken swagger gen 

cc @GeoffreyHuck


smadbe:
The doc fix breaks lint. I don't know how to fix both at the same time. I'll let you do that @GeoffreyHuck

PR #932, opened 2023-03-27
Attempt to update to GO1.20 with the latest Godog

GeoffreyHuck:
Updating the project to go 1.20 required many other updates in cascade.

-> Update swagger to last version
Because go 1.18 introduced a change in the formatting of comments that broke swagger doc (https://github.com/go-swagger/go-swagger/pull/2808).
CircleCI config as been updated

-> All swagger docs are updated with tabs instead of spaces
To follow the new guidelines of swagger doc, and also to prevent the comments of breaking every time the file is saved because of automatic reformatting.

-> Update godog to last version
Updated because of an incompatibility with go >= 1.19. The new version doesn't require the godog executable anymore and uses "go test" directly. Makefile has been updated.

-> Update go-messages (types used to handle tables int Gherkin test features)
For compatibility with the last version of godog. The Pickle_*** types have been simplified, which makes the code easier to read.

-> Update linter to the latest version because of an incompatibility with CI
I fixed a few issues, ran automatic formatting tools, and disabled a lot of linters. The number of added linters is really huge and the rate of new ones is fast too. The documentation of golangci-lint shows that only a few of them are enabled by default (https://golangci-lint.run/usage/linters/). Maybe we could change our default policy of "enable-all" and enable only those we are really interested in.

-> Update go to 1.20
Finally there! CircleCI config has been updated.


GeoffreyHuck:
I couldn't make the test coverage work with the latest godog version. But I managed to make it work with the current old one in another PR.

PR #933, opened 2023-03-28
Fix before_update_items trigger and fix the corresponding test

GeoffreyHuck:
When an item's url was updated to something that didn't match any platform regex, the platform_id wasn't set to NULL.

The corresponding test was written in a weird way that made the test pass:
- The initial data was inconsistent: it's not possible to have an url "1234" with a platform_id 4 where the regex states `don't start with a "4"`.
- There is no reason to update the platform_id to 4 prior to the tests. Or is it to test something else that I'm missing? In this case we should make a separate test. 

After the update to go1.20, the test doesn't pass anymore. I checked the trigger directly into mysql and noticed that it wasn't working in this case. I'm still not sure why the test used to pass prior to the update.

### DIFF
--- a/app/database/item_store_integration_test.go
+++ b/app/database/item_store_integration_test.go
@@ -1063,7 +1063,7 @@ func TestItemStore_TriggerBeforeUpdate_SetsPlatformID(t *testing.T) {
 		updateMap      map[string]interface{}
 		wantPlatformID *int64
 	}{
-		{name: "url is unchanged", updateMap: map[string]interface{}{"type": "Chapter"}, wantPlatformID: ptrInt64(4)},
+		{name: "url is unchanged", updateMap: map[string]interface{}{"type": "Chapter"}, wantPlatformID: ptrInt64(1)},
 		{name: "new url is null", updateMap: map[string]interface{}{"url": nil}, wantPlatformID: nil},
 		{name: "chooses a platform with higher priority", updateMap: map[string]interface{}{"url": ptrString("12345")},
 			wantPlatformID: ptrInt64(2)},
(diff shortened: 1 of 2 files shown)


app/database/item_store_integration_test.go
```diff
@@ -1063,7 +1063,7 @@ func TestItemStore_TriggerBeforeUpdate_SetsPlatformID(t *testing.T) {
 		updateMap      map[string]interface{}
 		wantPlatformID *int64
 	}{
-		{name: "url is unchanged", updateMap: map[string]interface{}{"type": "Chapter"}, wantPlatformID: ptrInt64(4)},
+		{name: "url is unchanged", updateMap: map[string]interface{}{"type": "Chapter"}, wantPlatformID: ptrInt64(1)},
```
smadbe, line 1066: Was working as "url" wasn't changed, it was not recomputed so still `4`.

app/database/item_store_integration_test.go
```diff
@@ -1074,18 +1074,18 @@ func TestItemStore_TriggerBeforeUpdate_SetsPlatformID(t *testing.T) {
...
+					- {id: 4, regexp: "^2.*", priority: 2}
+					- {id: 3, regexp: "^1.*", priority: 1}
 				languages: [{tag: fr}]
 				items:
-					- {id: 1, platform_id: 4, url: 1234, default_language_tag: fr}`)
+					- {id: 1, platform_id: 1, url: 444, default_language_tag: fr}`)
 			defer func() { _ = db.Close() }()
 
 			itemStore := database.NewDataStore(db).Items()
-			assert.NoError(t, itemStore.ByID(1).UpdateColumn("platform_id", 4).Error())
```
smadbe, line 1087: The reason was probably that because the `url` was changed during `SetupDBWithFixtureString`, it was probably resetted to its correct value "2".

db/migrations/2303281630_fix_before_update_items_trigger.sql
```diff
@@ -0,0 +1,38 @@
...
+CREATE TRIGGER `before_update_items` BEFORE UPDATE ON `items` FOR EACH ROW
+BEGIN
+  IF NOT OLD.url <=> NEW.url THEN
+    IF NEW.url IS NOT NULL THEN
+      SET @platformID = (SELECT platforms.id FROM platforms WHERE NEW.url REGEXP platforms.regexp ORDER BY platforms.priority DESC LIMIT 1);
+
+      SET NEW.platform_id=@platformID;
+    ELSE
+      SET NEW.platform_id = NULL;
+    END IF;
```
smadbe, line 14: why this change? Isn't it just another way to write the same think in SQL?

smadbe: probably you don't not the `@platformID` local var anymore (but not really important)

smadbe (approved):

PR #936, opened 2023-03-29
Latest lint

GeoffreyHuck:
-> Update linter to the latest version because of an incompatibility with CI and go1.20
I fixed a few issues, ran automatic formatting tools, and disabled a lot of linters. The number of added linters is really huge and the rate of new ones is fast too. The documentation of golangci-lint shows that only a few of them are enabled by default (https://golangci-lint.run/usage/linters/). Maybe we could change our default policy of "enable-all" and enable only those we are really interested in?


smadbe (approved):
yes, we should take some time to review the linter to see those which may be relevant

Issue #888, opened 2023-01-30
Forum: Thread listing service


smadbe:
## Overview 

Service to list the visible threads for a user.

On `GET /threads/`

## Input

Query args:
* `watched_group_id`
* `is_mine` (bool)
* `item_id` (optional)

Exactly one of [`watched_group_id`, `is_mine`] is given

## Output

* List of thread info with:
  - `item` : 
     - id
     - type
     - title
     - language_tag
  - `participant`
    - id 
    - login
    - first_name
    - last_name
  - `status`
  - `latest_update_at`
  - `message_count`

* If `is_mine = 1`, only threads whose the participant is the current-user and whose the item is visible (can_view >= content) by the current-user are returned
* If `is_mine = 0`, only [threads that the current-user can list](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/#listing--reading-a-thread) and whose the current-user is NOT the participant are returned
* If watched_group_id is given, only threads whose the participant is descendant (including self) of the `watched_group_id` are returned
* If `item_id` is given, only threads whose the `item_id` is or is descendant of the given `item_id` are returned
* `first_name` and `last_name` are only returned for the current user or if the user approved access to their personal info for some group managed by the current user

## Validations
* if `watched_group_id` given: the current-user must be (implicitly or explicitly) a manager with `can_watch_members` of `watched_group_id`
* if `item_id` is given, the current user needs `can_view >= content' on it

## Extra

* Use ordering and page limit as usual. By default, ordering should be `latest_update_at` desc (+ tie-breaking rules)
* The service should support filtering by `status` 
* The service should support filtering by `latest_update_gt` only return entries whose the `latest_update_at` is greater than a given datetime (given in UTC8601)

GeoffreyHuck:
A question regarding: if

>  `item_id` is given, the current user needs `can_view >= content' on it

Is the can_view >= content only required for the `item_id`parameter, or for all the threads on children that can be returned ?

Also, when watched_group_id is used, we don't check can_view at all. Is it correct ?

smadbe:
> > if `item_id` is given, the current user needs `can_view >= content' on it
> 
> Is the can_view >= content only required for the `item_id`parameter, or for all the threads on children that can be returned ?

Yes, for all.

> Also, when watched_group_id is used, we don't check can_view at all. Is it correct ?

No. We cannot visit an item cannot view all (nor we can see the title), so there is no point / no way of listing them.

PR #941, opened 2023-04-04
List threads service with watch_group_id

GeoffreyHuck:
Issue #888 

Only implemented `watch_group_id` for now.

A new package have been included: [github.com/PaesslerAG/jsonpath](https://github.com/PaesslerAG/jsonpath)

It is used so that we can directly check that some value at some location in the JSON matches. [*] means that the order doesn't matter. With this functionality, we don't need to verify the whole response for each test anymore, so:

- The tests are less brittle. If we add/rename/remove a field, or change the order of the fields in the response, we don't have to change all the tests anymore
-  The responsibility of each test is more focused on a single functionality, rather than each test testing the same things many times, + one single thing
- Easier to visualize when the result contains a list of items: only one line per item

The definitions are a higher level than the database tables to avoid defining too many tables and having to cross-check the references every time.

I already added the test data for the implementation of `is_mine`, which is not implement yet, because it had a big impact on the structure of the data.

### DIFF
--- a/app/api/threads/list_threads.feature
+++ b/app/api/threads/list_threads.feature
@@ -1,0 +1,97 @@
+Feature: Get threads
+  Background:
+    Given there are the following users:
+      | name           | @reference      |
+      | RichardFeynman | @RichardFeynman |
+      | StevenHawking  | @StevenHawking  |
+      | DavidBowie     | @DavidBowie     |
+      | ClaireStudent  | @ClaireStudent  |
+    Given there are the following threads:
+      | participant_id  |
+      | @RichardFeynman |
+      | @ClaireStudent  |
+      | @StevenHawking  |
+      | @DavidBowie     |
+
+  Scenario: Should have all the fields properly set
+    Given I am MarieCurie
+    And there is a group Laboratory referenced by @Laboratory
+    And I am a manager of the group Laboratory
+    And I can watch the group Laboratory
+    And there are the following users:
+      | @reference      | login          | first_name | last_name |
+      | @AlbertEinstein | AlbertEinstein | Albert     | Einstein  |
+      | @PaulDirac      | PaulDirac      | Paul       | Dirac     |
+      And AlbertEinstein the scientist is a member of the group Laboratory
+    And AlbertEinstein has approved access to his personal info for the group Laboratory
+    And PaulDirac the scientist is a member of the group Laboratory
+    And the database has the following table 'items':
+      | id | type | default_language_tag |
+      | 1  | Task | fr                   |
+      | 2  | Task | en                   |
+    And the database has the following table 'items_strings':
+      | item_id | language_tag | title      |
+      | 1       | en           | Beginning  |
+      | 1       | fr           | Debut      |
+      | 2       | en           | Experiment |
+    And the database has the following table 'threads':
+      | item_id | participant_id  | status                  | message_count | latest_update_at    | helper_group_id |
+      | 1       | @AlbertEinstein | waiting_for_trainer     | 0             | 2023-01-01 00:00:01 | @Laboratory     |
+      | 2       | @PaulDirac      | waiting_for_participant | 1             | 2023-01-01 00:00:02 | @Laboratory     |
+    When I send a GET request to "/threads?watched_group_id=@Laboratory"
+    Then the response code should be 200
+    And the response body should be, in JSON:
+    """
+      [
+        {
+          "item": {
+            "id": "1",
+            "language_tag": "fr",
+            "title": "Debut",
+            "type": "Task"
+          },
+          "latest_update_at": "2023-01-01T00:00:01Z",
+          "message_count": 0,
+          "participant": {
+            "id": "@AlbertEinstein",
+            "login": "AlbertEinstein",
+            "first_name": "Albert",
+            "last_name": "Einstein"
+          },
+          "status": "waiting_for_trainer"
+        },
+        {
+          "item": {
+            "id": "2",
+            "language_tag": "en",
+            "title": "Experiment",
+            "type": "Task"
+          },
+          "latest_update_at": "2023-01-01T00:00:02Z",
+          "message_count": 1,
+          "participant": {
+            "id": "@PaulDirac",
+            "login": "PaulDirac",
+            "first_name": "",
+            "last_name": ""
+          },
+          "status": "waiting_for_participant"
+        }
+      ]
+    """
+
+  Scenario: Should get the threads whose the participant is a descendant (or self) of the watched_group_id
+    Given I am RichardFeynman
+    And there is a group University referenced by @University
+    And I am a manager of the group University
+    And I can watch the group University
+    And the group FirstYear is a descendant of the group University
+    And ClaireStudent the student is a member of the group FirstYear
+    And StevenHawking the professor is a member of the group University
+    When I send a GET request to "/threads?watched_group_id=@University"
+    Then the response code should be 200
+    And it should be a JSON array with 2 entries
+    And the response should match the following JSONPath:
+      | JSONPath            | value          |
+      | $[*].participant.id | @ClaireStudent |
+      | $[*].participant.id | @StevenHawking |
--- a/app/api/threads/list_threads.feature
+++ b/app/api/threads/list_threads.feature
@@ -1,7 +1,7 @@
 Feature: List threads
   Background:
     And there are the following groups:
-      | name          | parent        | members                                               |
+      | group         | parent        | members                                               |
       | @Consortium   |               | @ConsortiumMember                                     |
       | @A_University | @Consortium   | @A_UniversityMember,@A_UniversityManagerCanWatch      |
       | @B_University | @Consortium   | @B_UniversityMember                                   |
@@ -11,7 +11,7 @@ Feature: List threads
       | @B_Class      | @B_Section    |                                                       |
       | @OtherGroup   |               | @OtherGroupMember                                     |
     And @A_UniversityManagerCanWatch is a manager of the group @A_University and can watch its members
-    And there are the following item tasks:
+    And there are the following tasks:
       | item                                            |
       | @B_SectionMember2_CanViewInfo                   |
       | @B_SectionMember2_CanViewContent1               |
@@ -38,12 +38,6 @@ Feature: List threads
       | @B_SectionMember2_CanViewContent1               | @B_SectionMember2   | content                  |           |
       | @B_SectionMember2_CanViewContent2               | @B_SectionMember2   | content                  |           |
       | @B_SectionMember2_CanViewContentWithDescendants | @B_SectionMember2   | content_with_descendants |           |
-      | @B_UniversityMember_HasValidated1               | @B_UniversityMember |                          |           |
-      | @B_UniversityMember_HasValidated2               | @B_UniversityMember |                          |           |
-      | @B_UniversityMember_HasValidated3               | @B_UniversityMember |                          |           |
-      | @B_UniversityMember_HasValidated4               | @B_UniversityMember |                          |           |
-      | @B_UniversityMember_HasValidated5               | @B_UniversityMember |                          |           |
-      | @B_UniversityMember_HasValidated6               | @B_UniversityMember |                          |           |
       | @B_UniversityMember_HasNotValidated             | @B_UniversityMember |                          | answer    |
       | @B_UniversityMember_CanWatchAnswer1             | @B_UniversityMember |                          | answer    |
       | @B_UniversityMember_CanWatchAnswer2             | @B_UniversityMember |                          | answer    |
@@ -85,7 +79,7 @@ Feature: List threads
     Given I am @LaboratoryManagerCanWatch
     And I am a manager of the group @Laboratory and can watch its members
     And there are the following users:
-      | login                                               | first_name            | last_name            |
+      | user                                                | first_name            | last_name            |
       | @LaboratoryMember_WithApprovedAccessPersonalInfo    | FirstName_Approved    | LastName_Approved    |
       | @LaboratoryMember_WithoutApprovedAccessPersonalInfo | FirstName_NotApproved | LastName_NotApproved |
     And @LaboratoryMember_WithApprovedAccessPersonalInfo is a member of the group @Laboratory who has approved access to his personal info
@@ -148,15 +142,14 @@ Feature: List threads
   Scenario: Should get the threads whose the participant is a descendant of the watched_group_id
     Given I am @A_UniversityManagerCanWatch
     When I send a GET request to "/threads?watched_group_id=@A_University"
-    And it should be a JSON array with 2 entries
     And the response at $[*].participant.id should be:
       | @A_UniversityMember |
-      | @A_ClassMember              |
+      | @A_ClassMember      |
 
   Scenario: Should get the threads whose the participant is equal to the watched_group_id
     Given I am @A_UniversityManagerCanWatch
     When I send a GET request to "/threads?watched_group_id=@A_ClassMember"
-    And it should be a JSON array with 1 entries
+    And the response should be a JSON array with 1 entries
     And the response at $[0].participant.id should be "@A_ClassMember"
 
   Scenario: Should return only the threads in which the participant is the current user and the item is visible when is_mine=1
--- a/app/api/threads/list_threads.feature
+++ b/app/api/threads/list_threads.feature
@@ -1,19 +1,79 @@
 Feature: Get threads
   Background:
     Given there are the following users:
-      | name           | @reference      |
-      | RichardFeynman | @RichardFeynman |
-      | StevenHawking  | @StevenHawking  |
-      | DavidBowie     | @DavidBowie     |
-      | ClaireStudent  | @ClaireStudent  |
+      | login               |
+      | PresidentConsortium |
+      | PresidentUniversity |
+      | RichardFeynman      |
+      | Mary                |
+      | EtienneKlein        |
+      | Charlotte           |
+      | Baptiste            |
+      | Thibaut             |
+      | DavidBowie          |
+    And there are the following groups:
+      | parent               | name                 |
+      |                      | UniversityConsortium |
+      | UniversityConsortium | University           |
+      | University           | PresidentUniversity  |
+      | University           | FirstYear            |
+      | FirstYear            | Classroom            |
+      | UniversityConsortium | Université           |
+      | Université           | PremièreAnnée        |
+      | PremièreAnnée        | Classe               |
+      |                      | Superstar            |
+    And there are the following group members:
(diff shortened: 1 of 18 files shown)


app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
+Feature: Get threads
+  Background:
+    Given there are the following users:
+      | login               |
+      | PresidentConsortium |
+      | PresidentUniversity |
```
smadbe, line 6: `ConsortiumPresident`, `UniversityPresident`instead?

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+      | group                | member              |
+      | UniversityConsortium | PresidentConsortium |
+      | University           | RichardFeynman      |
+      | Classroom            | Mary                |
+      | Université           | EtienneKlein        |
+      | PremièreAnnée        | Charlotte           |
+      | PremièreAnnée        | Baptiste            |
+      | PremièreAnnée        | Thibaut             |
+      | Superstar            | DavidBowie          |
+    And there are the following items with permissions:
```
smadbe, line 35: In the same way as "groups", it is items with permissions (a 1-n relationship btw, strange to show it as a table), I would expect the "key" (so item) to be the first column.

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    Given there are the following threads:
+      | participant         | item                                  | helper_group         | status                  | latest_update_at     | comment                                                                                                                |
+      | PresidentConsortium |                                       |                      |                         |                      |                                                                                                                        |
+      | PresidentUniversity |                                       |                      |                         |                      |                                                                                                                        |
+      | Mary                |                                       |                      |                         |                      |                                                                                                                        |
```
smadbe, line 56: Threads have participant and item as key so I don't know how to interpret this. I should not have to try to guess what it means.

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+      | Thibaut             | BaptisteCanViewContent2               |                      |                         |                      | Baptiste is_mine=1 -> notok: not the participant                                                                       |
+      | Baptiste            | BaptisteCanViewContentWithDescendants |                      |                         |                      | Baptiste is_mine=1 -> ok: can_view >= content                                                                          |
+      | DavidBowie          |                                       |                      |                         |                      |                                                                                                                        |
+    And the time now is "2022-01-01T00:00:00Z"
+
+  Scenario: Should have all the fields properly set, including first_name and last_name when the access is approved
+    Given I am MarieCurie
+    And there is a group Laboratory referenced by @Laboratory
```
smadbe, line 78: Doesn't gherkin rule usually quote string so that the reader know what is part of a static rule and what is a "variable"?

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    And I am a manager of the group Laboratory
+    And I can watch the group Laboratory
```
smadbe, line 80: This rule makes only sense for manager so I think it should be merged with the previous rule.

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    And there are the following users:
+      | @reference      | login          | first_name | last_name |
+      | @AlbertEinstein | AlbertEinstein | Albert     | Einstein  |
+      | @PaulDirac      | PaulDirac      | Paul       | Dirac     |
```
smadbe, line 84: I don't get why I have "@..." while I have no "@..." for user definitions at the top of the file.
I like the gherkin convention that "background" rules are the same as "scenario" ones.

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    And AlbertEinstein the scientist is a member of the group Laboratory
```
smadbe, line 85: "the scientist" ? what is this info used for? 🤔

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    And AlbertEinstein has approved access to his personal info for the group Laboratory
```
smadbe, line 86: it is a property of membership, so it is better if directly with it

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+      | Thibaut             | BaptisteCanViewContent2               |                      |                         |                      | Baptiste is_mine=1 -> notok: not the participant                                                                       |
+      | Baptiste            | BaptisteCanViewContentWithDescendants |                      |                         |                      | Baptiste is_mine=1 -> ok: can_view >= content                                                                          |
+      | DavidBowie          |                                       |                      |                         |                      |                                                                                                                        |
+    And the time now is "2022-01-01T00:00:00Z"
+
+  Scenario: Should have all the fields properly set, including first_name and last_name when the access is approved
+    Given I am MarieCurie
+    And there is a group Laboratory referenced by @Laboratory
+    And I am a manager of the group Laboratory
```
smadbe, line 79: you just wrote that the group if referenced by ` @Laboratory` and then you reference to it without using it ;-)

app/api/threads/list_threads.feature
smadbe, line 78: probably no need to repeat twice "Laboratory"

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    And I can watch the group Laboratory
+    And there are the following users:
+      | @reference      | login          | first_name | last_name |
+      | @AlbertEinstein | AlbertEinstein | Albert     | Einstein  |
+      | @PaulDirac      | PaulDirac      | Paul       | Dirac     |
+    And AlbertEinstein the scientist is a member of the group Laboratory
+    And AlbertEinstein has approved access to his personal info for the group Laboratory
+    And PaulDirac the scientist is a member of the group Laboratory
+    And the database has the following table 'items':
```
smadbe, line 88: the rest of the scenario is clear

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+    And I can watch the group University
+    And the user Mary is referenced by @Mary
+    And the user PresidentUniversity is referenced by @PresidentUniversity
+    And the group University is referenced by @University
+    When I send a GET request to "/threads?watched_group_id=@University"
+    And it should be a JSON array with 2 entries
+    And the response should match the following JSONPath:
+      | JSONPath            | value                |
+      | $[*].participant.id | @PresidentUniversity |
+      | $[*].participant.id | @Mary                |
```
smadbe, line 154: Actually I used json path equivalent in gherkin tests for another project. That works well as long as you have a least an example of full response (as you did above).

I don't think there is a benefit from using a gherkin table anymore when using jsonpath (except if the value to check can be represented as a table with an actual benefit). I would be shorter/clearer to have rules like `And the response at $[*].participant.id should be "...." , and possibly have several of this rule.

EDIT: uhm here you want to check a list actually... so what what I would expect instead:
```
And the response at $[*].participant.id should be:
  | value                |
  | @PresidentUniversity |
  | @Mary                |
```

smadbe (changes requested):
here are my initial comments about the feature files. I haven't read the rule implementation but I don't think it should to understand the tests. As discussed beware at high-level abstractions that may make thing less clear than the previous method.

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
GeoffreyHuck: fixed, there is a version for multiple checks with a table, and the inline for a single check

app/api/threads/list_threads.feature
GeoffreyHuck: It's up to the rule definition, it would make sense here to add " because it's not a natural sentence

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
GeoffreyHuck: fixed, it was to add contextual information, but the scenario was updated since and it doesn't add anything anymore

GeoffreyHuck: It can improve readability when a scenario features managers and students at the same time, and when the names don't obviously indicate who's the manager and who's the student. We'll see when we have the case

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,172 @@
...
+      | Baptiste     | BaptisteCanViewInfo                   |               | info                     |           |
+      | Baptiste     | BaptisteCanViewContent1               |               | content                  |           |
+      | Baptiste     | BaptisteCanViewContent2               |               | content                  |           |
+      | Baptiste     | BaptisteCanViewContentWithDescendants |               | content_with_descendants |           |
+      | EtienneKlein | EtienneKleinHasValidated1             | 1             |                          |           |
+      | EtienneKlein | EtienneKleinHasValidated2             | 1             |                          |           |
+      | EtienneKlein | EtienneKleinHasValidated3             | 1             |                          |           |
+      | EtienneKlein | EtienneKleinHasValidated4             | 1             |                          |           |
+      | EtienneKlein | EtienneKleinHasValidated5             | 1             |                          |           |
+      | EtienneKlein | EtienneKleinHasValidated6             | 1             |                          |           |
```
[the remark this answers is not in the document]
GeoffreyHuck: Since "has_validated" is something between an item and a user, it seemed a good idea to put it there. But there is a semantic problem: it doesn't make sense to say that "a group" validated an item, does it?

Maybe we can separate the definitions and add something like:
```
    And there are the following items which have been validated:
      | item                      | member       |
      | EtienneKleinHasValidated1 | EtienneKlein |
      | EtienneKleinHasValidated2 | EtienneKlein |
      | EtienneKleinHasValidated3 | EtienneKlein |
      | EtienneKleinHasValidated4 | EtienneKlein |
      | EtienneKleinHasValidated5 | EtienneKlein |
      | EtienneKleinHasValidated6 | EtienneKlein |
```

Would that be better? Also, is "member" the right name here?

app/api/threads/list_threads.feature
GeoffreyHuck: In this case it means that the item can be whatever item because its parameters don't matter.

The alternative would be to do:
```
participant	item
ConsortiumPresident	Item1
UniversityPresident	Item2
```

But since we don't need to define anything more on those items, that would be the only place where they appear. Would that be clear enough?

app/api/threads/list_threads.feature
GeoffreyHuck: The implementation of those @... feels a bit hacky. There're only here to accommodate for when we need to explicitly get the ID of something, in those cases:

url:
- `When I send a GET request to "/threads?watched_group_id=@Laboratory"`

responses:
- `And the response at $[0].participant.id should be "@Mary"`
-
```
"participant": {
            "id": "@AlbertEinstein",
            "login": "AlbertEinstein",
            "first_name": "Albert",
            "last_name": "Einstein"
},
```


**Internally, the references already exist** because they are used by the Gherkin features. For example:
`And I am a manager of the group Laboratory and can watch its members`
Laboratory is a reference here, but we know with the context that it's a reference. We could add @... everywhere but it would not be natural.

In the urls and in the responses, we don't know when we're referring to an entity's id.

This means that statements like:
`And the group University is referenced by @University`
could be removed, and the test would still work. But this line doesn't do nothing: it's possible to replace the name of the reference and it would create this other reference.

The options that we have here are:
- Replacing every `Name` by `@Name`, in all tables and all Gherkin definitions, every time we refer to an entity. It would remove ambiguity but reduce readability.
- Removing all the definitions of `@references` in tables and `referenced by` in features because it works without it. Simpler but has a feeling of "magic" in it.
- A hybrid version like we have now, with explicit references declarations, even if it's technically not necessary. Not sure if any better than the two other options?
- Another solution?

app/api/threads/list_threads.feature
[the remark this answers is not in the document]
smadbe: > Also, is member the right name here?

"participant" (=team or user) would be the right name (see `participant_id` in the `results` table ... actually to avoid confusion, naming should be close to those use in the db)`

> Would that be better?

It is just related to what you put in your `results` table... couldn't it just be the result table? or a "factory" around this table

app/api/threads/list_threads.feature
smadbe: > In this case it means that the item can be whatever item because its parameters don't matter.

I don't understand what you mean

app/api/threads/list_threads.feature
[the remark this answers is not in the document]
GeoffreyHuck: fixed, added validated in "And there are the following results:"

app/api/threads/list_threads.feature
GeoffreyHuck: What I mean is that we don't need to define anything special about this item, the only thing is that it must exist so that the thread can be created, to fulfill database constraints.

smadbe:
Missing ref to the issue it is fixing.

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,145 @@
...
+  Background:
+    And there are the following groups:
+      | name                 | parent               | members                            |
+      | UniversityConsortium |                      | ConsortiumPresident                |
+      | University           | UniversityConsortium | UniversityPresident,RichardFeynman |
+      | FirstYear            | University           |                                    |
+      | Classroom            | FirstYear            | Mary                               |
+      | Université           | UniversityConsortium | EtienneKlein                       |
+      | PremièreAnnée        | Université           | Charlotte,Baptiste,Thibaut         |
+      | Classe               | PremièreAnnée        |                                    |
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed, with A and B it is indeed easier to read

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,168 @@
...
+          },
+          "status": "waiting_for_participant"
+        }
+      ]
+    """
+
+  Scenario: Should get the threads whose the participant is a descendant of the watched_group_id
+    Given I am @A_UniversityManagerCanWatch
+    When I send a GET request to "/threads?watched_group_id=@A_University"
+    And it should be a JSON array with 2 entries
```
smadbe, line 151: The check seems redundant with the fact that the next table has two rows

smadbe (changes requested):
few minor remarks left

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,168 @@
...
+    And the response at $[*].participant.id should be:
+      | @A_UniversityMember |
+      | @A_ClassMember              |
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,168 @@
...
+      | @Consortium   |               | @ConsortiumMember                                     |
+      | @A_University | @Consortium   | @A_UniversityMember,@A_UniversityManagerCanWatch      |
+      | @B_University | @Consortium   | @B_UniversityMember                                   |
+      | @A_Section    | @A_University |                                                       |
+      | @B_Section    | @B_University | @B_SectionMember1,@B_SectionMember2,@B_SectionMember3 |
+      | @A_Class      | @A_Section    | @A_ClassMember                                        |
+      | @B_Class      | @B_Section    |                                                       |
+      | @OtherGroup   |               | @OtherGroupMember                                     |
+    And @A_UniversityManagerCanWatch is a manager of the group @A_University and can watch its members
+    And there are the following item tasks:
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,168 @@
...
+      | @B_SectionMember2_CanViewInfo                   | @B_SectionMember2   | info                     |           |
+      | @B_SectionMember2_CanViewContent1               | @B_SectionMember2   | content                  |           |
+      | @B_SectionMember2_CanViewContent2               | @B_SectionMember2   | content                  |           |
+      | @B_SectionMember2_CanViewContentWithDescendants | @B_SectionMember2   | content_with_descendants |           |
+      | @B_UniversityMember_HasValidated1               | @B_UniversityMember |                          |           |
+      | @B_UniversityMember_HasValidated2               | @B_UniversityMember |                          |           |
+      | @B_UniversityMember_HasValidated3               | @B_UniversityMember |                          |           |
+      | @B_UniversityMember_HasValidated4               | @B_UniversityMember |                          |           |
+      | @B_UniversityMember_HasValidated5               | @B_UniversityMember |                          |           |
+      | @B_UniversityMember_HasValidated6               | @B_UniversityMember |                          |           |
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed, not required anymore

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,168 @@
+Feature: List threads
+  Background:
+    And there are the following groups:
+      | name          | parent        | members                                               |
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed, renamed the column group, which is consistent with the rest.

app/api/threads/list_threads.feature
```diff
@@ -0,0 +1,162 @@
...
+          },
+          "status": "waiting_for_participant"
+        }
+      ]
+    """
+
+  Scenario: Should get the threads whose the participant is a descendant of the watched_group_id
+    Given I am @A_UniversityManagerCanWatch
+    When I send a GET request to "/threads?watched_group_id=@A_University"
+    And it should be a JSON array with 2 entries
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed

PR #943, opened 2023-04-11
ListThread service: add is_mine parameter

GeoffreyHuck:
Issue #888 

This PR handles the `is_mine` parameter for the listThread service.

Some rights between this and the getThread have been factorized into `WhereUserCanHelp`:
```
	// the following rules all matches:
	// the current-user is descendant of the thread helper_group
	// the thread is either open (=waiting_for_participant or =waiting_for_trainer), or closed for less than 2 weeks
	// the current-user has validated the item
```

The function `NewThreadStore` has been added because when we use a function defined on *DB, we end up being unable to use functions defined on *ThreadStore anymore, because we get a *DB in return. The reason relates to types: A *ThreadStore is a *DataStore is a *DB, but a *DB is not a *ThreadStore.

I'm not sure what's the best thing to do here, maybe there exists a pattern that would handle this case smoothly in go. The code seems to struggle with this everywhere, where the pattern used is:
```
func (s *ItemAncestorStore) DescendantsOf(ancestorID int64) *ItemItemStore {
	return &ItemItemStore{NewDataStoreWithTable(
		s.Where("items_ancestors.ancestor_item_id = ?", ancestorID), s.tableName,
	)}
}
```
At other times, it returns a *DB instead of a DataStore to avoid this issue. But then we lose consistence with some methods defined on a specific store which returns a specific *DataStore (eg. *ThreadStore, *GroupStore, ...), and others who return a *DB. It would be nice to decide which pattern to use for uniformization, and if possible for simplicity.

Note: I would have liked to add a "default" to the switch statement in list_threads to guarantee the correctness of the parameters. But the parameters are checked before in a function, and it made the code within the default unreachable, so it didn't pass the code coverage tests.

### DIFF
--- a/testhelpers/feature_context.go
+++ b/testhelpers/feature_context.go
@@ -37,7 +37,6 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^there is a group (@\w+)$`, ctx.ThereIsAGroup)
 	s.Step(`^I am a member of the group (@\w+)$`, ctx.IAmAMemberOfTheGroup)
 	s.Step(`^I am a member of the group with id "([^"]*)"$`, ctx.IAmAMemberOfTheGroupWithID)
-	s.Step(`^there are the following group members:$`, ctx.ThereAreTheFollowingGroupMembers)
 	s.Step(`^(@\w+) is a member of the group (@\w+)$`, ctx.UserIsAMemberOfTheGroup)
 	s.Step(
 		`^(@\w+) is a member of the group (@\w+) who has approved access to his personal info$`,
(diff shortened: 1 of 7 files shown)


GeoffreyHuck:
> Missing ref to the issue it is fixing.

It's not finished yet, still requires the parameter `item_id` and the sorting/filter/pagination.

testhelpers/feature_context.go
```diff
@@ -37,11 +37,13 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^there is a group (@\w+)$`, ctx.ThereIsAGroup)
 	s.Step(`^I am a member of the group (@\w+)$`, ctx.IAmAMemberOfTheGroup)
 	s.Step(`^I am a member of the group with id "([^"]*)"$`, ctx.IAmAMemberOfTheGroupWithID)
+	s.Step(`^there are the following group members:$`, ctx.ThereAreTheFollowingGroupMembers)
```
smadbe, line 40: I don't see this rule used anywhere

smadbe (approved):
looks good

testhelpers/feature_context.go
GeoffreyHuck: fixed, indeed I removed it

PR #944, opened 2023-04-11
Add make validate-swagger to validate swagger annotations

GeoffreyHuck:
Adds:
`make validate-swagger`

To avoid going in the README to copy-paste the generate and the validate commands every time.

It would be even nicer to have a pre-commit rule, but running the command takes ~30s. We would have to fine-tune it a little more to make sure it runs only if required.


smadbe (approved):
Possibly the CI rule of building for testing could be replaced by that... but not that important.

PR #946, opened 2023-04-17
Get item service returns item subtitle and description even when can_view=info

GeoffreyHuck:
Remove the filter that was currently used.


smadbe (approved):
looks good 👍

Issue #931, opened 2023-03-23
bestAnswerGet: allow the user to know if there are no answer (>< other errors)


smadbe:
## Motivations

`bestAnswerGet` currently returns a 403 not only if one of the 3 preconditions fails as the spec claims but also if the participant has no answer for the given item.  

The service should provide a way for the frontend to identify such a case. (returning 404 instead?)

## Subtasks
* provide a way for the frontend to identify such a case
* update the spec so that this case is documented

PR #947, opened 2023-04-17
bestAnswerGet returns a 404 if permissions are ok but there is no answer

GeoffreyHuck:
Implements the 404 when there is no answer.

Those little queries: CanWatchItemAnswer, CanWatchGroupMembers, etc. seem to appear a lot and are probably executed many times during a user's session. We can consider caching them if we ever find they have a performance impact.

### DIFF
--- a/app/api/answers/get_best.go
+++ b/app/api/answers/get_best.go
@@ -1,6 +1,7 @@
 package answers
 
 import (
+	"errors"
 	"net/http"
 
 	"github.com/go-chi/render"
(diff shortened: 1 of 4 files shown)


GeoffreyHuck:
fixes #931

app/api/answers/get_best.go
```diff
@@ -11,48 +12,50 @@ import (
...
+//		"200":
+//			"$ref": "#/responses/itemAnswerGetResponse"
+//		"400":
+//			"$ref": "#/responses/badRequestResponse"
+//		"401":
+//			"$ref": "#/responses/unauthorizedResponse"
+//		"403":
+//			"$ref": "#/responses/forbiddenResponse"
+//		"500":
+//			"$ref": "#/responses/internalErrorResponse"
```
smadbe, line 51: 404 missing

app/api/answers/get_best.go
```diff
@@ -62,35 +65,25 @@ func (srv *Service) getBestAnswer(rw http.ResponseWriter, httpReq *http.Request)
...
 	} else {
 		// check 'can_view'>='content' permission on the answers.item_id
-		itemPerms := store.Permissions().MatchingUserAncestors(user).
-			WherePermissionIsAtLeast("view", "content").
-			Where("permissions.item_id = answers.item_id").
-			Select("1").
-			Limit(1)
+		if !user.CanViewItemContent(store, itemID) {
+			return service.InsufficientAccessRightsError
+		}
```
smadbe, line 84: By https://france-ioi.github.io/algorea-devdoc/design/access-rights/items/#changing-and-propagating-permissions, a user cannot have `can_watch>=answer` without having `can_view ≥ content` so this check is useless. 
(anyway, this kind of checks should have been done in one single sql request)

smadbe (changes requested):

app/api/answers/get_best.go
GeoffreyHuck: fixed

app/api/answers/get_best.go
GeoffreyHuck: That good to know! In this case though the can_view>= is checked independently in else {} block. Is this comment a mistake or am I missing something?

app/api/answers/get_best.go
```diff
@@ -11,48 +12,52 @@ import (
...
+//		"200":
+//			"$ref": "#/responses/itemAnswerGetResponse"
+//		"400":
+//			"$ref": "#/responses/badRequestResponse"
+//		"401":
+//			"$ref": "#/responses/unauthorizedResponse"
+//		"403":
+//			"$ref": "#/responses/forbiddenResponse"
+//		"404":
+//			"$ref": "#/responses/forbiddenResponse"
```
smadbe, line 51: it is not forbidden response I assume...

smadbe (changes requested):
🤷‍♂️

app/api/answers/get_best.go
GeoffreyHuck: oops, fixed

smadbe (approved):

PR #948, opened 2023-04-19
List Thread Service: item_id parameter

GeoffreyHuck:
Related to issue #888 

Handles the `item_id` parameter.

A few remarks:

1) The test have a new column in the definition of `threads`: `visible_by_participant`. The reasoning is the following:

In order to test for the `item_id` filter, we need that the defined threads are visible to the user. There are more than one way for a thread to be visible for a user, depending on the options that are used (`watched_id`, `is_mine`). And even then, the rules contains many conditions. Those ways were all already tested by the previous tests.

If we were to add manually the permissions for the threads to be visible again, we'll have a redundancy in the tests. And if one day we change the rules by which the threads are visible, we'll need to change all the tests that are unrelated to this change.

So we define a notion that the thread is visible in a column, and the test engine makes whatever necessary to make it visible. If we ever need to change how, we'll just have one update to do. We can't be sure now that `visible_by_participant` is the right approach, but all the following tests for the other filters, sorting and pagination will have to use the same method, which might change.

This issue of redundancy where every time we redefine some notions with their low level correspondence is present in many other tests in the app. It makes the whole architecture very rigid. It would be nice to reach a level where each low level notion is tested only once, and then for all tests that uses those same notions, a higher level correspondence is used.

2) The new tests automatically fills the tables `permissions_generated`, `items_ancestors` and `groups_ancestors` in the same way that is used within the app. So we only have to define the contents of `items_items`, `groups_groups` and `permissions_granted`.

This automatic filling have been disabled for older tests because they were breaking. Some of those tests define tables with inconsistencies (eg. defining both `permissions_granted` and `permissions_generated` with different values like in `update_item.feature`: https://github.com/France-ioi/AlgoreaBackend/blob/getThreadsItemID/app/api/items/update_item.feature#L25). Then, those tests also check that the content of ill-defined tables didn't change (eg. `And the table "items_ancestors" should stay unchanged`). 

This point is actually linked to the first one. It probably happened because of the lack of higher-level definitions. Everything was defined at the lowest level, then one day the low level details changed, but all the tests weren't updated. Those tests didn't break but are now in an inconsistent state. It is something to keep in mind because it'll make further refactoring and/or decoupling harder.

### DIFF
--- a/app/api/threads/list_threads.feature
+++ b/app/api/threads/list_threads.feature
@@ -195,3 +195,37 @@ Feature: List threads
       | @B_UniversityMember_HasValidated5   |
       | @B_UniversityMember_CanWatchAnswer4 |
       | @B_UniversityMember_CanWatchAnswer5 |
+
+    Scenario: Should return only thread from item or descendant when item_id is given
+      Given I am @John
+      And there are the following items:
+        | item              | parent      | type    |
+        | @Root_Task        |             | Task    |
+        | @Chapter1         |             | Chapter |
+        | @Chapter1_Task    | @Chapter1   | Task    |
+        | @Chapter2         |             | Chapter |
+        | @Chapter2_Task    | @Chapter2   | Task    |
+        | @Chapter2_1       | @Chapter2   | Chapter |
+        | @Chapter2_1_Task1 | @Chapter2_1 | Task    |
+        | @Chapter2_1_Task2 | @Chapter2_1 | Task    |
+        | @Chapter3         |             | Chapter |
+      And there are the following threads:
+        | participant | item              | visible_by_participant | message_count |
+        | @John       | @Root_Task        | 1                      | 100           |
+        | @John       | @Chapter1         | 1                      | 101           |
+        | @John       | @Chapter1_Task    | 1                      | 102           |
+        | @John       | @Chapter2         | 1                      | 103           |
+        | @John       | @Chapter2_Task    | 1                      | 104           |
+        | @John       | @Chapter3         | 1                      | 105           |
+        | @John       | @Chapter2_1       | 1                      | 106           |
+        | @John       | @Chapter2_1_Task1 | 1                      | 107           |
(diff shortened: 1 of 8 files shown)


app/api/threads/list_threads.feature
```diff
@@ -195,3 +195,37 @@ Feature: List threads
...
+        | @John       | @Chapter2_1_Task2 | 1                      | 108           |
+      And I am @John
+      When I send a GET request to "/threads?is_mine=1&item_id=@Chapter2"
+      Then the response code should be 200
+      And the response at $[*].item.id should be:
+        | @Chapter2         |
+        | @Chapter2_Task    |
+        | @Chapter2_1       |
+        | @Chapter2_1_Task1 |
+        | @Chapter2_1_Task2 |
```
smadbe, line 231: ok, clear 👍

testhelpers/steps_app_language.go
```diff
@@ -852,9 +925,9 @@ func (ctx *TestContext) IAmPartOfTheHelperGroupOfTheThread() error {
 // to a given item.
 func (ctx *TestContext) ICanRequestHelpToTheGroupWithIDOnTheItemWithID(group, item string) error {
 	ctx.addPermissionGranted(
-		ctx.user,
 		ctx.user,
 		item,
+		"can_request_help_to",
```
smadbe, line 930: this file is becoming very long.. probably should be split during a future change...

smadbe (approved):
👍 
For your 1, I agree with the point of redundancy. Just pay attention not to add too much abstraction or at a moment it will become not really understandable without having doubt... 
It is ok for now if the old tests stay as they are for now... the next time these services will change (one at a time) it will still be time to change the tests if it looks useful.

testhelpers/steps_app_language.go
GeoffreyHuck: Yes, I was thinking about it too. It'll be split soon.

PR #950, opened 2023-04-24
List threads: order, pagination and filters

GeoffreyHuck:
#fixes 888

The tests for sorting, pagination and filters are as simple as possible. Hopefully without sacrificing reliability.

Only `latest_update_at` has been added for sorting. But we can add more parameters.

For paging, `from.item_id` and `form.participant_id` have been added, those are the primary key of threads.

If `status` filter is given with an invalid value, the service will return nothing because nothing matches. We could also check that the value is a valid status.

The issue specified format ISO8601 for date-time. RFC3339 (eg. "2006-01-02T15:04:05Z07:00") seems to be equivalent and there already was a function ResolveURLQueryGetTimeField using it.

This is the last PR for the listThreads feature.

### DIFF
--- a/app/api/threads/list_threads.feature
+++ b/app/api/threads/list_threads.feature
@@ -159,7 +159,7 @@ Feature: List threads
           "status": "waiting_for_trainer"
         }
       ]
-    """
+      """
 
   Scenario: Should get the threads whose the participant is a descendant of the watched_group_id
     Given I am @A_UniversityManagerCanWatch
@@ -266,7 +266,7 @@ Feature: List threads
     And there are the following threads:
       | participant | item             | visible_by_participant | latest_update_at    |
       | @John       | @TaskMinUpdateAt | 1                      | 2023-01-01 00:00:01 |
-      | @John       | @TaskMaxUpdateAt | 1                      | 2023-01-01 00:00:02  |
+      | @John       | @TaskMaxUpdateAt | 1                      | 2023-01-01 00:00:02 |
     And I am @John
     When I send a GET request to "/threads?is_mine=1&limit=1&sort=latest_update_at&from.item_id=<from.item_id>&from.participant_id=<from.participant_id>"
     Then the response code should be 200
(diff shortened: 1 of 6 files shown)


app/api/threads/list_threads.feature
```diff
@@ -122,25 +122,8 @@ Feature: List threads
     When I send a GET request to "/threads?watched_group_id=@Laboratory"
     Then the response code should be 200
     And the response body should be, in JSON:
-    """
+      """
```
smadbe, line 125: (I may be wrong) does not seem aligned with the closing one

app/api/threads/list_threads.go
```diff
@@ -70,7 +84,41 @@ type participant struct {
...
+//			in: query
+//			type: string
+//			format: date-time
+//		- name: sort
+//			in: query
+//			default: [-latest_update_at]
+//			type: array
+//			items:
+//				type: string
+//				enum: [latest_update_at,-latest_update_at]
```
smadbe, line 103: Sort must always ends with tie-breaker (basically an unique key for the listed entity), which should be part of the "default" sort, be part of the enum, and automatically added to the sort if not present.
Otherwise, the output is deterministic.

smadbe (changes requested):

app/api/threads/list_threads.feature
GeoffreyHuck: fixed

app/api/threads/list_threads.go
GeoffreyHuck: After checking, the pagination system it is already appending the "TieBreakers" parameters to the default sort:
`ORDER BY threads.latest_update_at DESC, items.id ASC, threads.participant_id ASC LIMIT 500`

It is added if no sort parameter is used (default), or when `&sort=latest_update_at` is used.

The only thing done was adding the TieBreakers parameter here:
```
	queryDB, apiError = service.ApplySortingAndPaging(r, queryDB, &service.SortingAndPagingParameters{
		Fields: service.SortingAndPagingFields{
			"latest_update_at": {ColumnName: "threads.latest_update_at"},
			"item_id":          {ColumnName: "items.id"},
			"participant_id":   {ColumnName: "threads.participant_id"},
		},
		DefaultRules: "-latest_update_at",
		TieBreakers: service.SortingAndPagingTieBreakers{
			"item_id":        service.FieldTypeInt64,
			"participant_id": service.FieldTypeInt64,
		},
	})
```

Is it still necessary to add item_id and participant_id in the sort paramter?

smadbe (commented):
> Is it still necessary to add item_id and participant_id in the sort paramter?

My remark was more on the doc. The current doc states that the allowed doc is only `[latest_update_at,-latest_update_at]` while at least `item_id` and `participant_id` are sortable.... and the default is not only `[-latest_update_at]`.  

So just fix the doc.

app/api/threads/list_threads.go
GeoffreyHuck: fixed the doc

smadbe (approved):

PR #951, opened 2023-04-25
ParticipantProgress: return children if current user has started result on item only

### LINKED ISSUE #915: Fix participant-progress service
## Motivations

The backend is supposed to return children of a chapter only if 2 conditions are met: the current user has `can_view>=content` on it and it has a (started) result.

The current participant-progress implementation does not implement the second rule as it return children even when the chapter has no result. 

## Subtasks
* do not return children in "participant-progress" when the current-user does not have started result on the item (including in observation mode)
* the "progress" over the item itself should still be returned (as long as it is visible to the current user)

GeoffreyHuck:
fixes #915 

Added a test with `as_team_id` and `watched_group_id`. Existing tests have been updated to add a started result if it was expected that they return children.

Also refactor common little queries that were written in the service.

### DIFF
--- a/app/api/groups/get_participant_progress.go
+++ b/app/api/groups/get_participant_progress.go
@@ -66,7 +66,7 @@ type groupParticipantProgressResponse struct {
 	// required: true
 	Item groupParticipantProgressResponseCommon `json:"item"`
 	// required: true
-	Children []groupParticipantProgressResponseChild `json:"children"`
+	Children *[]groupParticipantProgressResponseChild `json:"children,omitempty"`
 }
 
 type rawParticipantProgressRaw struct {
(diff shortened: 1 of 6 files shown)


app/api/groups/get_participant_progress.go
```diff
@@ -91,73 +91,84 @@ type rawParticipantProgressRaw struct {
...
-//		summary: Get progress of a participant
-//		description: >
-//	             Returns the current progress of a participant on a given item.
+//	---
+//	summary: Get progress of a participant
+//	description: >
+//		Returns the current progress of a participant on a given item.
+//
 //
+//		For `{item_id}` and all its visible children,	displays the results of the given participant
```
smadbe, line 110: tabs in the middle of the text

app/api/groups/get_participant_progress.go
```diff
@@ -91,73 +91,84 @@ type rawParticipantProgressRaw struct {
...
+//		(current user or `as_team_id` (if given) or `watched_group_id` (if given)).
 //
-//	             For `{item_id}` and all its visible children,
-//	             displays the results of the given participant
-//	             (current user or `as_team_id` (if given) or `watched_group_id` (if given)).
-//	             Only one of `as_team_id` and `watched_group_id` can be given.
-//	             The results are sorted by `items_items.child_order`.
+//		The children are returned only if the current user has a started result on the given `item_id`.
```
smadbe, line 113: Have you changed the doc to reflect this? I would expect no "children" key in the output when not allowed. (to be done if it does not complicate things too much / if it does not add a new sql request)

smadbe (approved):

app/api/groups/get_participant_progress.go
GeoffreyHuck: fixed, now the "children" key appears only if the current user have a started result on the item_id, tested with `And the response should not be defined at $.children`. There's also a test with "children": [], when there is no children but the current user has a started result.

PR #952, opened 2023-04-25
GetThread now returns a token attribute

### LINKED ISSUE #890: Forum: add authorization token in thread service
## Motivations

As the actual forum content (messages, events) are handled in a different service, we need to provide this service with a way to validate the access the user is requesting on the thread.

The token is not related at all with the task token (hint token, answer token, ...) which are used to communicate with tasks, but similar mechanism may be used. Basically, the idea is to provide a JWT which contain a set of key-value and to sign these with the backend asymmetric private key so that it can be verified by any other service.

## Subtasks

- [ ] In the GET thread service which has been developed in #884, add a "token" attribute
- [ ] The token should contain, in its payload: 
   - `item_id`
   - `participant_id`
   - `user_id`: the current user id
   - `is_mine` (bool) : whether the thread is for the current-user... for now just `user_id == participant_id`
   - `can_watch` (bool): whether the current user can watch the participant on this item (i.e. `can_watch>=answer` permission on the item AND `can_watch_members` on the participant)
   - `can_write` (bool): whether the user can post new content on the thread ([writing in a thread](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/#writing-in-a-thread))
- [ ] The token should be signed by the backend

GeoffreyHuck:
fixes #890 

token_test.go:generateSignedTestToken is a function that can be use when we need to generate a test token, because we need to create test tokens for the tests. I'm not sure if there is a good way to handle those things.

Note: `exp` has been added in the Thread token, but the system that **decrypt** the tokens doesn't support `exp`. The check that is made verify that we are between one day before and one day after the generation of the token: https://github.com/France-ioi/AlgoreaBackend/blob/master/app/token/token.go#L107 . Maybe that's not needed (does the backend need to decrypt thread token?), but we're currently not following the specification.

There is a lot of duplication in the token system, many methods that look very similar are duplicated for each type of token. Furthermore, all tokens are defined in the same files regardless of their domain. It would be nice to refactor some of this the next time we work on token. For now, I'll write some documentation on what I found.

### DIFF
--- a/app/api/threads/get_thread.feature
+++ b/app/api/threads/get_thread.feature
@@ -51,7 +51,7 @@ Feature: Get thread
     Given I am the user with id "1"
     When I send a GET request to "/items/21/participant/1/thread"
     Then the response code should be 200
-    And the following token "threadToken" signed by the app is distributed:
+    And "threadToken" is a token signed by the app with the following payload:
       """
         {
           "can_watch": false,
--- a/app/api/answers/submit.feature
+++ b/app/api/answers/submit.feature
@@ -31,7 +31,7 @@ Feature: Submit a new answer
   Scenario: User is able to submit a new answer
     Given I am the user with id "101"
     And time is frozen
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idUser": "101",
@@ -83,7 +83,7 @@ Feature: Submit a new answer
   Scenario: User is able to submit a new answer (with all fields filled in the token)
     Given I am the user with id "101"
     And time is frozen
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idItem": "50",
--- a/app/api/answers/submit.robustness.feature
+++ b/app/api/answers/submit.robustness.feature
@@ -59,7 +59,7 @@ Feature: Submit a new answer - robustness
 
   Scenario: Missing answer
     Given I am the user with id "101"
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idUser": "101",
@@ -80,7 +80,7 @@ Feature: Submit a new answer - robustness
 
   Scenario: Wrong idUser
     Given I am the user with id "101"
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idUser": "",
@@ -102,7 +102,7 @@ Feature: Submit a new answer - robustness
 
   Scenario: Wrong idItemLocal
     Given I am the user with id "101"
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idUser": "101",
@@ -123,7 +123,7 @@ Feature: Submit a new answer - robustness
 
   Scenario: Wrong idAttempt
     Given I am the user with id "101"
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idUser": "101",
@@ -145,7 +145,7 @@ Feature: Submit a new answer - robustness
 
   Scenario: idUser doesn't match the user's group id
     Given I am the user with id "101"
-    And the following token "userTaskToken" signed by the app is distributed:
+    And "userTaskToken" is a token signed by the app with the following payload:
       """
       {
         "idUser": "20",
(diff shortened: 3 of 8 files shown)


smadbe:
> does the backend need to decrypt thread token?

Not this one at least.

app/api/threads/get_thread.feature
```diff
@@ -47,57 +47,50 @@ Feature: Get thread
...
+        "can_write": true,
+        "date": "20-01-2020",
+        "exp": "1579485600",
+        "is_mine": true,
+        "item_id": "21",
+        "participant_id": "1",
+        "user_id": "1"
+      }
     }
     """
```
smadbe, line 71: As discussed previously, gherkin tests should be clear enough so that the reader knows how to interpret these without checking the rules. Here, I do not really know how to interpret that without reading the "ThreadTokenResponse" decoder (which is not very "readable").

Actually, as a reader, I would expect something like:
```
    """
    {
      "participant_id": 1,
      "item_id": 10,
      "status": "not_started"
      "item_id": 21,
      "status": "waiting_for_trainer",
      "token": $atoken
    }
    """
  And $atoken is a valid "ThreadToken" with in its payload:
   """
     {
        "can_watch": false,
        "can_write": true,
        "date": "20-01-2020",
        "exp": "1579485600",
        "is_mine": true,
        "item_id": "21",
        "participant_id": "1",
        "user_id": "1"
      }
   """
```
(or "$.token of the response is.... " and/or "is a valid JWT signed by the backend". 

Because the "token" is quite opaque for the frontend, which is mainly interested in knowing the attributes in the response, but not the token that it will just pass as a string.  The "AlgoreaForum" backend service which will receives the token is only interested by what it is supposed to find in the token. The format that you describe here is something completely fake as there is no "can_watch" key in the same "structure" as the `status` key.

Of course, that's just an opinion, let me know.

GeoffreyHuck: I tried to use a wildcard <*> to replace the token but it makes the comparison with the actual response very difficult to do. I didn't even try: it involves replacing <*> with what we get in the actual response.

The easiest way is to define the token first, and replace it in the response. There's already an existing Gherkin feature for it.

It would be difficult to prefix the variable by $ because it would conflict with the reference system.

This is what I came up with:

```
  Scenario: Should return all fields when the thread exists
    Given I am the user with id "1"
    When I send a GET request to "/items/21/participant/1/thread"
    And the following token "threadToken" signed by the app is distributed:
      """
        {
          "can_watch": false,
          "can_write": true,
          "date": "20-01-2020",
          "exp": "1579485600",
          "is_mine": true,
          "item_id": "21",
          "participant_id": "1",
          "user_id": "1"
        }
      """
      """
        {
          "participant_id": 1,
          "item_id": 21,
          "status": "waiting_for_trainer",
          "token": "{{threadToken}}"
        }
      """
```

smadbe:
> Note: `exp` has been added in the Thread token, but the system that **decrypt** the tokens doesn't support `exp`. The check that is made verify that we are between one day before and one day after the generation of the token: https://github.com/France-ioi/AlgoreaBackend/blob/master/app/token/token.go#L107 . Maybe that's not needed (does the backend need to decrypt thread token?), but we're currently not following the specification.

We are not at all in the same scenario as the other token, that's a different kind of token... so we cannot reuse some part but no need to copy weird stuff. In our case here anyway, the backend does not need to decode the token, right, and as the spec claims we need an expiry of now+2h. Is it done? What is the `date` for in the token ? That does not seem to apply here

GeoffreyHuck:
The `date` is automatically added by the token signing process, right before encryption. This can be changed if we want to remove it, but it might not be very easy to do it in a clean way.

`exp` is implemented as now+2h, the format is the number of seconds since EPOCH (1970), as in the RFC: https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4

smadbe:
> The `date` is automatically added by the token signing process, right before encryption. I can try to change this if needed.

Not really "needed" to remove it as I can ignore it but a date (without the time) is clearly useless in most cases except probably the case it was added for... probably a hack with the login-module.

> `exp` is implemented as now+2h, the format is the number of seconds since EPOCH (1970), as in the RFC: https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4

👍

app/api/threads/get_thread.feature
```diff
@@ -47,57 +47,54 @@ Feature: Get thread
       | 50      | 3              | closed                  | 20              | 2020-01-10 00:00:00 |
     And the time now is "2020-01-20T00:00:00Z"
 
-  Scenario: Status should be not_started if the thread doesn't exists
+  Scenario: Should return all fields when the thread exists
     Given I am the user with id "1"
-    When I send a GET request to "/items/10/participant/1/thread"
+    When I send a GET request to "/items/21/participant/1/thread"
     Then the response code should be 200
+    And the following token "threadToken" signed by the app is distributed:
```
smadbe, line 54: Detail: the word "distributed" is a bit strange here... actually I assume you just mean that the treadToken is set to `sign(content)` and can be used at the next payload check. Maybe just "computed". Or even `And "threadToken" is a token signed by the app with the following payload:`.

smadbe (approved):

app/api/threads/get_thread.feature
GeoffreyHuck: fixed

PR #954, opened 2023-05-03
Do not allow unsigned requests for Hint request service

### LINKED ISSUE #824: Hint request: do not allow unsigned requests
The request for hint does not have to support unsigned (as json) requests. If the task has no public key for its platform, we can just consider this service cannot be called.

Please update the implementation and the doc (and the tests?) accordingly.

-------

Initial issue: 

As Thomas reported in https://github.com/France-ioi/AlgoreaFrontend/pull/856, I think we have a problem with the format of the hint tokens for the [hint token service](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemGetHintToken).

This service takes as an arg a task token and a hint token. If we send both as JWTs, we get a `"Invalid hint_requested: json: cannot unmarshal string into Go value of type map[string]formdata.Anything"` error from the backend. However, if the hint token is decoded and sent a json, it works. 

I suppose this is not the expected behavior of the service. Let me know if there is a misunderstanding from my side.

GeoffreyHuck:
fixes #824 

- Little refactor in the test file with the item url to make it easier to read.
- Added tests to reject when no public key exists  for the platform, and another for when the hint parameters are sent in clear JSON instead of a JWT
- UnmarshalDependingOnItemPlatform has been refactored because it was very weirdly written: it used an internal "panic+recover" with a custom type of error.


smadbe (changes requested):
The test coverage is not good anymore, I let you fix it first.

GeoffreyHuck:
fixed test coverage and taken into account the discussions at https://github.com/France-ioi/AlgoreaBackend/issues/824

PR #956, opened 2023-05-03
Fix nil pointer dereference when update group root_skill_id to null

### LINKED ISSUE #860: Runtime error: invalid memory address or nil pointer dereference
If I call, (the group has a root skill assigned when I call this, the problem seems to be related with setting the root skill to null, was 239326154416624928 before)
```
curl 'https://dev.algorea.org/api/groups/3192699464035117100' -X PUT -H 'Authorization: Bearer ***' --data-raw '{"name":"RootItemsTest","description":null,"root_activity_id":null,"root_skill_id":null}'
```
I get a 
```
{"success":false,"message":"Internal Server Error","error_text":"Runtime error: invalid memory address or nil pointer dereference"}
```

GeoffreyHuck:
fixes #860 

Added a test to reproduce the issue.

Was due to the fact that a nil interface{} is != nil in go when the type of the interface is not null. Variable names are renamed to be explicit about it.

### DIFF
--- a/app/api/groups/update_group.feature
+++ b/app/api/groups/update_group.feature
@@ -323,6 +323,38 @@ Feature: Update a group (groupEdit)
     And the table "group_pending_requests" should stay unchanged
     And the table "group_membership_changes" should stay unchanged
 
+  Scenario: Should be able to update the description, root_activity_id and root_skill_id from a value to null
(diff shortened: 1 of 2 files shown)


smadbe (approved):
great! :+1:

Issue #850, opened 2022-09-22
403 on path-from-root while there is a path


smadbe:
## Motivations
Fix problem described below (see "Issue")

## Subtasks
Update the itemPathFromRootFind service so that
* the rule about chain of attempts becomes: " The chain of attempts of the path cannot have missing results for items requiring explicit entry FOR ANY PATH ELEMENT EXCEPT FOR THE LAST ONE or not started results
within or below ended/not-allowing-submissions attempts. "
* fix the implementation to reflect this change

# Issue

The "Find an item path" service state "The path consists only of the items visible to the participant (can_view>='content' for all the items except for the last one and can_view>='info' for the last one)".

However, as a temp user:
```
curl 'https://dev.algorea.org/api/items/4297212147988794347/path-from-root' -H 'Cookie: access_token=3!pru7gasyd7a2zb9hpalc185pl84w73qh!dev.algorea.org!/api/' 
{"success":false,"message":"Forbidden","error_text":"Insufficient access rights"}
```
While:
```
$ curl 'https://dev.algorea.org/api/items/4702/1625159049301502151/4297212147988794347/breadcrumbs?parent_attempt_id=0'  -H 'Cookie: access_token=3!pru7gasyd7a2zb9hpalc185pl84w73qh!dev.algorea.org!/api/' 
[{"attempt_id":"0","item_id":"4702","language_tag":"fr","title":"Parcours officiels","type":"Chapter"},{"attempt_id":"0","item_id":"1625159049301502151","language_tag":"fr","title":"MOTIF ART","type":"Chapter"},{"item_id":"4297212147988794347","language_tag":"en","title":"Sur le fil","type":"Task"}]
```
and perm on `4702`, `1625159049301502151` and `4297212147988794347` are respectively `content`, `content` and`info`

GeoffreyHuck:
Two questions about how this change should  be reflected in `resultStartPath`?

From my understanding, `resultStartPath` gets for each item in the path:
- An `attempt_id`
- Whether it has a started result: `has_started_result`. If it doesn't, a result is added in the DB and a results propagation is done.

* With the change, the last `item` can have `requires_explicit_entry=1`, so it's possible there is no `attempt`. What should we do in this case?

* Also, the response of `resultStartPath` contains an `attempt_id`. It is the `attempt_id` associated with the last item in the path. What do we return in this case?

smadbe:
An important aspect I've just realized rechecking all doc and the frontend implem:
[...]

PR #961, opened 2023-05-16
403 on path-from-root while there is a path

GeoffreyHuck:
Related to #850

The queries were too difficult to read in their current form, there were adapted in the following way to make it easier (in the first commit):
- Following the formatting style guide here https://www.sqlstyle.guide/
- All the WITH declarations have been grouped at the same level, instead of having them nested

Given the response here: https://github.com/France-ioi/AlgoreaBackend/issues/850#issuecomment-1584288727 . `resultStartPath` hasn't been changed.

New tests have been added in `path_from_root_integration_test.go`:
- The item with id `23` is a contest sub-item. It has been added as a child of `22` (22 has `require_explicit_entry=1`)
- For all the previous tests that were asking a path with `22`:
* If they **returned no path**, because `22` has `require_explicit_entry=1` and didn't pass the other conditions, they now **PASS** because `22` is the last element of the path and doesn't require those other conditions anymore
* If they **returned a path**, because `22` was entered and has the other conditions were fulfilled, we now ask for `23`, so that `22` is a non-last element, and thus requires those conditions to be fulfilled.

### DIFF
--- a/app/api/items/path_from_root.go
+++ b/app/api/items/path_from_root.go
@@ -100,53 +100,92 @@ func findItemPath(store *database.DataStore, participantID, itemID int64) []stri
 	canViewContentIndex := store.PermissionsGranted().ViewIndexByName("content")
 
 	var pathStrings []string
-	service.MustNotBeError(store.Raw(`
-			WITH RECURSIVE paths (path, last_item_id, last_attempt_id, score, attempts, is_active) AS (
-				WITH groups_with_root_items AS ?,
-					visible_items AS ?,
-					root_items AS (
-						SELECT visible_items.id AS id FROM groups_with_root_items JOIN visible_items ON visible_items.id = root_activity_id
-						UNION
-						SELECT visible_items.id FROM groups_with_root_items JOIN visible_items ON visible_items.id = root_skill_id),
-					item_ancestors AS (
-						SELECT visible_items.id, requires_explicit_entry, can_view_generated_value
-						FROM items_ancestors
-						JOIN visible_items ON visible_items.id = items_ancestors.ancestor_item_id WHERE child_item_id = ?
-						UNION
-						SELECT id, requires_explicit_entry, can_view_generated_value FROM visible_items WHERE id = ?),
-					root_ancestors AS (
-						SELECT item_ancestors.id, requires_explicit_entry, can_view_generated_value
-						FROM item_ancestors
-						JOIN root_items ON root_items.id = item_ancestors.id)
-				(SELECT CAST(root_ancestors.id AS CHAR(1024)), root_ancestors.id, attempts.id, results.started_at IS NULL,
-				        CAST(LPAD(attempts.id, 20, 0) AS CHAR(1024)), attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until
-				FROM root_ancestors
-				JOIN attempts ON attempts.participant_id = ? AND
-					(NOT root_ancestors.requires_explicit_entry OR attempts.root_item_id = root_ancestors.id)
-				LEFT JOIN results ON results.participant_id = attempts.participant_id AND
-					attempts.id = results.attempt_id AND results.item_id = root_ancestors.id
-				WHERE (root_ancestors.id = ? OR root_ancestors.can_view_generated_value >= ?) AND
-					(NOT root_ancestors.requires_explicit_entry OR results.attempt_id IS NOT NULL) AND
-					(results.started_at IS NOT NULL OR attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until) AND
-					(results.attempt_id IS NOT NULL OR attempts.id = 0))
-				UNION
-				(SELECT CONCAT(paths.path, '/', item_ancestors.id), item_ancestors.id, attempts.id, (paths.score << 1) + (results.started_at IS NULL),
-				        CONCAT(paths.attempts, '/', LPAD(attempts.id, 20, 0)),
-				        paths.is_active AND attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until
-				FROM paths
-				JOIN items_items ON items_items.parent_item_id = paths.last_item_id
-				JOIN item_ancestors ON item_ancestors.id = items_items.child_item_id
-				JOIN attempts ON attempts.participant_id = ? AND
-					(NOT item_ancestors.requires_explicit_entry OR attempts.root_item_id = item_ancestors.id) AND
-					IF(attempts.root_item_id = item_ancestors.id, attempts.parent_attempt_id, attempts.id) = paths.last_attempt_id
-				LEFT JOIN results ON results.participant_id = attempts.participant_id AND
-						attempts.id = results.attempt_id AND results.item_id = item_ancestors.id
-				WHERE paths.last_item_id <> ? AND (item_ancestors.id = ? OR item_ancestors.can_view_generated_value >= ?) AND
-					(NOT item_ancestors.requires_explicit_entry OR results.attempt_id IS NOT NULL) AND
-					(results.started_at IS NOT NULL OR attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until AND paths.is_active)))
-			SELECT path FROM paths WHERE paths.last_item_id = ? ORDER BY score, attempts DESC LIMIT 1`,
-		groupsWithRootItems.SubQuery(), visibleItems.SubQuery(), itemID, itemID, participantID, itemID, canViewContentIndex,
-		participantID, itemID, itemID, canViewContentIndex, itemID).
+	service.MustNotBeError(store.Raw(
+		`
+			WITH RECURSIVE
+				groups_with_root_items AS ?,
+				visible_items AS ?,
+				root_items AS (
+					(SELECT visible_items.id AS id
+						 FROM groups_with_root_items
+									JOIN visible_items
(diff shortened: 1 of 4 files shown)


GeoffreyHuck:
I tried to refactor the query of `itemBreadcrumbsFromRootsGet` and `itemPathFromRootFind`. The problem is there's not just a difference in the fields that are returned, but also in the rules to what is a path. I don't think we can just factorize those two requests without make it very ugly, so for now I reverted this.

This is the result I got (`itemPathFromRootFind=true` for `itemPathFromRootFind`:
`func FindItemPathStrings(itemPathFromRootFind bool, store *database.DataStore, userID, participantID, itemID int64) []string {
	participantAncestors := store.ActiveGroupAncestors().Where("child_group_id = ?", participantID).
		Joins("JOIN `groups` ON groups.id = groups_ancestors_active.ancestor_group_id").
		Select("groups.id, root_activity_id, root_skill_id")
	groupsManagedByParticipant := store.ActiveGroupAncestors().ManagedByGroup(userID).
		Joins("JOIN `groups` ON groups.id = groups_ancestors_active.child_group_id").
		Select("groups.id, root_activity_id, root_skill_id")
	groupsWithRootItems := participantAncestors.Union(groupsManagedByParticipant.SubQuery())

	visibleItems := store.Permissions().MatchingGroupAncestors(userID).
		WherePermissionIsAtLeast("view", "info").
		Joins("JOIN items ON items.id = permissions.item_id").
		Select("items.id, requires_explicit_entry, MAX(can_view_generated_value) AS can_view_generated_value").
		Group("items.id")

	canViewContentIndex := store.PermissionsGranted().ViewIndexByName("content")

	query := ""
	if itemPathFromRootFind {
		query += "WITH RECURSIVE paths (path, last_item_id, last_attempt_id, score, attempts, is_active) AS ("
	} else {
		query += "WITH RECURSIVE paths (path, last_item_id, last_attempt_id) AS ("
	}

	query += `
			WITH groups_with_root_items AS ?,
				visible_items AS ?,
				root_items AS (
					SELECT visible_items.id AS id FROM groups_with_root_items JOIN visible_items ON visible_items.id = root_activity_id
					UNION
					SELECT visible_items.id FROM groups_with_root_items JOIN visible_items ON visible_items.id = root_skill_id),
				item_ancestors AS (
					SELECT visible_items.id, requires_explicit_entry, can_view_generated_value
					FROM items_ancestors
					JOIN visible_items ON visible_items.id = items_ancestors.ancestor_item_id WHERE child_item_id = ?
					UNION
					SELECT id, requires_explicit_entry, can_view_generated_value FROM visible_items WHERE id = ?),
				root_ancestors AS (
					SELECT item_ancestors.id, requires_explicit_entry, can_view_generated_value
					FROM item_ancestors
					JOIN root_items ON root_items.id = item_ancestors.id)
		(SELECT CAST(root_ancestors.id AS CHAR(1024)),
					  root_ancestors.id,
						attempts.id`

	if itemPathFromRootFind {
		query += `,
							results.started_at IS NULL,
				      CAST(LPAD(attempts.id, 20, 0) AS CHAR(1024)),
							attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until
		`
	}
	query += `
				FROM root_ancestors
				JOIN attempts ON attempts.participant_id = ? AND
					(NOT root_ancestors.requires_explicit_entry OR attempts.root_item_id = root_ancestors.id)
	`

	if itemPathFromRootFind {
		query += " LEFT "
	}
	query += `JOIN results ON results.participant_id = attempts.participant_id AND
					attempts.id = results.attempt_id AND results.item_id = root_ancestors.id
				WHERE (root_ancestors.id = ? OR root_ancestors.can_view_generated_value >= ?) AND
	`

	if itemPathFromRootFind {
		query += `
			(NOT root_ancestors.requires_explicit_entry OR results.attempt_id IS NOT NULL) AND
				(results.started_at IS NOT NULL OR attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until) AND
				(results.attempt_id IS NOT NULL OR attempts.id = 0))
			UNION
			(SELECT CONCAT(paths.path, '/', item_ancestors.id), item_ancestors.id, attempts.id, (paths.score << 1) + (results.started_at IS NULL),
							CONCAT(paths.attempts, '/', LPAD(attempts.id, 20, 0)),
							paths.is_active AND attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until
		`
	} else {
		query += `
				(results.started_at IS NOT NULL))
				UNION
				(SELECT CONCAT(paths.path, '/', item_ancestors.id), item_ancestors.id, attempts.id
		`
	}

	query += `
				FROM paths
				JOIN items_items ON items_items.parent_item_id = paths.last_item_id
				JOIN item_ancestors ON item_ancestors.id = items_items.child_item_id
				JOIN attempts ON attempts.participant_id = ? AND
					(NOT item_ancestors.requires_explicit_entry OR attempts.root_item_id = item_ancestors.id) AND
					IF(attempts.root_item_id = item_ancestors.id, attempts.parent_attempt_id, attempts.id) = paths.last_attempt_id
	`

	if itemPathFromRootFind {
		query += "LEFT "
	}
	query += `JOIN results ON results.participant_id = attempts.participant_id AND
						attempts.id = results.attempt_id AND results.item_id = item_ancestors.id
				WHERE paths.last_item_id <> ? AND (item_ancestors.id = ? OR item_ancestors.can_view_generated_value >= ?) AND
	`

	if itemPathFromRootFind {
		query += `
			(NOT item_ancestors.requires_explicit_entry OR results.attempt_id IS NOT NULL) AND
					(results.started_at IS NOT NULL OR attempts.ended_at IS NULL AND NOW() < attempts.allows_submissions_until AND paths.is_active)))
		`
	} else {
		query += " (results.started_at IS NOT NULL))) "
	}

	query += " SELECT path FROM paths WHERE paths.last_item_id = ? "

	if itemPathFromRootFind {
		query += "ORDER BY score, attempts DESC LIMIT 1 "
	} else {
		query += "GROUP BY path ORDER BY path "
	}

	var pathStrings []string
	service.MustNotBeError(store.Raw(query,
		groupsWithRootItems.SubQuery(), visibleItems.SubQuery(), itemID, itemID, participantID, itemID, canViewContentIndex,
		participantID, itemID, itemID, canViewContentIndex, itemID).
		ScanIntoSlices(&pathStrings).Error())

	return pathStrings
}`

GeoffreyHuck:
When we try to read big SQL queries, it's very difficult to follow their parameters because they are not named (`?`).

I tried to use `named parameters` to correct this, but I found some issues in implementation. Here's what went wrong and a proposition to overcome those issues:

From this definition:
```
[complex query with ? ... ? ... ? ... ?],
groupsWithRootItems.SubQuery(),
visibleItems.SubQuery(),
itemID,
itemID,
participantID,
itemID,
canViewContentIndex,
participantID,
itemID,
itemID,
canViewContentIndex,
itemID,
```

To this definition:
```
[complex query with @parameter1 ... @parameter2 ... @parameter1 ... @parameter1...],
sql.Named("groupsWithRootItems", groupsWithRootItems.SubQuery()),
sql.Named("visibleItems", visibleItems.SubQuery()),
sql.Named("itemID", itemID),
sql.Named("participantID", participantID),
sql.Named("canViewContent", canViewContent),
```

Note: this also solves the issue of having many arguments present multiple times.

I encountered two issues:
- It doesn't work for GORM SubQueries. It looks like GORMv2 implemented this though, so it could solve this issue.
- Even when there is no SubQuery, I encountered `Mysql: driver does not support the use of Named Parameters`. Not sure if it would work with GORMv2. It depends if they use the sql driver implementation of if they implemented it themselves.

Proposition:
It would be nice to check if this would work with GORMv2.
Otherwise, it could be possible to implement this feature on top of GORM. Since the database module is already built on top of GORM, we could add it there, and convert named parameters into their equivalent interrogation marks when passing queries to the GORM layer.

Defining queries could become like this:
```
[complex query with @groupsWithRootItems... @itemID... @participantID... @itemID...], map[string]interface{}{
  "groupsWithRootItems": groupsWithRootItems.SubQuery(),
  "visibleItems": visibleItems.SubQuery(),
  "itemID": itemID,
  "participantID": participantID,
  "canViewContent": canViewContent,
}
```

app/api/items/path_from_root.go
```diff
@@ -17,24 +17,31 @@ import (
...
+//			* preferring paths having less missing/not-started results,
+//			* and having higher values of `attempt_id`.
 //
-//		If `as_team_id` is given, the attempts/results of the path are linked to the `as_team_id` group instead of the user's self group.
+//		For a path to be returned, each of its items must:
+//			* Either have `require_explicit_entry`=0 ,
+//			* Or if it has `require_explicit_entry=1`,
+//				then the following condition must be fulfilled, except if it is the last item of the path:
+//				the item must have at least one result with `started`=1 AND its attempt must have
+//					(`NOW()` < `attempt.ended_at` ) OR (`NOW()` < `attempt.allows_submissions_until`)).
```
smadbe, line 33: Shouldn't it be a "AND" between the 2 conditions?

smadbe (approved):
Of course, it is difficult to review the query and insure it is correct. We'll see 🙄

I hope your changes haven't broken performances, but we'll notice quite quickly.

app/api/items/path_from_root.go
GeoffreyHuck: fixed, actually it was wrong. I updated it with `(`attempt.ended_at` IS NULL) AND (`NOW()` < `attempt.allows_submissions_until`)`, which is what the query is doing.

PR #962, opened 2023-05-23
Item navigation: do not return activities as skill children

### LINKED ISSUE #959: Item navigation: do not return activities as skill children
## Motivations

In the frontend, the navigation service is used for the left menu. However, this left menu for skills only display skills, not children activities. Activities are currently filtered client-side but properties such as `has_visible_children` is still wrong, and so give "strange" behaviors.

## Subtasks
In the item navigation service: 
- if the id given as input is a skill, only return skills (if it is a chapter, returns tasks and chapter... but it is already the only possible children)
- `has_visible_children` should be computed on children skills only.

GeoffreyHuck:
fixes #959 

- Added new test case with a navigation request on a Skill

- Updated Gherkin helper `TheResponseAtShouldBe` to handle comparisons on boolean values

- Added the go-cmp package by Google (https://github.com/google/go-cmp) to make it easier to compare two slices of maps.

- `TheResponseAtShouldBe` has been updated to handle the check of multiple fields like this:
```
    And the response at $.children[*] should be:
      | string.title       | has_visible_children |
      | Child lvl1 Skill 1 | false                |
      | Child lvl1 Skill 2 | true                 |
      | Child lvl1 Skill 3 | false                |
```

The column is a `JSONPath`, or just the name of the field, which can also be considered as a `JSONPath`.

We need that because if we do two checks in a row:
- `And the response at $.children[*].string.title should be:`
- `And the response at $.children[*].has_visible_children should be:`
 
We would validate that there are two rows with `has_visible_children=false`, but it would not validate that they correspond to the rows with titles:
- `Child lvl1 Skill 1`
- `Child lvl1 Skill 3`

### DIFF
--- a/app/api/items/get_children.go
+++ b/app/api/items/get_children.go
@@ -249,7 +249,7 @@ func (srv *Service) getItemChildren(rw http.ResponseWriter, httpReq *http.Reques
 			`items.allows_multiple_attempts, category, score_weight, content_view_propagation,
 				upper_view_levels_propagation, grant_view_propagation, watch_propagation, edit_propagation,
 				items.id, items.type, items.default_language_tag,
-				validation_type, display_details_in_parent, duration, entry_participant_type, no_score,
+				items.validation_type, items.display_details_in_parent, items.duration, items.entry_participant_type, items.no_score,
 				IFNULL(can_view_generated_value, 1) AS can_view_generated_value,
 				IFNULL(can_grant_view_generated_value, 1) AS can_grant_view_generated_value,
 				IFNULL(can_watch_generated_value, 1) AS can_watch_generated_value,
@@ -283,7 +283,9 @@ func constructItemChildrenQuery(dataStore *database.DataStore, parentItemID, gro
 		dataStore, groupID, requiredViewPermissionOnItems, watchedGroupIDSet, watchedGroupID, columnList, columnListValues,
 		externalColumnList,
 		func(db *database.DB) *database.DB {
-			return db.Joins("JOIN items_items ON items_items.parent_item_id = ? AND items_items.child_item_id = items.id", parentItemID)
+			return db.Joins("JOIN items_items ON items_items.parent_item_id = ? AND items_items.child_item_id = items.id", parentItemID).
+				Joins("JOIN items AS parent_item ON parent_item.id = ?", parentItemID).
+				Where("(parent_item.type = 'Skill' AND items.type = 'Skill') OR parent_item.type <> 'Skill'")
 		},
 		func(db *database.DB) *database.DB {
 			return db.Joins("JOIN items_items ON items_items.child_item_id = item_id").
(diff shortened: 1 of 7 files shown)


app/api/items/get_children.go
```diff
@@ -283,7 +283,9 @@ func constructItemChildrenQuery(dataStore *database.DataStore, parentItemID, gro
 		dataStore, groupID, requiredViewPermissionOnItems, watchedGroupIDSet, watchedGroupID, columnList, columnListValues,
 		externalColumnList,
 		func(db *database.DB) *database.DB {
-			return db.Joins("JOIN items_items ON items_items.parent_item_id = ? AND items_items.child_item_id = items.id", parentItemID)
+			return db.Joins("JOIN items_items ON items_items.parent_item_id = ? AND items_items.child_item_id = items.id", parentItemID).
+				Joins("JOIN items AS parent_item ON parent_item.id = ?", parentItemID).
```
smadbe, line 287: If you introduce the `"JOIN items AS parent_item ON parent_item.id = ?", parentItemID`, IMO, you have to do the previous join `items_items` on `items_items.parent_item_id=parent_item.item_id`  instead of using the id again.

app/api/items/navigation_data_mapper.go
```diff
@@ -54,15 +54,27 @@ func getRawNavigationData(dataStore *database.DataStore, rootID, groupID, attemp
...
 	service.MustNotBeError(itemsQuery.Error())
 
 	hasVisibleChildrenQuery := dataStore.Permissions().MatchingGroupAncestors(groupID).
 		WherePermissionIsAtLeast("view", "info").
 		Joins("JOIN items_items ON items_items.child_item_id = permissions.item_id").
+		Joins(`
+			JOIN items AS child_items
+			  ON child_items.id = items_items.child_item_id
+			 AND (items.type = "Skill" AND child_items.type = "Skill")
+				OR items.type <> "Skill"
```
smadbe, line 72: Don't you mean 
```suggestion
			  ON child_items.id = items_items.child_item_id
			 AND ((items.type = "Skill" AND child_items.type = "Skill")
				OR items.type <> "Skill")
```
?
If so, couldn't you add a test to detect a similar regression in the future?

smadbe (changes requested):

app/api/items/get_children.go
GeoffreyHuck: fixed, it looks better

app/api/items/navigation_data_mapper.go
GeoffreyHuck: fixed

Wow, this one was tricky.

I couldn't come with a test to cover the case because without the extra parenthesis, it still works. But the fact it works without it is more a coincidence.

It is interpreted as:
```
JOIN items AS child_items
  ON (child_items.id = items_items.child_item_id AND (items.type = "Skill" AND child_items.type = "Skill"))
  OR items.type <> "Skill"
```

When `items.type="Skill"`, it is as expected. But when `items.type<>"Skill"`, it doesn't make any restriction on the other joined tables, so it has no effect. Which is also what we want because a Chapter cannot have Skills as children.

But the issue is that without restricting the other joined tables, it internally makes a cross product between the tables, which can be very slow and consume a lot of memory.

So, I added the parenthesis, but couldn't make a test for verification because it doesn't change the behavior.

smadbe (approved):

PR #963, opened 2023-05-24
All responses have a Backend-Version header with the version injected in main

### LINKED ISSUE #840: Expose app version
## Overview
Expose the app version so that a caller knows what version of the backend it is currently using.

The version number is currently injected at compile time in `main` (can be moved somewhere else).

## Acceptance criteria
- [ ] Add the "injected" version in a "Backend-Version" response header.

GeoffreyHuck:
fixes #840 

Uses a middleware to inject `Backend-Version` on all responses.

The version is globally accessible as `version.Version`.

I also considered storing the version in the Application{} struct, but it required passing the version in many different places, in tests, and in all commands.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -17,6 +17,7 @@ import (
 	"github.com/France-ioi/AlgoreaBackend/app/domain"
 	"github.com/France-ioi/AlgoreaBackend/app/logging"
 	"github.com/France-ioi/AlgoreaBackend/app/rand"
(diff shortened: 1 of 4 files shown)


smadbe (approved):
Looks good. Tests are not great (if "Backend-Version" header would be always empty, it would not necessarily be detected) but that should do the job for this non-critical and not changing feature.

PR #969, opened 2023-06-20
GetCurrentAnswer now returns a 200 response when there is no current answer

### LINKED ISSUE #837: Change expected errors into sucesses: get current answer
## Overview

Modern browsers report any 4xx return code as an error logged in the console. These errors may be considered as issues by our customers and reported as such.  

So let's try to return non-error for error cases which are not considered by the frontend as errors (used on purpose to give a specific interpretation to errors)

### Specific case: get current answer service

The "Get a current answer" service returns the current answer if any, and 403 if there is none. As the frontend does not know that in advance, it tries to fetch the current answer and just starts the task from scratch if there is no current answer (currently if receives a 403). 

### Constraints

- it is better if the change is backward compatible (or at least does not need major frontend update which needs to be deployed in sync with the backend)
- let's not add another call if not needed

## Proposal

- When there is no current answer, return a HTTP 200 with, as payload,`null` (valid json) instead of the record.

GeoffreyHuck:
fixes #837 

To return a null JSON response, the standard way of doing it is to set one property of the JSON to `null` (see https://stackoverflow.com/questions/21120999/representing-null-in-json ). We cannot just return `null` without a property, and `{}` would require weird javascript to be detected.

The `type` attribute was chosen because it's the one that makes the most sense with `null`. Furthermore, it avoids adding a new field.

The existing response with all the fields are:
```
    {
      "id": "103",
      "attempt_id": "1",
      "participant_id": "11",
      "score": 93.0,
      "answer": "print(3)",
      "state": "State103",
      "created_at": "2017-05-29T06:38:38Z",
      "type": "Current",
      "item_id": "200",
      "author_id": "11",
      "graded_at": "2018-05-29T06:38:33Z"
    }
```

An answer "no answer" is now:
```
    {
      "type": null,
    }
```

- Previously, the existence of the answer was checked along with the permissions in a single SQL query. The method `IsVisible` was used to append the permissions check in the query. The previous tests for `IsVisible` have been adapted and reused to check the new method `user.CanSeeAnswer()`.
- The creation of the query to get the answer has been refactored into a method `GetCurrentAnswerQuery` to simplify the service

Note: PR https://github.com/France-ioi/AlgoreaBackend/pull/970 adds an additional refactoring


smadbe (approved):
👍

smadbe (changes requested):
The service spec must be updated to explain that new behavior. Is the response format even still valid? And as it will become unclear... (all attributes (but type) will become optional , it should be clear enough which ones have are expected to be set in the non-null case)

GeoffreyHuck:
Added comments to state the case.

Note: there is no way to state that the response can have different schema depending on the context. There is a way with OpenAPI 3.* with `oneOf`, but nothing with OpenAPI 2.0. Also, I checked and there is no plan for go-swagger to support OpenAPI 3.*: https://github.com/go-swagger/go-swagger/issues/1122

smadbe (approved):

Issue #968, opened 2023-06-14
Update services for the recent forum permission


smadbe:
## Motivations
Recently, we have added a new permission for the forum. This new perm should be added in the services related with permission management.

## Subtasks

- [x] add `can_request_help_to` to `grantedPermissionsView`
- [x] add `can_request_help_to` to `permissionsView` in `granted`. The value should be either null (if not the perm), either a tuple containing the `id`, `name` only if it is visible to the current-user, either a `is_all_users_group` boolean to indicate whether the group is the "all-users" group.
- [x] add  `can_request_help_to` to `updatePermissions`: [as written in the doc](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/), the group must visible to both the giver and the received. In order to be able to set the "all group" group, the value should be a tuple with either `id` or `is_all_users_group: true` (not both) (or at root `can_request_help_to_all_users:bool`|`can_request_help_to:number` ?)
- [x] add `request_help_propagation` to `itemCreate`
- [x] add `request_help_propagation` to `itemUpdate`
- [x] add `request_help_propagation` to `itemChildrenView`

GeoffreyHuck:
Note: Found out it has already been done for `itemUpdate` here: https://github.com/France-ioi/AlgoreaBackend/commit/d64a5abbc20c034245d23160233c11afab38ef70 (but not for `itemCreate` yet)

PR #971, opened 2023-06-20
Add can_request_help_to to grantedPermissionsView

GeoffreyHuck:
Part 1 of https://github.com/France-ioi/AlgoreaBackend/issues/968

Here we want to add `can_request_help_to` only to the service `grantedPermissionsView`. The problem is that the structure containing the permissions is reused in many different services: basically, every time `permissions` appears in a response.

There are a few options:
- Either we update all the services to add the new field. But it's not that easy because we would have to update all the tests and all the SELECT statements.
- We can also redefine the permission structure in this case. Not easy because it is nested into other structures. If we want a different structure, we have to duplicate a tree of structures.
- The solution implemented: add `omitempty` to the definition of the structure. With this, `can_request_help_to` appears in the resulting JSON only if it is defined. So the output of existing services doesn't change. **But the catch is:** if `can_request_help_to` is `null`, the parameter will not appear at all in the response. A comment was added to appear in the API doc.

### DIFF
--- a/app/structures/structures.go
+++ b/app/structures/structures.go
@@ -21,8 +21,8 @@ type ItemPermissions struct {
 	CanEdit string `json:"can_edit"`
 	// required: true
 	IsOwner bool `json:"is_owner"`
-	// appears only in services that support this permission, and only if it is set
-	// required: true
+	// appears only in services that support this permission (see service description), and only if it is set
+	// required: false
 	CanRequestHelpTo *int64 `json:"can_request_help_to,omitempty"`
 }
 
--- a/app/structures/structures.go
+++ b/app/structures/structures.go
@@ -21,6 +21,9 @@ type ItemPermissions struct {
 	CanEdit string `json:"can_edit"`
 	// required: true
 	IsOwner bool `json:"is_owner"`
+	// appears only in services that support this permission, and only if it is set
+	// required: true
+	CanRequestHelpTo *int64 `json:"can_request_help_to,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


app/structures/structures.go
```diff
@@ -21,6 +21,9 @@ type ItemPermissions struct {
 	CanEdit string `json:"can_edit"`
 	// required: true
 	IsOwner bool `json:"is_owner"`
+	// appears only in services that support this permission, and only if it is set
```
smadbe, line 24: reading the doc, how do I know if a given service support this permission?

app/structures/structures.go
```diff
@@ -21,6 +21,9 @@ type ItemPermissions struct {
...
+	// required: true
```
smadbe, line 25: if it may not appear, it is not "required"

GeoffreyHuck:
> Cannot you just define (don't remember if doable in Go) a struct "extending" this one with just 1 additional field ?

Yes, but then you would also have to redefine the parent struct to declare that the other struct is used... etc until the root struct. So it would lead to a lot of duplication.

app/structures/structures.go
GeoffreyHuck: fixed

app/structures/structures.go
GeoffreyHuck: fixed, added better doc to be explicit about it. It doesn't feel very satisfying to handle it like this, but the alternative of creating multiple structs just for one field feels worse...

smadbe (approved):

PR #974, opened 2023-07-03
PermissionView: add can_request_help_to in granted permissions

GeoffreyHuck:
Related to #968 

The query is a little tricky, added a detailed comment to explain why is was done like this.
The tests have been put in another Gherkin file, reason is in comments.

Also for the Gherkin tests:
* Added support for edge cases in checking response parts:
- When an object is `null`
- When we want to compare to a `bool`
* Added the possibility to compare a full object with the feature `the response at XXX should be:": checks that the designed jsonPath contains exactly the properties listed.
* Added a feature for setting a group as the `AllUsers` group without having to refer to the configuration in the test*
* Added a feature to give the manager the permission `can grant group access`

### DIFF
--- a/app/api/groups/get_permissions_can_request_help_to.feature
+++ b/app/api/groups/get_permissions_can_request_help_to.feature
@@ -1,0 +1,59 @@
+# Those scenario cannot, for now, be merged with those in get_permissions.feature
+#
+# Reason: The scenario in this file are defined with new Gherkin features which allows higher-level definitions.
+#         Those features require the propagation of permissions to run.
+# Problem: the permissions defined in get_permissions.feature contain inconsistent data.
+#          It means that if we move the definitions of the table permissions_generated into the equivalent in permissions_granted,
+#          and then we run the propagation of permissions, we get a different result than
+#          the permissions currently defined in permissions_generated, and many tests then fail.
+#          If those permissions definitions get fixed, then this file can be merged with them.
+Feature: Get permissions can_request_help_to for a group
+  Background:
+    Given allUsersGroup is defined as the group @AllUsers
+    And there are the following groups:
+      | group     | parent | members  |
+      | @AllUsers |        |          |
+      | @School   |        | @Teacher |
+      | @Class    |        |          |
+    And @Teacher is a manager of the group @Class and can grant group access
+    And there are the following tasks:
+      | item  |
+      | @Item |
+    And there are the following item permissions:
+      | item  | group    | is_owner | can_request_help_to |
+      | @Item | @Teacher | true     |                     |
+
+  Scenario: Should return helper group when set and visible by the current user
+    Given I am @Teacher
+    And there is a group @HelperGroup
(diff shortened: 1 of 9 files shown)


app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -0,0 +1,59 @@
...
+    And there are the following item permissions:
+      | item  | group  | is_owner | can_request_help_to |
+      | @Item | @Class | false    | @HelperGroup        |
+    When I send a GET request to "/groups/@Class/permissions/@Class/@Item"
+    Then the response code should be 200
+    And the response at $.granted.can_request_help_to should be "null"
+
+  Scenario: Should return helper group as "All Users" group when set to its value
+    Given I am @Teacher
+    And the group @AllUsers is visible to the user @Teacher
```
smadbe, line 51: Hum I don't get this. 

1. Group visibility should "checkable" (=gettable) but not "settable" as it may be through one of the three conditions (public, member or manager)... depending on the way you implement this gherkin rule, it may have serious side effects.

2. Anyway, having "all users" may be the case for anybody, and the point of having this "is_all_users_group" property instead of the group id is not to require any visibility conditions on the group id. So if I understand correctly `    And the group @AllUsers is visible to the user @Teacher` , it misses the point.

May be discussed if not clear.

smadbe (changes requested):
👍  the rest looks good

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: fixed:

1. The feature is now specific about what is done, with a comment stating the intent:
```
    # @HelperGroup is visible by @Teacher
    And the group @Teacher is a descendant of the group @HelperGroup
```

2a. The check IsVisible is not done anymore for the AllUsers group.
2b. The feature `allUsersGroup is defined as the group @AllUsers` has been improved, and now all the users are really members of the allUsers group.

app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -0,0 +1,60 @@
...
+    And there are the following item permissions:
+      | item  | group    | is_owner | can_request_help_to |
+      | @Item | @Teacher | true     |                     |
+
+  Scenario: Should return helper group when set and visible by the current user
+    Given I am @Teacher
+    And there is a group @HelperGroup
+    # @HelperGroup is visible by @Teacher
+    And the group @Teacher is a descendant of the group @HelperGroup
+    And the group @HelperGroup is visible to the user @Teacher
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed, forgot to delete. Also deleted the code of the feature.

GeoffreyHuck:
Note: The whole commit history of this PR has been cleared because I realized the commit contained many small changes to the test system that were packed into a big commit. All the small changes are in an atomic commit now.

smadbe (approved):
👍

PR #977, opened 2023-07-05
updatePermissions: can update can_request_help_to with a group or with group AllUsers

GeoffreyHuck:
Related to #968 

As in previous PR, tests have been put in different *.feature files to use the new system.

Notes:

1:
Setting `can_request_help_to` with the id of the `AllUsers` group has been done by modifying directly the dataMap used to update the fields in database. If we find the field `is_all_users_group`, we replace it by `can_request_help_to`: `allUsersGroupID`.

2:
Multiple validators have been used to validate the input. This allows to have a different error message when it fail. It also allows the robustness test to be more robust, because it is checked if they fail for the right reason.

3. Added a comment about why MAX() was used for the request as I found it weird.

4. The methods to check if a `group is visible` were implemented only with a user. But in this case we need to use it on a group. So the functions have been duplicated and factorized. The names are a bit ugly though... `ManagedUsersAndAncestorsOfManagedGroupsForGroup`, `AncestorsOfJoinedGroupsForGroup`, `PickVisibleGroupsForGroup`.

### DIFF
--- a/app/api/groups/update_permissions.go
+++ b/app/api/groups/update_permissions.go
@@ -9,6 +9,7 @@ import (
 	"github.com/jinzhu/gorm"
 
 	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/domain"
 	"github.com/France-ioi/AlgoreaBackend/app/formdata"
 	"github.com/France-ioi/AlgoreaBackend/app/service"
 )
@@ -43,6 +44,23 @@ type updatePermissionsInput struct {
 	// The current user should have permissions_generated.can_grant_view_generated >= 'enter' in order to
 	// increase this field's value.
 	CanEnterUntil time.Time `json:"can_enter_until" validate:"can_enter_until"`
+	// Optional
+	// The current user should have `permissions_generated.can_grant_view` >= 'content',
+	// in order to set this field's value.
+	CanRequestHelpTo setCanRequestHelpTo `json:"can_request_help_to" validate:"can_request_help_to_set,can_request_help_to_consistent,can_request_help_to_can_grant_view_content,can_request_help_to_visible"` //nolint:lll
+}
+
+// setCanRequestHelpTo
+// Cannot set `id` and `is_all_users_group` at the same time,
+// but one of them have to be set if `can_request_help_to` is provided.
+type setCanRequestHelpTo struct {
+	// The given group must be visible by both the current user and `group_id`.
+	// Can be set to `null` to set the helper group to no group.
+	// Optional
+	// Nullable
+	ID *int64 `json:"id" sql:"column:can_request_help_to"`
+	// Optional
+	IsAllUsersGroup bool `json:"is_all_users_group"`
 }
 
 type userPermissions struct {
@@ -54,6 +72,7 @@ type userPermissions struct {
 	CanEnterUntil          database.Time
 	CanMakeSessionOfficial bool
 	IsOwner                bool
+	CanRequestHelpTo       *int64
 }
 
 type managerGeneratedPermissions struct {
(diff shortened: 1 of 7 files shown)


app/api/groups/update_permissions.go
```diff
@@ -43,6 +44,19 @@ type updatePermissionsInput struct {
...
+	// The current user should have `permissions_generated.can_grant_view` >= 'content',
+	// in order to set this field's value.
+	CanRequestHelpTo setCanRequestHelpTo `json:"can_request_help_to" validate:"can_request_help_to_set,can_request_help_to_consistent,can_request_help_to_can_grant_view_content,can_request_help_to_visible"` //nolint:lll
+}
+
+// Cannot set `id` and `is_all_users_group` at the same time.
+type setCanRequestHelpTo struct {
+	// The given group must be visible by both the current user and `group_id`.
+	// Can be set to `null` to set the helper group to no group.
+	ID *int64 `json:"id" sql:"column:can_request_help_to"`
```
smadbe, line 57: So I assume it should be marked as`Optional` as well.

I assume it should be marked as `nullable` (I don't remember if it is an "official" tag but at least it is how we did so far)

app/api/groups/update_permissions.go
```diff
@@ -43,6 +44,19 @@ type updatePermissionsInput struct {
 	// The current user should have permissions_generated.can_grant_view_generated >= 'enter' in order to
 	// increase this field's value.
 	CanEnterUntil time.Time `json:"can_enter_until" validate:"can_enter_until"`
+	// Optional
+	// The current user should have `permissions_generated.can_grant_view` >= 'content',
+	// in order to set this field's value.
+	CanRequestHelpTo setCanRequestHelpTo `json:"can_request_help_to" validate:"can_request_help_to_set,can_request_help_to_consistent,can_request_help_to_can_grant_view_content,can_request_help_to_visible"` //nolint:lll
+}
+
+// Cannot set `id` and `is_all_users_group` at the same time.
```
smadbe, line 53: but one of the two must be set I assume

smadbe, line 53: does that comment appear in the doc? it probably should

app/api/groups/update_permissions.go
```diff
@@ -356,6 +392,97 @@ func registerCanEnterUntilValidator(
 	})
 }
 
+func registrerCanRequestHelpToSetValidator(data *formdata.FormData, currentPermissions *userPermissions, modified *bool) {
```
smadbe, line 395: typo "register" I assume

app/api/groups/update_permissions_can_request_help_to.feature
```diff
@@ -0,0 +1,79 @@
+# Those scenario cannot, for now, be merged with those in get_permissions.feature
```
smadbe, line 1: `update_permissions.feature` I assume

app/api/groups/update_permissions_can_request_help_to.feature
```diff
@@ -0,0 +1,79 @@
...
+#
+# Reason: The scenario in this file are defined with new Gherkin features which allows higher-level definitions.
+#         Those features require the propagation of permissions to run.
+# Problem: the permissions defined in get_permissions.feature contain inconsistent data.
```
smadbe, line 5: same

smadbe (commented):
Would be useful to have a short description of what this PR does, even just 1-2 sentences. "Related to 968" does not tell me what is the intend / scope of the PR (what it includes, what it does not).

app/api/groups/update_permissions.go
GeoffreyHuck: fixed

app/api/groups/update_permissions.go
GeoffreyHuck: fixed, added ```one of them have to be set if `can_request_help_to` is provided``` in the comment.

GeoffreyHuck: it does, I checked

app/api/groups/update_permissions.go
GeoffreyHuck: fixed

app/api/groups/update_permissions_can_request_help_to.feature
GeoffreyHuck: fixed

app/api/groups/update_permissions_can_request_help_to.feature
GeoffreyHuck: fixed

GeoffreyHuck:
Note: the complete history of git has been redone on the commits in this PR too because it was completely messed up.

PR #978, opened 2023-07-11
accessTokenCreate: Create a temp user in case of failed authorization only if create_temp_user_if_not_authorized=true

GeoffreyHuck:
Retrieve and adapted the test that was previously deleted.

The way the error is returned is a little weird with the use of `GetAuthErrorCodeFromError`.
This is because we use the `auth middleware service` to check if the authorization is valid, and this service cannot use the module used to create apiError (`service`) because it would make a `circular dependency error` (was tried last time).

We could change that by putting `apiError` in another module, but since it's the only exception, it might be too much.

Also documented the reason of this functionality in the service comment to keep a record.

Finally, added a function to retrieve a boolean from the query with a default value if it is not present as the code to do it is a little bit ugly and complex.

### DIFF
--- a/app/api/auth/create_access_token.go
+++ b/app/api/auth/create_access_token.go
@@ -204,13 +204,9 @@ func (srv *Service) createAccessToken(w http.ResponseWriter, r *http.Request) se
 			service.AppHandler(srv.refreshAccessToken).
 				ServeHTTP(w, r.WithContext(context.WithValue(requestContext, parsedRequestData, requestData)))
 		} else {
-			createTempUser := false
-			if len(r.URL.Query()["create_temp_user_if_not_authorized"]) > 0 {
-				var err error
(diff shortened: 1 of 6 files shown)


app/api/auth/create_access_token.go
```diff
@@ -122,7 +131,12 @@ const parsedRequestData ctxKey = iota
 //			type: integer
 //			enum: [0,1]
 //			default: 0
+//		- name: create_temp_user_if_not_authorized
+//			description: Whether to create a temporary user if the token is not provided or expired.
+//			in: query
+//			type: boolean
```
smadbe, line 137: may be useful to mention that if not given, it is assumed `false`

smadbe (approved):

app/api/auth/create_access_token.go
GeoffreyHuck: fixed, added `default`. Also changed type `boolean` to `integer` with enum `[0, 1]`, because that's what we want.

PR #980, opened 2023-07-12
Disable GCO on build

GeoffreyHuck:
The last deployment failed with this problem:
`"/var/task/AlgoreaBackend-linux: /lib64/libc.so.6: version GLIBC_2.34' not found (required by /var/task/AlgoreaBackend-linux)"`

GO was updated since last time.

By default, `go build` adds a dynamic dependency to `libc`, and other libraries. It can be disabled with the flag `CGO_ENABLED=0`.

Note: We need GCO for the tests, because we use the `-race` flag, which is used to detect race conditions bugs: https://go.dev/blog/race-detector

It seems like a good idea to keep it for tests. If we use the flags for `go test`, it makes the test fail with an error stating the gco must be enabled to use -race.

So:
- We disable `GCO` for the build
- We keep it for the tests

With GCO, this is an analysis of the binary
```
readelf -d bin/AlgoreaBackend

Dynamic section at offset 0xe4f060 contains 20 entries:
  Tag        Type                         Name/Value
 0x0000000000000004 (HASH)               0xde5f40
 0x0000000000000006 (SYMTAB)             0xde6420
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000005 (STRTAB)             0xde61e0
 0x000000000000000a (STRSZ)              546 (bytes)
 0x0000000000000007 (RELA)               0xde5b50
 0x0000000000000008 (RELASZ)             24 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x0000000000000003 (PLTGOT)             0x124ef40
 0x0000000000000015 (DEBUG)              0x0
-----------> 0x0000000000000001 (NEEDED)             Shared library: [libresolv.so.2]
-----------> 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
-----------> 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000006ffffffe (VERNEED)            0xde5ee0
 0x000000006fffffff (VERNEEDNUM)         2
 0x000000006ffffff0 (VERSYM)             0xde5e80
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000002 (PLTRELSZ)           768 (bytes)
 0x0000000000000017 (JMPREL)             0xde5b68
 0x0000000000000000 (NULL)               0x0
```

Without GCO:
```
readelf -d bin/AlgoreaBackend

There is no dynamic section in this file.
```

It's way better to have no dynamic dependencies. And if one day we need libc or something else, it's better to consider linking them statically instead of dynamically.


smadbe (approved):
👍

PR #982, opened 2023-07-12
Add arch info for go build

GeoffreyHuck:
To see if it fixes the new deployment issue

See https://stackoverflow.com/questions/50700979/exec-format-error-when-running-aws-golang-lambda


smadbe:
The general build command shouldn't force a specific arch. For instance, I used to compile this on OSX without issue. The linux build rule may be adapted...
Anyway this was not the source of the issue (it was a race condition in the deployment process), so let's close this PR.

PR #984, opened 2023-07-17
itemChildrenView: add request_help_propagation in response

### LINKED ISSUE #968: Update services for the recent forum permission

GeoffreyHuck:
fixes #968 

We simply get the value from the db and return it.

Note: This service also returns a `permissions` field, but we don't return `can_request_help_to` because it wasn't added in the issue. We could if we need too.

### DIFF
--- a/app/api/items/get_children.feature
+++ b/app/api/items/get_children.feature
@@ -43,10 +43,10 @@ Feature: Get item children
       | manager_id | group_id | can_watch_members |
       | 22         | 15       | true              |
     And the database has the following table 'items_items':
-      | parent_item_id | child_item_id | child_order | category  | score_weight | content_view_propagation | upper_view_levels_propagation | grant_view_propagation | watch_propagation | edit_propagation |
-      | 200            | 210           | 2           | Discovery | 1            | none                     | use_content_view_propagation  | true                   | false             | true             |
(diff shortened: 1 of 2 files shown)


smadbe (approved):
👍

PR #991, opened 2023-08-16
Breadcrumb from root: add not started paths

GeoffreyHuck:
```
dle
Bref, oui, en effet, pour certaines utilisations en effet (pas encore vrmt implémentée correctement dans le
frontend), il faut plutôt tous les paths, mm ceux pas encore visités...
donc soit il faut soit un autre service, soit un parameter pour dire d'inclure ou pas les non visités par le
participants, soit le mettre dans l'output.
Ce qui est probablement plus flex pour le frontend pour les utilisations souhaitées/futures, c'est mettre dans
l'output si le path est starté par le participant (en précisant bien que c'est visité par le participant) (edited) 

Geoffrey Huck
et pour `breadcrumb-from-root`, je vais utiliser la même requête que `path-from-root` mais en retournant tous
les chemins trouvés plutôt que seulement le premier, et en ajoutant un champ booléen `started_by_participant`
```

### Idea

`BreadcrumbsFromRoot` now returns an additional boolean field `started_by_participant` with each path.

 To find the path, the query was merged with the one `PathFromRoot`, with 2 variations:
- We get all results, whereas we take only the first one in `PathFromRoot`
- The root items of the path are computed for the `user`, whereas in `PathFromRoot` it is for the `participant`. I don't know the reason but many existing tests are supporting this.

Tests have been updated to reflect the changes.

**Note:** The query may return a path more than once, for example when a path exists with different attempts.
`FindItemsPath` function has been updated to de-duplicate the path returned

### `is_active` and `is_started`

There is two notions for a path:
- `is_active` is when the all the items have either a `result.started_at`, or if it is possible right `NOW()` to make a submission for the item/attempt, typically with the attempt 0
- `is_started`is when all the items have a `result.started_at`

A mistake was made initially when we returned `is_active` instead of `is_started`. It was corrected in commit https://github.com/France-ioi/AlgoreaBackend/pull/991/commits/b3deec450a1fae7d7a83c3dcc5ff707e005d29cc . We want `is_started`.

Also added a test to make sure `is_started` is computed for the `participant` and not to the `user`.

### DIFF
--- a/app/api/items/get_breadcrumbs_from_roots.feature
+++ b/app/api/items/get_breadcrumbs_from_roots.feature
@@ -8,7 +8,7 @@ Feature: Find all breadcrumbs to an item
       | 93  | Class | null             | 90            |
       | 94  | Club  | null             | null          |
       | 101 | User  | null             | null          |
-      | 102 | Team  | null             | 60            |
+      | 102 | Team  | 100              | 60            |
       | 111 | User  | null             | null          |
     And the database has the following table 'users':
       | login | group_id | default_language |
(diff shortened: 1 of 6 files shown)


smadbe (approved):
Should be good. The problem is that `get_breadcrumbs_from_roots` change is not backward compatible so I will have to sync that with a frontend release.

zenovich:
I translated the task description into English for readers who don't understand French, but I still don't get what the task was. Can someone explain me what the task was? Also, can someone explain why, when looking for root items, we consider groups managed by the user in itemBreadcrumbsFromRootsByTextIdGet and groups managed by the participant in all other services? Looks like it's just a bug and we should always consider groups managed by the participant or maybe I don't get something here.

zenovich:
@smadbe, @GeoffreyHuck

smadbe:
> I've translated the task description into English for readers who don't understand French, but I still don't get what the task was (the reason behind the described changes). Can someone explain me what the task was? 

Let me explain the problem that lead to this task (from what I remember as it is 1y old):

The [itemPathFromRootFind](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v1.9.3.html#tag/items/operation/itemPathFromRootFind) is called when a user arrives on an activity but has no-path, an invalid path or just a path that he has not started yet (so invalid for the user) (the most common case, you give a link to a user to a part of the content he hasn't visited yet). In this case, we were getting errors from the backend with this service as if the content didn't exist while it was just that the parent chapter had not been visited yet. So the goal of this PR was initially mainly to fix that unwanted behavior.

> Also, can someone explain why, when looking for root items, we consider groups managed by the user in itemBreadcrumbsFromRootsByTextIdGet and groups managed by the participant in all other services?  Looks like it's just a bug and we should always consider groups managed by the participant or maybe I don't get something here.

Yes, I assume that's a mistake.

PR #992, opened 2023-08-16
Migrate prod database

GeoffreyHuck:
- Cherry-picked migrations in https://github.com/France-ioi/AlgoreaBackend/pull/917
- Deduplicate `groups`.`sPassword`

### DIFF
--- a/db/migrations/1900000000__preprod-changes.sql
+++ b/db/migrations/1900000000__preprod-changes.sql
@@ -1,0 +1,15 @@
+-- +migrate Up
+
+ALTER TABLE `groups_login_prefixes` DROP COLUMN `idUserCreator`;
+ALTER TABLE `items_items` DROP COLUMN `iWeight`;
+
+ALTER TABLE `groups` DROP KEY `sPassword`;
+ALTER TABLE `groups` DROP KEY `sType`;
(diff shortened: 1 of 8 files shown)


smadbe (changes requested):
Could you check what fails in the test? It seems to be a legitimate error, so probably there is something to fix (otherwise, the next PR containing migrations will fail as well I think)

db/migrations/1900000000__preprod-changes.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Up
+
+ALTER TABLE `groups_login_prefixes` DROP COLUMN `idUserCreator`;
```
GeoffreyHuck, line 3: This is the line that makes the migration fail

GeoffreyHuck:
I annoted the line with the migration problem here: https://github.com/France-ioi/AlgoreaBackend/pull/992/files

It looks like the database used doesn't have the same schema as the one used to write the migrations.

smadbe:
Ok, but at the end we have to do something to prevent this error to raise each time we write a migration from now. Either fix the db seed and fix the migration file.

smadbe (changes requested):
So this is not waiting for my review anymore I think.

PR #993, opened 2023-08-17
Add `has_can_request_help_to` in `getItem`

### LINKED ISSUE #989: Add `can_request_help_to` in `getItem`
## Request

Add info about the `can_request_help` perm  in the `itemView` service. Both in `permission` and `watched_group.permission`. 

## Motivation

The goal behind that request is, for the frontend, to know whether, on a task, to display button to display the forum (and to request forum information).  It can just be a single boolean, the detail of groups (not aggreggable) could be served by another service.

* So maybe that should only be given for the tasks. 

## Notes

* This should be aggregation on groups **and** items using the propagation condition on items_items (so using recursive queries?) 

* Don't forget to update the service description as `can_request_help_to`: `appears only in services that support this permission (see service description), and only if it is set`

# Initial request (outdated)
 
## Request

Add the `can_request_help_to` permission in the `itemView` service. Both in `permission` and `watched_group.permission`. 

## Motivation

The goal behind that request is, for the frontend, to know whether, on a task, to display button to display the forum (and to request forum information). Then once we open the forum pane, we need to know to what groups we can assign the request.

* So maybe that should only be given for the tasks. 

## Notes

* I assume that should be an array as they may be several groups

* This should be aggregation on groups **and** items using the propagation condition on items_items (so using recursive queries?) 

* As done at another place, if the group is "all users", it should be something like "{ all_group: true }" as the user does not what is the "all group" group. 

* If this makes this get-item request too expensive (called very often), we may reconsider including it in this request... or at least have `can_request_help: bool` in this request and the details in another request that we would call only when we need to know this list of groups (not that often actually)

* Don't forget to update the service description as `can_request_help_to`: `appears only in services that support this permission (see service description), and only if it is set`

GeoffreyHuck:
fixes #989 

Added a boolean field `HasCanRequestHelpTo` in the permission structure. As it was done with `CanRequestHelpTo`, it is annoted to be available only if it is mentioned in the description of the service. I tried to find another way to do it, by duplicating the structure for permissions, but it ended up requiring many changes in other files.

The recursive query to check `help_request_propagation` was already made so it's been reused.

### DIFF
--- a/app/api/items/get_item.feature
+++ b/app/api/items/get_item.feature
@@ -1,5 +1,4 @@
 Feature: Get item view information
-
   Background:
     Given the database has the following table 'groups':
       | id | name       | grade | type  |
@@ -137,7 +136,8 @@ Feature: Get item view information
         "can_grant_view": "enter",
         "can_view": "solution",
         "can_watch": "result",
-        "is_owner": true
+        "is_owner": true,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -190,7 +190,8 @@ Feature: Get item view information
         "can_grant_view": "none",
         "can_view": "solution",
         "can_watch": "none",
-        "is_owner": false
+        "is_owner": false,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -242,7 +243,8 @@ Feature: Get item view information
         "can_grant_view": "none",
         "can_view": "content_with_descendants",
         "can_watch": "none",
-        "is_owner": false
+        "is_owner": false,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -299,7 +301,8 @@ Feature: Get item view information
         "can_grant_view": "none",
         "can_view": "solution",
         "can_watch": "none",
-        "is_owner": false
+        "is_owner": false,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -356,7 +359,8 @@ Feature: Get item view information
         "can_grant_view": "none",
         "can_view": "solution",
         "can_watch": "none",
-        "is_owner": false
+        "is_owner": false,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -408,7 +412,8 @@ Feature: Get item view information
         "can_grant_view": "none",
         "can_view": "info",
         "can_watch": "none",
-        "is_owner": false
+        "is_owner": false,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -465,7 +470,8 @@ Feature: Get item view information
         "can_grant_view": "enter",
         "can_view": "solution",
         "can_watch": "result",
-        "is_owner": true
+        "is_owner": true,
+        "has_can_request_help_to": false
       }
     }
     """
@@ -482,7 +488,7 @@ Feature: Get item view information
       | 27       | 220     | none               | <can_grant_view_generated_ancestor> | none               | none                  | false              |
     And the template constant "permissions" is:
     """
-      "permissions": {"can_edit": "all", "can_grant_view": "solution_with_grant", "can_view": "content_with_descendants", "can_watch": "answer", "is_owner": true}
+      "permissions": {"can_edit": "all", "can_grant_view": "solution_with_grant", "can_view": "content_with_descendants", "can_watch": "answer", "is_owner": true, "has_can_request_help_to": false}
     """
     And the template constant "average_score" is:
     """
@@ -542,7 +548,8 @@ Feature: Get item view information
         "can_grant_view": "<expected_can_grant_view>",
         "can_view": "solution",
         "can_watch": "<can_watch_generated>",
-        "is_owner": false
+        "is_owner": false,
+        "has_can_request_help_to": false
       }
       <expected_watched_group_part>
     }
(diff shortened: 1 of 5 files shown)


smadbe (commented):
> Added a boolean field HasCanRequestHelpTo in the permission structure. As it was done with CanRequestHelpTo, it is annoted to be available only if it is mentioned in the description of the service. I tried to find another way to do it, by duplicating the structure for permissions, but it ended up requiring many changes in other files.

It pollutes a lot all the other services if you have one attribute which is never present but just there just as the struct is shared among several services. I don't really like that. It makes things difficult to understand for those using the API.
Cannot the Go struct extension be used to have this attribute only available in the service using it?

GeoffreyHuck:
Another approach implemented in the last commit: put the `has_request_help_to` fields out of the `permissions` substructures.

So:
- `$.permissions.has_request_help_to`  becomes `$.has_request_help_to`  
- `$.watched_gorup.permissions.has_request_help_to`  becomes `$.watched_group.has_request_help_to`  

The other solution would have been to create other structures just for this service. The problem is that the code is using methods to fill up the existing structures `asItemCommonFields()`. As far as I know, we would have to duplicate those methods to make them work with both structures. It's also possible I'm missing something.

GeoffreyHuck:
Found another solution in the last commit. I put back the fields in `permissions`. The solution was to "explode" the structs during construction.

app/api/items/get_item.feature
```diff
@@ -137,7 +136,8 @@ Feature: Get item view information
         "can_grant_view": "enter",
         "can_view": "solution",
         "can_watch": "result",
-        "is_owner": true
+        "is_owner": true,
+        "has_can_request_help_to": false
```
smadbe, line 140: Why `has_can_request_help_to`  instead of `can_request_help` as suggested?
It looks more confusing to me.

app/api/items/get_item_can_request_help_to_groups.feature
```diff
@@ -0,0 +1,67 @@
+Feature: Get permissions has_can_request_help_to for an item
```
smadbe, line 1: good idea to make a test file just to test this property! 👍

app/api/items/get_item_can_request_help_to_groups.feature
```diff
@@ -0,0 +1,67 @@
...
+      | @Class              | @School | @Student,@HelperGroup |
+    And @Teacher is a manager of the group @Class and can watch its members
+    And there are the following items:
+      | item                 | parent    | type    | request_help_propagation |
+      | @Chapter1            |           | Chapter |                          |
+      | @Chapter2            |           | Chapter |                          |
+      | @Chapter3            |           | Chapter |                          |
+      | @Chapter4            |           | Chapter |                          |
+      | @Item1               | @Chapter1 | Task    |                          |
+      | @Item2               | @Chapter2 | Task    | true                     |
```
smadbe, line 18: Not clear to me what `request_help_propagation` column means on an item (as it is a property of a relation). Don't you want to call it "request_help_propagation_with_parent"? Maybe a bit long... :-/
And what does it mean at L20 when there is a parent but no value?

app/api/items/get_item_can_request_help_to_groups.feature
```diff
@@ -0,0 +1,67 @@
...
+    When I send a GET request to "/items/<item_id>"
+    Then the response code should be 200
+    And the response at $.permissions.has_can_request_help_to should be "<has_can_request_help_to>"
+    Examples:
+      | item_id              | has_can_request_help_to |
+      | @Item1               | true                    |
+      | @Item2               | true                    |
+      | @Item2_NoPropagation | false                   |
+      | @Item3               | true                    |
+      | @Item4               | true                    |
```
smadbe, line 43: Do you test group propagation here or just item propagation? On `@Item3` maybe? (depends on the answer on my prev question actually)

smadbe (changes requested):

app/api/items/get_item.feature
GeoffreyHuck: fixed

app/api/items/get_item_can_request_help_to_groups.feature
GeoffreyHuck: fixed for `request_help_propagation `. Definition of item relations was added for `viewPermissions`, so it's been reused.

No value for `request_help_propagation` means: "this value wasn't necessary to write the test". So that the property is not relevant to the test. It makes reading the test (at least for me), easier because I don't have to focus on information that is not relevant.

app/api/items/get_item_can_request_help_to_groups.feature
GeoffreyHuck: I updated the comments so they're more explicit:
- @Item1: the permission is on the item
- @Item2: item propagation enabled and disabled
- @Item3: group propagation
- @Item4: group+(item propagation enabled and disabled)

Also added a test with group+(item propagation disabled) to be more explicit.

item4: did you call me?

smadbe: > did you call me?

🤣 he was just showing a example with an item number 4.

smadbe (approved):

app/api/items/get_item_can_request_help_to_groups.feature
GeoffreyHuck: > did you call me?

Woops it wasn't expected ! I hope everything's fine for you in Korea ! Greetings from France !

Issue #988, opened 2023-08-07
Docfixes related to thread services


smadbe:
1) Question: why "token" not required in `threadGet`? 

There might be a reason but it seems not obvious as I cannot guess it immediately. It may make sense to add the reason in the service description, otherwise I don't know what to do with not-set `token`.

2) `threadUpdate`

![Screenshot 2023-08-07 at 12 33 29](https://github.com/France-ioi/AlgoreaBackend/assets/1053150/63ea9b8d-01a8-4202-af2f-e7128f9cd2a8)

`ItemID` and `ParticipantID` does not follow the convention of camel case for the backend.  As there are given in the url, I assume the itemID and participantID do not have to be given in body anyway I assume.

3. In `threadGet`, I would find useful to describe what is in the token.

4. Some services have `can_request_help_to` `appears only in services that support this permission (see service description), and only if it is set`. I would expect the value to be `null` when it is set but the permission not given to the user. It is not clear if it is valid value as the field is not indicated as `nullable`. Anyway, shouldn't it be an array as we cannot aggregate groups? 

5. `threadGet`: there is probably a line break missing in `the thread hasn't been started Restrictions:` 

6. In the swagger doc, the "thread" section disappear from the menu as soon as you click on another service, for instance [this one](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v1.10.0.html#tag/group-memberships/operation/groupRemoveMembers). Swagger bug or something bad in the metadata?

6. Unlike most other services,`threadGet`, `listThreads`, `threadUpdate` do not clearly state what happens if validations or restrictions are failing.

GeoffreyHuck:
PR here with all points but 4: https://github.com/France-ioi/AlgoreaBackend/pull/995

Regarding point 4,

4. Some services have `can_request_help_to` `appears only in services that support this permission (see service description), and only if it is set`. I would expect the value to be `null` when it is set but the permission not given to the user. It is not clear if it is valid value as the field is not indicated as `nullable`. Anyway, shouldn't it be an array as we cannot aggregate groups? 

Strictly speaking it should be an array. It's not because for now it has only been implemented without aggregation.

But this comes with an issue, with the current technique, we cannot represent a `null` permission differently than by removing the field from the answer. So we cannot distinguish whether there is no permission (field is null because there is no permission), or whether the service doesn't support this permission (field is null because the permission is not set).

[...]

PR #995, opened 2023-08-18
Docfixes related to thread services

GeoffreyHuck:
Related to #988 

DONE: Just point 6 would require to find out why it happens because I don't have the issue in local.

- [x] 1. Question: why "token" not required in threadGet?
There might be a reason but it seems not obvious as I cannot guess it immediately. It may make sense to add the reason in the service description, otherwise I don't know what to do with not-set token.

> FIXED. The field is always present so it should be required.

- [x] 2. threadUpdate
ItemID and ParticipantID does not follow the convention of camel case for the backend. As there are given in the url, I assume the itemID and participantID do not have to be given in body anyway I assume.

> FIXED. Those fields were copied in the structure for code factorization reasons and weren't intended to appear in the doc. They're now removed.

- [x] 3. In threadGet, I would find useful to describe what is in the token.

> FIXED. But with a little hack: Added a field named `token_not_present_only_for_doc` which shows the content in the documentation. This field never appear in the response, but has to be here for the doc. Didn't find another way.

- [x] 4. Some services have can_request_help_to appears only in services that support this permission (see service description), and only if it is set. I would expect the value to be null when it is set but the permission not given to the user. It is not clear if it is valid value as the field is not indicated as nullable. Anyway, shouldn't it be an array as we cannot aggregate groups?

> We might consider that this point will be resolved when we implement aggregation of `can_request_help_to` (see comment: https://github.com/France-ioi/AlgoreaBackend/issues/988#issuecomment-1684024829)

- [x] 5. threadGet: there is probably a line break missing in the thread hasn't been started Restrictions:

> FIXED.

- [ ] 6. In the swagger doc, the "thread" section disappear from the menu as soon as you click on another service, for instance [this one](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v1.10.0.html#tag/group-memberships/operation/groupRemoveMembers). Swagger bug or something bad in the metadata?

> MAYBE. It works on my machine where I'm using the latest swagger. Upon examination, there is a slight difference in the CSS on the menu. Made a PR that upgrades swagger in CI, hopefully it'll resolve the issue: https://github.com/France-ioi/AlgoreaBackend/pull/996

- [x] 7. Unlike most other services,threadGet, listThreads, threadUpdate do not clearly state what happens if validations or restrictions are failing.

> FIXED.

### DIFF
--- a/app/api/threads/get_thread.go
+++ b/app/api/threads/get_thread.go
@@ -42,13 +42,14 @@ type threadGetResponse struct {
 //
 //
 //		Restrictions:
-//			* one of these conditions matches:
+//			* one of these conditions must match:
 //				- the current-user is the thread participant and allowed to "can_view >= content" the item
 //				- the current-user has the "can_watch >= answer" permission on the item
 //				- the following rules all matches:
 //					* the current-user is descendant of the thread helper_group
 //					* the thread is either open (=waiting_for_participant or =waiting_for_trainer), or closed for less than 2 weeks
 //					* the current-user has validated the item
+//			Otherwise, a forbidden error is returned.
 //
 //	parameters:
 //		- name: item_id
--- a/app/api/threads/get_thread.go
+++ b/app/api/threads/get_thread.go
@@ -8,21 +8,26 @@ import (
 	"github.com/go-chi/render"
 
 	"github.com/France-ioi/AlgoreaBackend/app/database"
+	"github.com/France-ioi/AlgoreaBackend/app/payloads"
 	"github.com/France-ioi/AlgoreaBackend/app/service"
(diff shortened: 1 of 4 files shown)


smadbe:
This was supposed to be done very quickly... it is difficult to work on the forum with all these "pending" fixes.  I see you started it in mid-August, one month later I think it is time to complete it. 

If some sub-requests requires some more time, just create a specific issue/PR so that this one can be merged.

smadbe (approved):
I am not fully convinced by: 
> FIXED. But with a little hack: Added a field named token_not_present_only_for_doc which shows the content in the documentation. This field never appear in the response, but has to be here for the doc. Didn't find another way.

But okay. Of course, I haven't checked the generated doc as well.

PR #997, opened 2023-08-21
Bug: Duplicate in thread listing service

### LINKED ISSUE #986: Bug: Duplicate in thread listing service
If I call
```
curl 'https://dev.algorea.org/api/threads?item_id=899800937596940136&is_mine=1'  -H 'Authorization: Bearer <usualtoken>' 
```
I get 8 times the same entry in the output.

Note: that's maybe because the thread has been set on the chapter, not a task ? But anyway we should maybe (not sure as it is not really critical) restrict the create of threads to tasks?

GeoffreyHuck:
Fixes #986 

The bug appeared when the asked item was in more than one parent chapter. Came from the fact we use a join on `item_ancestors` to get the item's children: the same thread appeared as many times as it was a `child_item_id` in `item_ancestors`.

To reproduce, added the ability to set more than one `parent` to an item in Gherkin definitions.

### FIX:
Added a `DISTINCT`.

### Also considered:
- Using a subquery to filter `item_id` with all its children instead of a join, but I think its less efficient

### DIFF
--- a/app/api/threads/list_threads.feature
+++ b/app/api/threads/list_threads.feature
@@ -197,17 +197,19 @@ Feature: List threads
   Scenario: Should return only thread from item or descendant when item_id is given
     Given I am @John
     And there are the following items:
-      | item              | parent      | type    |
-      | @Root_Task        |             | Task    |
-      | @Chapter1         |             | Chapter |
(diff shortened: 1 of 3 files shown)


smadbe (approved):
👍

Issue #838, opened 2022-03-30
Find solution to allow editing children when the propagation is too long


smadbe:
## Overview

It is very common to add a new chapter as a child to a chapter close to the public root items (for instance in item `4702`). This is currently not doable through the UI / curl as the propagation takes too long and the request times out... and so the change is reverted.

The only way to add such a chapter now is to add it by hand in the DB, mark the propagation and run the propagation by hand.. this is both painful and dangerous.

## Proposal

Propagation affecting other users such as adding a chapter in another chapter could be done in "background". There is "no reason" to wait for the end of the propagation to respond to the current user that the children list has been updated.

So you should mark the propagation as to be done, respond to the request as a success, and then launch the propagation "in background" (in a safe way) as this propagation does not take so much time for the moment (a few minutes for the longer).

Note that if "respond success and continue processing" is difficult with the current lib, a "easy" way to do it with our current deployment setup is to run processing [as a sql trigger in AWS Aurora](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraMySQL.Integrating.Lambda.html). In this case, you just have to define what needs to be called as executable file (for instance `AlgoreaBackend db-recompute` or a specific executable file).

## Acceptance criteria
- [x] Make a proposal before implementing it

GeoffreyHuck:
[Proposal not finished yet]

### Remark on transactions and propagation

Currently, the propagation is made in the same transaction as the creation/update of items. This can create deadlock issues because with InnoDB, **all the locks** that are acquired inside a transaction are only released at the end of the transaction (source: High Performance MySQL - O'Reilly). Thus, the more things we do inside a transaction, the more locks we acquire, and the higher the probability that we face a deadlock with other concurrent transactions.

When everything we want is updated but the propagation has not been made yet, the database is in a consistent state. The propagation is a transformation from a consistent state to another consistent state. So, we can end the transaction there, and start a new transaction for the propagation.

> We should run each propagation in its own transaction where only the propagation is done.

It also simplifies the implementation of async propagation.

### Notes on concurrent propagations

The propagation is time consuming and is inside a big transaction. It may acquire many locks until it completes. It may be possible to break it down into smaller chunks but for now we have no reason to assume we need to do that. We can consider this if we find that some transaction from another place that is not a propagation, deadlocks with a transaction of a propagation.

[...]

PR #999, opened 2023-08-23
Async propagation

GeoffreyHuck:
For issue #838

TODO:

- [x] add propagation for `updateItem`
- [x] be sure that it's the same propagation than `createItem`
- [x] remove the ancestors propagation from propagation and do it before calling the propagation endpoint
- [x] remove the table `propagations` and triggers
- [x] add endpoint to schedule propagation in the config
- [x] if the endpoint is specified, call it with parameters the type of propagation (array of string)
- [x] if the endpoint is not specified, do the propagation synchronously: test
- [x] timeout of the endpoint max 3s
- [x] make sure the propagation is scheduled: test
- [x] make the propagation synchronously if the endpoint returns an error
- [x] log an error if endpoint returns an error / is not reachable ; test that the error is logged

Note: there is no test for the timeout delay, not sure if it's good practice to test such things.

### For the deployment

- Requires to execute the `install` command
- Requires to add the config `database.aws_aurora_propagation_trigger`. And maybe update `.github/conf/config.lambda.yaml` ?

### Propagation process and choices

It we find that the propagations run big expensive queries even if there is nothing to do, we can start them with a check that a row with the right characteristics (`state=todo`, `state=to_be_propagated`,...) is present in the database.

**Important:** The Lambda Trigger will be called even if the transaction is rolled back. In this case, the propagation would do nothing and it shouldn't be a problem.

**Important 2:** Since the trigger is called before the completion of the transaction, it's possible that when the propagation is run, there is nothing to do because the transaction didn't finish yet.
Can we mitigate this?
- With a delay? But it would defeat the purpose of the trigger.
- We can assume that the delay to call and execute the propagation via Lambda is enough to cover the time it takes to execute any transaction. That may not be the case, but if the number of cases is small enough, it shouldn't be a problem.
  Moreover, as the system is used, there will be other propagations scheduled soon enough.
  
There are 4 types of propagations:
- `groups_ancestors`: sometimes, `results` or `permissions` propagation is called after this one
- `items_ancestors`: this one is always executed along with `permissions` and `results` propagations
- `permissions`: most of the time, `results` propagation is called after this one
- `results`

Combinations of different propagations are called in many services. We have a few options:
- We make 4 different propagations and trigger those which have to be called.
  This would model how the code is currently working, where only the required propagations are called manually.
  The catches are:
    * It is hard to tell which propagations has to be called after a service, because one type of propagation can require another in certain cases. It's easy to forget to call one. It's even possible that some services don't call all the required propagations.
- We bundle all 4 propagations together, and whenever we want to call the propagation, we call all four.
  This might be a little more expensive for the database, yet it has to be measured because a propagation always start by checking if there is any row that matches a constraint. Those checks can even be simplified. The advantage is that we cannot forget to call a specific propagation.

For now, since the issue happens only for `itemCreate`, we start by implementing the async propagation for `itemCreate` only. Since it is doing `items_ancestors`, `permissions` and `results` propagations, we bundle those 3 together.

### Propagation command

The command `propagation` have been added to run the propagation.

Something weird: When the propagation is called in the service, it takes 7s to complete.
When it is in a command: it takes 100ms.
I tried to put the propagation in another transaction in the service, but it still took 7s.
	   
The propagation uses a database lock to ensure it is not called concurrently.
**Issue:** I don't think it is possible to test locks.

**Issue 2:** Commands are currently not tested.

### Trigger:

When a propagation has to be run, an entry in table `propagation` has to be added. This calls a trigger that:
- Triggers AWS Lambda for propagation
- Set `propagation`=1 on the entry that was just added

**Pro:**
- We can easily test that the trigger has been called by checking that `propagation`=1
- If we need to, we can extend this system to schedule explicitly different types of propagation (groups, items, results, permissions),
  but this can be tricky before under some circumstances, we need to call other types of propagations after a propagation.

**Con:**
- The code needs to explicitly state that the propagation has to be run by adding a record in table `propagation`.
  We could have made the triggers in `results_propagation`, `items_propagation`, `groups_propagation` and `items_propagation`.
  But this would have resulted in more complexity, and it would have been difficult to make sure to run the propagation only once.

### `CheckConfig()`/`CreateMissingData()` integration test and refactoring

Those function makes sure that the database contains some mandatory rows in `groups` and `groups_groups` tables. They have been modified to make sure that there is always a row in table `propagation` with `id=1`. This is used so that we know there is a propagation to be made, and it is set in the same trigger that triggers AWS Lambda.

An issue encountered was that those functions used mocks instead of a database in their tests. Those tests were very hard to modify. Instead, integration tests with a real database were added because the probability that those functions will change again in the future is very high. Then, those functions had to be moved into another package: `configdb`, because a circular dependency issue (`app` -> `testhelpers` (to load the fixtures in the database) -> `app`) made it impossible to have an integration test in the `app` package.

**Note:** A bug was found with `CreateMissingData()` which was reported here: https://github.com/France-ioi/AlgoreaBackend/issues/1001

### Make sure the trigger is called only once

We want to make sure we don't call the trigger multiple times while the propagation has not been done yet. It is easy to implement this in the trigger with `IF OLD.propagate = 0 THEN ...`. The tricky part was to unit test this behavior. This was implemented by adding a counter in `propagations` table: `scheduled_counter`. It is incremented every time the propagation is scheduled when `propagate = 0`. The AWS Lambda trigger will be put in the same place, which guarantees it is not called more than needed.

### Call AWS Lambda Trigger

We need to have `lambda_async('arn:aws:lambda:REGION:ACCOUNT_ID:function:Propagate', '{}');` in our SQL Trigger. Which means we;
- Need to specify the REGION and ACCOUNT_ID somewhere, and maybe the name of the function called
- We need to make sure this line is not added in CI, Dev, or other places that are not an AWS Aurora database, otherwise we'll get a syntax error

**Solution retained:**
The SQL migration file contains a line `-- %%aws_aurora_propagation_trigger%%`. This will be replaced during the migration by the value in the configuration file `database.aws_aurora_propagation_trigger`. This way, we can disable the AWS trigger by making it empty `""`, or a comment `"--disabled"`.
In production, we put the value `"lambda_async('arn:aws:lambda:REGION:ACCOUNT_ID:function:Propagate', '{}');"`. It allows to configuration everything in the same configuration line.

**Con:** The implementation is ugly because the package we use for the migration, `sql-migrate`, doesn't support variables like this. So the function `ExecMax` that is called from the package has been copied into the code of the command, and modified where we replace the value. I don't know how we can make this cleaner.


smadbe (commented):
> For now, since the issue happens only for itemCreate, we start by implementing the async propagation for itemCreate only. Since it is doing items_ancestors, permissions and results propagations, we bundle those 3 together. 

This happens equally on `itemUpdate` when the `children` attribute given.

But is there a problem of doing this everywhere there is propagation ? 

I think the answer "yes", as doing an async propagation will probably delay the actual propagation, and so, as a consequence, we may have inconsistencies in the GET immediately following the update. 

However, the problem that we have is not with item ancestors as far as I know (and we should never have I think? don't you think?).  I think only the "results" propagation is the problem for us. 

My point is that for `itemCreate`/`itemUpdate`, we should probably have such an async system for scores, but not the item ancestors... otherwise we might have behavior which are complicated to explain for the user who just did the change. 

> Something weird: When the propagation is called in the service, it takes 7s to complete.
When it is in a command: it takes 100ms.

Wow, a 70x factor, it is annoying not to have the explanation.

> When a propagation has to be run, an entry in table propagation has to be added. This calls a trigger that:

Table using ends with a "s" in our db by conventions. Could we make it more explicit that it is for the propagation trigger system by calling it something like "propagation_triggers".

> When a propagation has to be run, an entry in table propagation has to be added. This calls a trigger that...
> ...
> Con:
>
>    The code needs to explicitly state that the propagation has to be run by adding a record in table propagation.
>    We could have made the triggers in results_propagation, items_propagation, groups_propagation and items_propagation.
>    But this would have resulted in more complexity, and it would have been difficult to make sure to run the propagation only once.


The "problem" I have with that fina solution is at the end, the SQL trigger, is ONLY used to call a lambda... but is it the best (easier/cleaner) way to just call a lambda? If it would have been transparent with a "todo" value in `results_propagation`, `items_propagation`, `groups_propagation` or `items_propagation` (as you say in your "cons"), the trigger was making sense. 
If it is not good to use the trigger in the `*_propagation` tables (which I understand), don't we come back to a case where it would be more "transparent" (with the benefit of being cloud-provider agnostic) to just do a ping on a http endpoint when a propagation needs to be done ? Then it is easier to use a basic queuing system to run it without duplication, with reports if it fails, ...

> Make sure the trigger is called only once
> We want to make sure we don't call the trigger multiple times while the propagation has not been done yet. It is easy to implement this in the trigger with IF OLD.propagate = 0 THEN .... The tricky part was to unit test this behavior. This was implemented by adding a counter in propagations table: scheduled_counter. It is incremented every time the propagation is scheduled when propagate = 0. The AWS Lambda trigger will be put in the same place, which guarantees it is not called more than needed.

Ok if you have a counter which is only supposed to be 0 or 1, but then you do not need the boolean next to it IMO.

> We need to have lambda_async('arn:aws:lambda:REGION:ACCOUNT_ID:function:Propagate', '{}'); in our SQL Trigger. 

The problem is also that it is not a config but kind hardcoded... if we change the function, changing the app config is not enough, we need to change the db schema... which is weird and error prone. (To come back to the http endpoint solution, it would be a config)

PR #1006, opened 2023-09-08
Get permission service: clean "can_request_help" usage

### LINKED ISSUE #1004: Get permission service: clean "can_request_help" usage
## Motivations

The current spec of the service does not really work how the "can_request_help" is designed. So let's fix it.

## Subtasks
- [x] in `computed` and all `granted_via_*`, `can_request_help_to` is aggregate of several groups, so return an array of the (id, is_all_users_group, name) tuple. This should be `[]` is there is no such groups.
- [x] in `granted`, the `can_request_help_to` is unique to keep (id, is_all_users_group, name). `can_request_help_to` should be null if there is no value set.
- [x] In both cases, if a group is not visible by the current user, the `name` should not be given, but the `id` or `is_all_users_group` should be given in any case.

GeoffreyHuck:
fixes #1004 

We now return an array of all groups with `can_request_help_to` permissions set in all `granted_via_*`.

In `computed`, we also return all permissions coming from parent items while `request_help_propagation` is set. The uses a recursive query that is only called once, `GetAncestorsRequestHelpPropagatedQuery()`.

### Format of `can_request_help_to` in response

- `id` always set
- `name` present only if it is visible
- `is_all_users_group` to `true` or `false` depending on the case 

### Performance notice:
- We use  the query `IsVisibleForGroup()` once for each group we find. So it might be expensive if we have cases where we have many groups `can_request_help_to` permissions set.

The typical query is the following:
```
SELECT 1 FROM `groups` 
WHERE (groups.is_public OR groups.id IN(
    SELECT groups_ancestors_active.ancestor_group_id
    FROM `groups_groups_active`
    JOIN groups_ancestors_active ON groups_ancestors_active.child_group_id = groups_groups_active.parent_group_id
    JOIN `groups` AS ancestor_group ON ancestor_group.id = groups_ancestors_active.ancestor_group_id
    WHERE (groups_groups_active.child_group_id = 5549855599337303585)
    AND (ancestor_group.type != 'ContestParticipants'
))
OR groups.id IN(
    SELECT ancestors_of_managed.ancestor_group_id FROM `groups_ancestors_active` 
    JOIN group_managers ON group_managers.group_id = `groups_ancestors_active`.ancestor_group_id 
    JOIN groups_ancestors_active AS group_ancestors ON group_ancestors.ancestor_group_id = group_managers.manager_id
        AND group_ancestors.child_group_id = 5549855599337303585
    JOIN `groups` ON groups.id = groups_ancestors_active.child_group_id 
    JOIN groups_ancestors_active AS ancestors_of_managed ON ancestors_of_managed.child_group_id = groups_ancestors_active.child_group_id
        AND (groups.type != 'User' OR ancestors_of_managed.is_self)
    JOIN `groups` AS ancestor_group ON ancestor_group.id = ancestors_of_managed.ancestor_group_id
    WHERE (ancestor_group.type != 'ContestParticipants')
))
AND (groups.id = 2646504854435585587)
LIMIT 1
```

**Note:** It might be easier to review commit by commit. Notes about choices are present in commit messages.

### DIFF
--- a/app/api/groups/get_permissions_can_request_help_to.feature
+++ b/app/api/groups/get_permissions_can_request_help_to.feature
@@ -11,12 +11,13 @@ Feature: Get permissions can_request_help_to for a group
   Background:
     Given allUsersGroup is defined as the group @AllUsers
     And there are the following groups:
-      | group              | parent             | members  |
-      | @AllUsers          |                    |          |
-      | @School            |                    | @Teacher |
-      | @ClassParentParent |                    |          |
-      | @ClassParent       | @ClassParentParent |          |
-      | @Class             | @ClassParent       |          |
+      | group                    | parent                   | members  |
+      | @AllUsers                |                          |          |
+      | @School                  |                          | @Teacher |
+      | @ClassParentParentParent |                          |          |
+      | @ClassParentParent       | @ClassParentParentParent |          |
+      | @ClassParent             | @ClassParentParent       |          |
+      | @Class                   | @ClassParent             |          |
     And @Teacher is a manager of the group @Class and can grant group access
     And there are the following tasks:
       | item  |
@@ -65,19 +66,20 @@ Feature: Get permissions can_request_help_to for a group
   Scenario: Should return can_request_help_to arrays when permissions with specific origins are defined
     Given I am @Teacher
     And there are the following item permissions:
-      | item  | group              | source_group | origin           | is_owner | can_request_help_to          | comment                          |
-      | @Item | @Class             | @Class       | self             | false    | @HelperGroupSelf1            |                                  |
-      | @Item | @ClassParent       | @Class       | self             | false    | @AllUsers                    |                                  |
-      | @Item | @ClassParentParent | @Class       | self             | false    |                              | check we don't get empty groups  |
-      | @Item | @Class             | @Class       | group_membership | false    | @HelperGroupGroupMembership1 | granted                          |
-      | @Item | @ClassParent       | @Class       | group_membership | false    | @HelperGroupGroupMembership2 | group_membership but not granted |
-      | @Item | @ClassParentParent | @Class       | group_membership | false    | @AllUsers                    | group_membership but not granted |
-      | @Item | @ClassParent       | @Class       | item_unlocking   | false    | @HelperGroupItemUnlocking    |                                  |
-      | @Item | @Class             | @Class       | item_unlocking   | false    | @AllUsers                    |                                  |
-      | @Item | @ClassParentParent | @Class       | item_unlocking   | false    | @HelperGroupNotVisible       | not visible                      |
-      | @Item | @Class             | @Class       | other            | false    | @AllUsers                    |                                  |
-      | @Item | @ClassParentParent | @Class       | other            | false    | @HelperGroupOther1           |                                  |
-      | @Item | @ClassParent       | @Class       | other            | false    | @HelperGroupOther2           |                                  |
+      | item  | group                    | source_group | origin           | is_owner | can_request_help_to          | comment                                     |
+      | @Item | @Class                   | @Class       | self             | false    | @HelperGroupSelf1            |                                             |
+      | @Item | @ClassParentParentParent | @Class       | self             | false    | @HelperGroupSelf1            | shouldn't contain duplicate of previous one |
+      | @Item | @ClassParent             | @Class       | self             | false    | @AllUsers                    |                                             |
+      | @Item | @ClassParentParent       | @Class       | self             | false    |                              | check we don't get empty groups             |
+      | @Item | @Class                   | @Class       | group_membership | false    | @HelperGroupGroupMembership1 | granted                                     |
+      | @Item | @ClassParent             | @Class       | group_membership | false    | @HelperGroupGroupMembership2 | group_membership but not granted            |
+      | @Item | @ClassParentParent       | @Class       | group_membership | false    | @AllUsers                    | group_membership but not granted            |
+      | @Item | @ClassParent             | @Class       | item_unlocking   | false    | @HelperGroupItemUnlocking    |                                             |
+      | @Item | @Class                   | @Class       | item_unlocking   | false    | @AllUsers                    |                                             |
+      | @Item | @ClassParentParent       | @Class       | item_unlocking   | false    | @HelperGroupNotVisible       | not visible                                 |
+      | @Item | @Class                   | @Class       | other            | false    | @AllUsers                    |                                             |
+      | @Item | @ClassParentParent       | @Class       | other            | false    | @HelperGroupOther1           |                                             |
+      | @Item | @ClassParent             | @Class       | other            | false    | @HelperGroupOther2           |                                             |
   # The following lines are to make the groups visible by @Teacher
   And the group @Teacher is a descendant of the group @HelperGroupSelf1
   And the group @Teacher is a descendant of the group @HelperGroupGroupMembership1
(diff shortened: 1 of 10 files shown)


app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -55,5 +69,124 @@ Feature: Get permissions can_request_help_to for a group
...
+  And the group @Teacher is a descendant of the group @HelperOtherSourceGroup2
+  And the group @Teacher is a descendant of the group @HelperOtherSourceGroup3
+  And the group @Teacher is a descendant of the group @HelperOtherSourceGroup4
+  When I send a GET request to "/groups/@Class/permissions/@Class/@Item"
+  Then the response code should be 200
+  And the response at $.granted_via_self.can_request_help_to[*] should be:
+    | id                       | name                          | is_all_users_group |
+    | @AllUsers                | AllUsers                      | true               |
+    | @HelperGroupSelf1        | Group HelperGroupSelf1        | false              |
+    | @HelperOtherSourceGroup1 | Group HelperOtherSourceGroup1 | false              |
```
smadbe, line 114: Just a remark: 
The response "checkers" (... should be) are usually JSON... even for an extract from json as here, it would make to have to have JSON. Is it on purpose to use a table instead? Typically, it is not clear (to me) if table may miss a column that we do not want to check.
I am fine with it but I was wondering it was on purpose.

app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -55,5 +69,124 @@ Feature: Get permissions can_request_help_to for a group
...
+  And the response at $.granted_via_group_membership.can_request_help_to[*] should be:
+    | id                           | name                              | is_all_users_group |
+    | @HelperGroupGroupMembership2 | Group HelperGroupGroupMembership2 | false              |
+    | @AllUsers                    | AllUsers                          | true               |
+    | @HelperOtherSourceGroup2     | Group HelperOtherSourceGroup2     | false              |
+  And the response at $.granted_via_item_unlocking.can_request_help_to[*] should be:
+    | id                        | name                           | is_all_users_group |
+    | @HelperGroupItemUnlocking | Group HelperGroupItemUnlocking | false              |
+    | @AllUsers                 | AllUsers                       | true               |
+    | @HelperGroupNotVisible    |                                | false              |
```
smadbe, line 127: This `name` value does not really validate if it is a omitted value, null or "". I would say it is is not just a detail as typically bugs hide in these kind of things.

app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -55,5 +69,124 @@ Feature: Get permissions can_request_help_to for a group
...
+    And the group @Teacher is a descendant of the group @HelperGroupSelf1
+    And the group @Teacher is a descendant of the group @HelperGroupGroupMembership1
+    And the group @Teacher is a descendant of the group @HelperGroupGroupMembership2
+    And the group @Teacher is a descendant of the group @HelperGroupItemUnlocking1
+    And the group @Teacher is a descendant of the group @HelperGroupItemUnlocking2
+    And the group @Teacher is a descendant of the group @HelperGroupOther1
+    And the group @Teacher is a descendant of the group @HelperGroupOther2
+    When I send a GET request to "/groups/@Class/permissions/@Class/@Item"
+    Then the response code should be 200
+    And the response at $.granted_via_self.can_request_help_to[*] should be ""
```
smadbe, line 178: `[*] should be ""` is a bit strange and difficult to interpret.
If it means "empty array", couldn't we have a rule like `And the response at $.granted_via_self.can_request_help_to should be an empty array` ?

app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -34,18 +46,20 @@ Feature: Get permissions can_request_help_to for a group
...
     And there is a group @HelperGroup
     And there are the following item permissions:
       | item  | group  | is_owner | can_request_help_to |
       | @Item | @Class | false    | @HelperGroup        |
     When I send a GET request to "/groups/@Class/permissions/@Class/@Item"
     Then the response code should be 200
-    And the response at $.granted.can_request_help_to should be "null"
+    And the response at $.granted.can_request_help_to should be:
+      | id           | name | is_all_users_group |
+      | @HelperGroup |      | false              |
```
smadbe, line 62: Again, this makes it difficult for me to check if this return a tuple (as requested) or just an array with one tuple (which would be more consistent with the other tables).

smadbe, line 62: I do not see (I may have missed it) example of `granted` returning a null as requested in #1004:

> in granted, the can_request_help_to is unique to keep (id, is_all_users_group, name). can_request_help_to should be null if there is no value set.

smadbe (changes requested):
You write in the PR: 

> ### Format of `can_request_help_to` in response
> 
> The fields are now always present:
> - `id` always set
> - `name` set or null if not visible
> - `is_all_users_group` to `true` or `false` depending on the case 

At other places of the API, when an attribute is not visible, it is omitted (not given)... `null` is reserved for cases when a value is visible but `null` (the db value).

GeoffreyHuck:
The ways to check for null, empty arrays and empty strings have been uniformed:
- <null> for null
- [] for empty array
- nothing for an empty string

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: fixed with `[]`, but it's also possible to create a feature for it

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: fixed, see comment below for how it is checked

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: Yes, it is the case for all tests defined in `get_permissions.feature`.

GeoffreyHuck: fixed with `<undefined>`, see comment below

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: This method of checking doesn't necessary include all the fields that are present. It is made so that we check only what we need.

The column names are JsonPaths themselves. So it's possible to check for nested objects like `group.name`.

It suppress the need to update all old tests when we have change on other properties, but it can't check that all the properties are tested.

app/api/groups/get_permissions_can_request_help_to.feature
smadbe: I think you didn't get my point. My question was : how can I know reading this if you checking for a `tuple<id, name, is_all_users_group>` or an array containing one tuple<id, name, is_all_users_group> ? What would be the different in the rule?

smadbe (changes requested):
Mainly questions actually...

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: The difference is in the JSONPath. To retrieve an array, there is a `[*]` at the end of the JSONPath:
- `$.granted.can_request_help_to` is an object
- `$.granted_via_membership.can_request_help_to[*]` is an array

app/api/groups/get_permissions_can_request_help_to.feature
```diff
@@ -55,5 +70,124 @@ Feature: Get permissions can_request_help_to for a group
...
+    And the group @Teacher is a descendant of the group @HelperGroupItemUnlocking1
+    And the group @Teacher is a descendant of the group @HelperGroupItemUnlocking2
+    And the group @Teacher is a descendant of the group @HelperGroupOther1
+    And the group @Teacher is a descendant of the group @HelperGroupOther2
+    When I send a GET request to "/groups/@Class/permissions/@Class/@Item"
+    Then the response code should be 200
+    And the response at $.granted_via_self.can_request_help_to[*] should be "[]"
+    And the response at $.granted.can_request_help_to should be "<null>"
+    And the response at $.granted_via_group_membership.can_request_help_to[*] should be "[]"
+    And the response at $.granted_via_item_unlocking.can_request_help_to[*] should be "[]"
```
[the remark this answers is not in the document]
GeoffreyHuck: fixed

app/api/groups/get_permissions_can_request_help_to.feature
GeoffreyHuck: fixed

smadbe (approved):

PR #1007, opened 2023-09-11
viewGrantedPermission: Add can_request_help_to into own structure for better doc

GeoffreyHuck:
Since we added `can_request_help_to`  in the `viewGrnatedPermission`, we had it in the field `permissions`of many services in the swagger generated documentation with the annotation "only available if the service description mentions it".

This PR fixes this. It now appears only in the relevant service.


smadbe (approved):
Of course, I cannot validate that the doc will be ok.

PR #1012, opened 2023-09-19
updateThread: when the user owns an item, he can request help to any visible group, including himself

### LINKED ISSUE #1011: Grant "request help" perm to owners
## Overview

With our current way of handling the `request_help_to` permission, using the `permissions_granted` table, I think (I might be wrong) we have forgotten users with `is_owner=1` on an item, which should imply that they can request help to anybody they see (as they could give any valid group value to `request_help_to` for themselves anyway).

Actually, what we want is just item owners to be able to request help to others. Following [forum permissions](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/), for that, we just need to allow them to open a thread with themselves as a participant. In practice, for that, I think we just need to adapt the service `threadUpdate`.

## Acceptance criteria
- [x] adapt `threadUpdate` service so that item owners are allowed to request help to any group visible to them and to open their thread.
- [x] check if the `threadGet` may need an update (I don't think so)
- [x] update the devdoc [forum permissions](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/) if needed

GeoffreyHuck:
fixes #1011 

A change has been made to `IsVisibleForGroup`: a group can now see himself. This wasn't the case before, it could only see its ancestors, but I assume it should have been.

No update required for `getThread`.

### DIFF
--- a/app/api/threads/update_thread.feature
+++ b/app/api/threads/update_thread.feature
@@ -6,6 +6,7 @@ Feature: Update thread
       | 2  | manager | User  |
       | 3  | jack    | User  |
       | 4  | jess    | User  |
+      | 5  | owner   | User  |
       | 10 | Class   | Class |
       | 11 | School  | Class |
       | 12 | Region  | Class |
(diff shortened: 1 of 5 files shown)


smadbe (approved):
> A change has been made to IsVisibleForGroup: a group can now see himself. This wasn't the case before, it could only see its ancestors, but I assume it should have been.

A group is an ancestor of himself so it should have been the case already (anyway that's probably good to have no sql for this case). Is it on purpose that no test have been added for that case?

GeoffreyHuck:
Added tests:
- to check a group is visible to itself
- and a robustness test to verify we check the group we ask request help is visible when we are owner of the item

PR #1013, opened 2023-09-25
getItem: rename can_request_help_to in can_request_help in response

GeoffreyHuck:
So it is more clear that it is a bool.


smadbe (approved):
👍

PR #1015, opened 2023-10-02
getItem: can_request_help should be true when is_owner=1

### LINKED ISSUE #1014: Bug: GetItem returns "can_request_help": false for owners
```
curl 'https://dev.algorea.org/api/items/3327328786400474746 -H 'Authorization: Bearer <usualtoken>'
```
returns for `permissions`:
```
{
		"can_view": "solution",
		"can_grant_view": "solution_with_grant",
		"can_watch": "answer_with_grant",
		"can_edit": "all_with_grant",
		"is_owner": true,
		"can_request_help": false
}
```
All perms are at their max due to `is_owner`, except `can_request_help` which is `false`.

Note : Should also work when an ancestor has `is_owner` and propagation allows `can_request_help` to propagate. (to be tested)

GeoffreyHuck:
fixes #1014 

Covers for both `permissions` and `watched_group.permissions`:
- When `is_owner=1` for the group, on the item
- When `is_owner=1` for the group, on a parent item with propagation (`true`) and without propagation (`false`)
- When `is_owner=1` for a parent group, on the item
- When `is_owner=1` for a parent group, on a parent item with propagation (`true`) and without propagation (`false`)

### DIFF
--- a/app/api/items/get_item_can_request_help_to_groups.feature
+++ b/app/api/items/get_item_can_request_help_to_groups.feature
@@ -9,38 +9,58 @@ Feature: Get permissions can_request_help for an item
       | @Class              | @School | @Student,@HelperGroup |
     And @Teacher is a manager of the group @Class and can watch its members
     And there are the following items:
-      | item                 | type    |
-      | @Chapter1            | Chapter |
-      | @Chapter2            | Chapter |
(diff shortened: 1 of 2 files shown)


smadbe (approved):
looks good 👍

PR #1020, opened 2023-12-14
Fix itemActivityLogForAllItems perf issue: add index and rewrite

### LINKED ISSUE #987: Analyze log service perf
`time curl 'https://dev.algorea.org/api/items/log?limit=20&watched_group_id=917454091139548680' -H 'Authorization: Bearer <usualtoken>'` takes over 8sec to respond.

Analyze (don't change yet!) what could improve the perf of this service (could be a change of the way the frontend call it). Don't spend too much time on details yet.

GeoffreyHuck:
fixes #987 

### Context

The service `itemActivityLogForAllItems` had a performance issue. It was coming from the big query that retrieves the events: result_started, submission, result_validated, saved_answer, current_answer.

The big query is comprised of 3 subqueries that are union together.

1 subquery works with table `answers`, and the 2 others with table `results`.

The subquery with table `answers` was at fault, the problem is that it had an `ORDER BY` that wasn't indexed.

The 2 subqueries with table `results` already have the right index, so it should be fine for the future.

The subquery with table `answers` had to be rewritten because it used a `CASE WHEN...` in its `ORDER BY` clause. 

### Changes to the results returned

**This fix changes the order of the result, but it shouldn't matter** because it's on a column that was used as a tie breaker: `activity_type`.

### How it worked Before

The `activity_type` was passed as a `string`, then converted into an `integer`, treated in the query as `integer`, and transformed back to a `string` in the result returned by the query.

### How it works After

We keep the `activity_type` has as a `string` all the way. We just transform `current` to `current_answer` and `saved` to `saved_answer` in the result to keep the old names.

Note: it's not clear why an internal integer was used as I can't see any benefits. It could be a relic of something in the past.

### Add index

A migration was added to add the index on table `answers`.

### Tests updated

The tests were updated to make them clearer. The process is explained in the git commit messages. Basically, the dates were simplified, and the ids were reordered to be consistent with the dates.

### Added utils function `Contains`

Added a utility function to check if a slice contains a string. This function exists in go1.21 but we're using go1.20, added a comment to keep the info.

### Future perf

- An `IN()` operator is used with the **number of visible items of the user**
- An `IN()` operator is used with **all the participants that are descendant of the watched group**

-> Those two could be problematic as those number get bigger. It seems Dimitry added something that could be related:
`STRAIGHT_JOIN /* tell the optimizer we don't want to convert IN(...) into JOIN */`
Maybe it is to optimize this case, though I'm not sure, and it's hard to tell at which number it'll be problematic.

- For the `results` table, it should be fine because the right index is set.


app/api/items/get_activity_log.feature
```diff
@@ -29,58 +29,58 @@ Feature: Get activity log
...
-      | 0          | 200     | 11             | 2017-05-29 06:38:38 | 2017-05-29 06:38:38 | 2020-05-29 06:38:38  |
-      | 0          | 200     | 30             | 2017-05-29 06:38:00 | 2017-05-30 12:00:00 | 2020-05-29 06:38:38  |
-      | 0          | 201     | 11             | 2017-05-29 06:38:00 | null                | 2020-05-29 06:38:38  |
-      | 0          | 202     | 11             | 2017-05-29 06:38:00 | 2017-05-30 12:00:00 | 2020-05-29 06:38:38  |
-      | 0          | 203     | 11             | 2017-05-29 06:38:00 | 2017-05-30 12:00:00 | 2020-05-29 06:38:38  |
-      | 1          | 200     | 11             | 2017-05-29 06:38:00 | 2017-05-30 12:00:00 | 2020-05-29 06:38:38  |
-      | 1          | 200     | 31             | 2017-05-29 06:38:00 | 2017-05-30 12:00:00 | 2020-05-29 06:38:38  |
-      | 1          | 200     | 41             | 2016-05-29 06:38:00 | 2016-05-30 12:00:00 | 2020-05-29 06:38:38  |
-      | 1          | 201     | 11             | null                | 2017-05-29 06:38:38 | 2020-05-29 06:38:38  |
-      | 1          | 202     | 11             | 2017-05-29 06:38:00 | 2017-05-30 12:00:00 | 2020-05-29 06:38:38  |
```
smadbe, line 41: `2020-05-29 06:38:38` 🤣. The former dates were insane.

smadbe (approved):
looks good 👍

PR #1024, opened 2024-01-25
Search: do a "AND" on words

### LINKED ISSUE #1023: Search: do a "AND" on words
## Motivations
Currently if I search for "ranger", I got the item named "ranger des billes": 
```
curl 'https://dev.algorea.org/api/items/search?search=ranger&limit=11&types_include=Chapter,Task,Task'  -H 'Authorization: ...' 
```
but if I look for "ranger billes", I get an empty result:
```
curl 'https://dev.algorea.org/api/items/search?search=ranger%20billes&limit=11&types_include=Chapter,Task,Task'  -H 'Authorization: ...' 
```

## Subtasks
* make sure a "AND" is applied between words for the search request
* check/fix if the group search does not suffer from the same problem

GeoffreyHuck:
fixes #1023 

The following searching services have been updated.
- `itemSearch`
- `groupsJoinableSearch`
- `groupsPossibleSubgroupsSearch`

### Modifications

- The search string is now exploded into words, and each word has to be found for the item or group to be returned.

Note: we use `LIKE %?%`, so that the words has to be found in any place of the field (ie. "the" matches in "ano**the**r").

- The construction of the queries have been refactored to make the services easier to read.

- The conditions for the search have been refactored in the function `WhereSearchStringMatches` to make sure we use the same logic for all the search queries.

### DIFF
--- a/app/api/currentuser/search_for_available_groups.feature
+++ b/app/api/currentuser/search_for_available_groups.feature
@@ -65,6 +65,28 @@ Feature: Search for groups available to the current user
     ]
     """
 
+  Scenario: Should treat the words in the search string as "AND"
+    Given I am the user with id "21"
+    When I send a GET request to "/current-user/available-groups?search=the%20team"
+    Then the response code should be 200
+    And the response body should be, in JSON:
+    """
+    [
+      {
+        "id": "2",
+        "name": "(the) Our Team ___",
+        "description": null,
+        "type": "Team"
+      },
+      {
+        "id": "7",
+        "name": "Another %%%Team",
+        "description": "Another team group",
+        "type": "Team"
+      }
+    ]
+    """
+
   Scenario: Search for groups with "the", start from the second row
     Given I am the user with id "21"
     When I send a GET request to "/current-user/available-groups?search=the&from.id=2"
--- a/app/api/currentuser/search_for_available_groups.feature
+++ b/app/api/currentuser/search_for_available_groups.feature
@@ -1,20 +1,20 @@
 Feature: Search for groups available to the current user
   Background:
     Given the database has the following table 'groups':
-      | id | type                | name                                      | description            | is_public |
-      | 1  | Class               | (the) Our Class                           | Our class group        | 1         |
-      | 2  | Team                | (the) Our Team ___                        | null                   | 1         |
-      | 3  | Club                | (the) Our Club                            | Our club group         | 1         |
-      | 4  | Friends             | (the) \|\|\|Our Friends \\\\\\%\\\\%\\ :) | Group for our friends  | 1         |
-      | 5  | Other               | Other people                              | Group for other people | 1         |
-      | 6  | Class               | Another Class                             | Another class group    | 1         |
-      | 7  | Team                | Another %%%Team                           | Another team group     | 1         |
-      | 8  | Club                | Another %%%Club                           | Another club group     | 1         |
-      | 9  | Friends             | Some other friends                        | Another friends group  | 1         |
-      | 10 | Class               | The third class                           | The third class        | 1         |
-      | 11 | User                | Another %%%User                           | Another user group     | 1         |
-      | 21 | User                | (the) user self                           |                        | 0         |
-      | 22 | ContestParticipants | the                                       |                        | 1         |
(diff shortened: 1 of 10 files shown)


smadbe (commented):
Could you add in the service spec how the search terms are processed?

Are you confident it will not strongly negatively alter the service performances?

smadbe (approved):

GeoffreyHuck:
Added the spec in the description.

For the performance, Dimitri added a note in the description of groupsJoinableSearch:

//		Note: The current implementation may be very slow because it uses `LIKE` with a percentage wildcard
//		at the beginning. This causes MySQL to explore every row having `is_public`=1. Moreover, actually
//		it has to examine every row of the `groups` table since there is no index for the `is_public` column.
//		But since there are not too many groups and the result rows count is limited, the search works almost well.

It might be a problem because now we do X LIKE for each search, where X is the number of words in the search query.

Only a test made on the database with enough entries would show if performance is a problem. If it is, there are two alternatives.

### Alternative 1: Full-Text index in MySQL

Implementation: https://www.w3resource.com/mysql/mysql-full-text-search-functions.php

Pro:
* Simple to use. It would require a database migration to add the index, and then the request is very straightforward: `SELECT ... WHERE MATCH(title) AGAINST('search string words -exclude')`. It also support the exclusion of a word.
Con:
* Cannot handle complex search, but it should be enough in our case
* It's a proprietary MySQL 8.0 feature. Using it on other databases would require adaptations.

### Alternative 2: Using an external search index

Pro:
* More powerful and efficient than full-text index
Con:
* We need to have an instance of the search software
* We need to keep the index in sync

smadbe:
> ### Alternative 1: Full-Text index in MySQL

It may be useful to first check that RDS Aurora mysql 8 compatible supports this feature.


> ### Alternative 2: Using an external search index

Our searches use non-trivial access right rules that would be complicated to sync in such an engine in real time.

GeoffreyHuck:
It is compatible: `Amazon Aurora MySQL-Compatible Edition (Aurora MySQL) supports all the native full-text capabilities of MySQL InnoDB full-text indexes.` (from https://docs.aws.amazon.com/dms/latest/sql-server-to-aurora-mysql-migration-playbook/chap-sql-server-aurora-mysql.tsql.fulltextsearch.html in section "MySQL").

Should we implement it in this PR?

smadbe:
Yes, that sounds good to use full-text indexes for item and group search. Go for it.

GeoffreyHuck:
Did a few tries with `FULLTEXT INDEX` and found some limitations:
- it has trouble with special characters, and there don't seem to be a way around it: https://stackoverflow.com/questions/290581/handling-mysql-full-text-special-characters . Searches like `C++`, `C`, `C#` and the likes might not be possible, unless we do some very strange things like adding a column with special characters encoded as alpha characters. I tried a few special characters like `%`, '(`, `)`, it just doesn't work at all.
- we can only do a partial search with the ending of the word: `friend*` will return `Friends` but we cannot do `*riends`. This might not be a big issue.

Not sure this solution is good enough in practice.

So we might want to use an external search engine, but it'll add some complexity, or to accept that the previous solution might perform poorly in some situations.

GeoffreyHuck:
`FULLTEXT` search implemented.

A few notes:
- The search ignores stop words https://dev.mysql.com/doc/refman/8.4/en/fulltext-stopwords.html (it seems the list can be configured)
- The search only take into account words than are >= 4 characters by default but it can be updated: https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_ft_min_word_len
- There are special characters that can be used to enhance the search: https://www.mysqltutorial.org/mysql-full-text-search/mysql-boolean-full-text-searches/ (first I wanted to remove them, but I don't see any reason to disallow that)
- It's not possible to search special characters with full text search
- The rule implemented is to make all the words appearing in the search string mandatory, and all the words are searched as "word starts with"
- For items, we search on the title of the user locale if there is one, or the default language. Before, `COALESCE` was used, but that's not possible with the syntax of the search, so the query is a little bit more complex.

smadbe (approved):
:+1:  Let's deploy and we'll see if it works well for our needs

PR #1027, opened 2024-02-14
itemActivityLogForItem & itemActivityLogForAllItems: Add can_watch_item_answer in response: whether the current user can watch the answer

### LINKED ISSUE #1025: Add watch item permission in the itemActivityLogForAllItems service
## Motivations
The way the itemActivityLogForAllItems service is used in frontend, we want the user to be able to view the answer which may be listed on a "row". The problem is that are guaranteed that the current user has "can_watch >= result" for the item while we need to know if the user  "can_watch >= answer" to display the "view answer" link.  
This is only valid when watching another user as we can always watch our own answers.

Edit: Hum that's the same for `itemActivityLogForItem` as we return "events" for all descendant items... but the current user may not have the same permissions on the descendant items

## Subtasks
In itemActivityLogForAllItems:
* When the user is observing (`watched_group_id` given), add a way for the frontend to know whether the current user can watch the answers of each entry or not (return `can_watch` value,  or a boolean, or we may also wonder if returning "answer_id" makes sense if we cannot view it.. but probably it is easier to keep it as it is used for pagination)
In itemActivityLogForItem:
* same

GeoffreyHuck:
fixes #1025 

Added boolean `can_watch_item_answer`.

Tests have been updated. Since there were already cases where the user can and cannot watch the answer, there is no new test.

I first considered a field `item.permissions.can_watch` with the value of the permission, but this would have required more work for the implementation. If we only want to know whether or not we can watch the answer, the boolean is simpler.

### DIFF
--- a/app/api/items/get_activity_log.feature
+++ b/app/api/items/get_activity_log.feature
@@ -91,7 +91,7 @@ Feature: Get activity log
     And the database has the following table 'permissions_generated':
       | group_id | item_id | can_view_generated | can_watch_generated |
       | 20       | 200     | none               | result              |
-      | 21       | 200     | info               | none                |
+      | 21       | 200     | info               | none                | # user 21 is in group 20, so he has can_watch=result on item 200
       | 21       | 201     | info               | result              |
       | 21       | 202     | info               | result              |
       | 21       | 203     | none               | result              |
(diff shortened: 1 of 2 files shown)


smadbe (approved):
Looks good 👍  thx

PR #1028, opened 2024-02-16
Add "isEmpty" info to group member services

### LINKED ISSUE #1026: Add "isEmpty" info to group member services
## Motivations
When we list group members (groups), the user can select a few of these subgroups to remove the them from the group. If these subgroups are empty, we want to ask the user if he wants to delete these groups as well.

So we need to know if subgroups are empty.

## Subtasks
In `/groups/{group_id}/children`:
- add in output `is_empty` which is true only if the group is empty, so has not any child groups or users.
- that info should be returned only if the current user is a manager

GeoffreyHuck:
fixes #1026 

Didn't add tests because the different cases were already covered:
* When the user is not a manager of the group, `is_empty` isn't in the response
* When the user is a manager and the group is empty: `is_empty=true`,
* When the user is a manager and the group is not empty: `is_empty=false`

Improved the indentation of the query to make the indent clearer, specifically the use of `IF()`.

In the response, `IsEmpty` is a pointer to allow us to have 3 values:
* `null` -> don't include it in the response, when the user is not a manager of the group
* `true/false` -> when the user is a manager

The SQL query uses `IF()` to compute `is_empty` only when the user is a manager.

### DIFF
--- a/app/api/groups/get_children.feature
+++ b/app/api/groups/get_children.feature
@@ -1,3 +1,4 @@
+@wip
 Feature: Get group children (groupChildrenView)
   Background:
     Given the database has the following table 'groups':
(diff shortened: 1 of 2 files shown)


app/api/groups/get_children.feature
smadbe, line 1: From what I see (I may have missed one case), tests are all `user_count: 0` -> `"is_empty": true` and `user_count: 1` ->  `"is_empty": false`. I think it would be useful to have at least one case where `"is_empty": false` with `user_count: 0`, i.e. where the group has one child group but no user.

app/api/groups/get_children.go
```diff
@@ -32,6 +32,8 @@ type groupChildrenViewResponseRow struct {
 	IsOpen bool `json:"is_open"`
 	// required:true
 	IsPublic bool `json:"is_public"`
+	// Only returned if the current user is a manager of the group.
+	IsEmpty *bool `json:"is_empty,omitempty"`
```
smadbe, line 36: Maybe adding what it means. `Whether the group has no child group (incl. users)`

smadbe (changes requested):

app/api/groups/get_children.go
GeoffreyHuck: fixed.

app/api/groups/get_children.feature
GeoffreyHuck: fixed. Test added.

smadbe (approved):
👍

Issue #1022, opened 2024-01-18
Allow multiple session in parallel 


smadbe:
## Motivations
Currently, if you log in on the platform a device, and then log in on another device, you get disconnected from the first one. This is not really convenient. We should try to improve this process without comprising the security.

## Subtasks
* When re-doing the authentication process, the other sessions should not be dropped if still valid
* (if not already the case) When a user refresh a token, the former token (used to request the refresh) should have its lifetime reduced (to ~2min?) so that we do not end up with too many valid token for a same user at a same time.
* We should make sure that expired tokens always get flushed at a moment so that the DB does not keep growing (prefer cleaning while we create new tokens instead of requiring a cron job)

GeoffreyHuck:
Observation: The lifetime is currently not reduced for the old token. It has to be added.

### TLDR:
1. What happens if a token has been refreshed, but the old one is used to refresh it again, within the 2min? It may happen if more than one tab is open. **Proposition**: send back the new valid token
2. We need to add a column `sessions`.`old_access_token`, which contains the `access_token` from which we obtained this one during a refresh. This makes it easy to:
  - Find the new access token for the process in point `1.`
  - Remove the old session
3. Someone could open many sessions in parallel without refreshing the token, by cleaning the browser, or programmatically. This can add many rows in `sessions`. **Proposition**: limit the number of parallel sessions to 10.


## The details:

Here's a review on the process of auth & refreshing the token, and what should be modified:

### Current process: when a user is authenticated:
- Add a row in table `sessions` with the new `access_token`
- Update `refresh_tokens` of the user in database

### Current process: when the token is refreshed:
- Back-end calls the Login Service with the `refresh_token` currently in database, and gets a new `access_token` and `refresh_token`
- Deletes all `sessions` of `user_id` with `access_token` != old access token
- We add a row in table `sessions` with the new access token
- Update the `refresh_token` in database

### Current process: notes on `sessions`
- At any point, we have a maximum of 2 entries for a `user_id`: the current `access_token`, and the old one used when the token was refreshed.
- Only the refresh of the token leads to the deletion of entries in `sessions`


### New process with multiple sessions
- `expires_in` is here to simplify the example, the database contains the actual date of expiration.
- **Proposition**: `old_access_token` will be added in the `sessions` table, so we know what rows we need to delete during a refresh token.

Assume we already have 2 entries for a session open on a desktop (A):
```
| access_token | expires_in | old_access_token |
[...]

PR #1030, opened 2024-02-22
GherkinTests: Set reference to the database value if the reference is not set

GeoffreyHuck:
### Context:
* We need to update tests related to authentication to implement the new database schema (https://github.com/France-ioi/AlgoreaBackend/issues/1022#issuecomment-1952161599)
* Problem, the tests look like this:
```
    Then the response code should be 201
      """
      {
        "success": true,
        "message": "created",
        "data": {
          <token_in_data>
          "expires_in": 31622400
        }
      }
      """
    And the response header "Set-Cookie" should be "<expected_cookie>"
    And the table "users" should be:
      | group_id            | latest_login_at     | latest_activity_at  | temp_user | registered_at       | latest_profile_sync_at | login_id  | login    | email                | first_name | last_name | student_id | country_code | birth_date | graduation_year | grade | address | zipcode | city | land_line_number | cell_phone_number | default_language | free_text           | web_site                      | sex  | email_verified | last_ip   | time_zone      | notify_news | photo_autoload | public_first_name | public_last_name |
      | 5577006791947779410 | 2019-07-16 22:02:28 | 2019-07-16 22:02:28 | 0         | 2019-07-16 22:02:28 | 2019-07-16 22:02:28    | 100000001 | mohammed | mohammedam@gmail.com | Mohammed   | Amrani    | 123456789  | dz           | 2000-07-02 | 2020            | 0     | null    | null    | null | null             | null              | en               | I'm Mohammed Amrani | http://mohammed.freepages.com | Male | 0              | 127.0.0.1 | Africa/Algiers | true        | true           | true              | true             |
    And the table "groups" should be:
      | id                  | name                                 | type  | description | ABS(TIMESTAMPDIFF(SECOND, NOW(), created_at)) < 3 | is_open | send_emails | text_id                                 |
      | 2                   | AllUsers                             | Base  | null        | false                                             | false   | false       | null                                    |
      | 4                   | TempUsers                            | Base  | null        | false                                             | false   | false       | null                                    |
      | 4037200794235010051 | Example badges                       | Other | null        | true                                              | false   | false       | https://badges.example.com/             |
      | 5577006791947779410 | mohammed                             | User  | mohammed    | true                                              | false   | false       | null                                    |
      | 6129484611666145821 | Example badges with multiple parents | Other | null        | true                                              | false   | false       | https://badges.example.com/parents      |
      | 8674665223082153551 | Example #1                           | Other | null        | true                                              | false   | false       | https://badges.example.com/examples/one |
    And the table "groups_groups" should be:
      | parent_group_id     | child_group_id      |
      | 2                   | 4                   |
      | 2                   | 5577006791947779410 |
      | 4037200794235010051 | 6129484611666145821 |
      | 6129484611666145821 | 8674665223082153551 |
      | 8674665223082153551 | 5577006791947779410 |
      And the table "groups_ancestors" should be:
        | ancestor_group_id   | child_group_id      | is_self |
        | 2                   | 2                   | true    |
        | 2                   | 4                   | false   |
        | 2                   | 5577006791947779410 | false   |
        | 4                   | 4                   | true    |
        | 4037200794235010051 | 4037200794235010051 | true    |
        | 4037200794235010051 | 5577006791947779410 | false   |
        | 4037200794235010051 | 6129484611666145821 | false   |
        | 4037200794235010051 | 8674665223082153551 | false   |
        | 5577006791947779410 | 5577006791947779410 | true    |
        | 6129484611666145821 | 5577006791947779410 | false   |
        | 6129484611666145821 | 6129484611666145821 | true    |
        | 6129484611666145821 | 8674665223082153551 | false   |
        | 8674665223082153551 | 5577006791947779410 | false   |
        | 8674665223082153551 | 8674665223082153551 | true    |
      And the table "attempts" should be:
        | participant_id      | id | creator_id          | ABS(TIMESTAMPDIFF(SECOND, NOW(), created_at)) < 3 | parent_attempt_id | root_item_id |
      | 5577006791947779410 | 0  | 5577006791947779410 | true                                              | null              | null         |
    And the table "group_membership_changes" should be:
      | group_id            | member_id           | ABS(TIMESTAMPDIFF(SECOND, NOW(), at)) < 3 | action          | initiator_id        |
      | 8674665223082153551 | 5577006791947779410 | true                                      | joined_by_badge | 5577006791947779410 |
    And the table "sessions" should be:
      | session_id | user_id             | refresh_token                |
      | 1          | 5577006791947779410 | {{refresh_token_from_oauth}} |
    And the table "access_tokens" should be:
      | session_id | token                       | expires_at          | issued_at           |
      | 1          | {{access_token_from_oauth}} | 2020-07-16 22:02:28 | 2019-07-16 22:02:28 |
    And the table "group_managers" should be:
      | group_id            | manager_id          | can_manage  | can_grant_group_access | can_watch_members | can_edit_personal_info |
      | 4037200794235010051 | 5577006791947779410 | memberships | true                   | true              | false                  |
```

All IDs are hard coded.

They are generated using a random generator. Since the tests set the same SEED for the random generator, the output is deterministic.
The issue is that the changes we implement for the database change the order of how those IDs are generated. So it makes all IDs checks fail.

### Proposition:

* We re-use the reference system. The rule is: if we check the database results with a **non-defined reference**, we set the reference to the value we find. Then, if we check against the same reference, we verify that we find the same value.

This way, we don't need to match against the numbers anymore, we only care that IDs are similar where they have to be the same.

**This makes the tests less brittle.**

### Implementation

* This PR implement this change. Only one test was converted to use references to validate it works.

* There's also a refactor of the function `matchingBooleanValue`, to pass lint tests of cyclomatic complexity.

### DIFF
--- a/app/api/auth/create_access_token.feature
+++ b/app/api/auth/create_access_token.feature
@@ -86,54 +86,54 @@ Feature: Create an access token
       """
     And the response header "Set-Cookie" should be "<expected_cookie>"
     And the table "users" should be:
-      | group_id            | latest_login_at     | latest_activity_at  | temp_user | registered_at       | latest_profile_sync_at | login_id  | login    | email                | first_name | last_name | student_id | country_code | birth_date | graduation_year | grade | address | zipcode | city | land_line_number | cell_phone_number | default_language | free_text           | web_site                      | sex  | email_verified | last_ip   | time_zone      | notify_news | photo_autoload | public_first_name | public_last_name |
-      | 5577006791947779410 | 2019-07-16 22:02:28 | 2019-07-16 22:02:28 | 0         | 2019-07-16 22:02:28 | 2019-07-16 22:02:28    | 100000001 | mohammed | mohammedam@gmail.com | Mohammed   | Amrani    | 123456789  | dz           | 2000-07-02 | 2020            | 0     | null    | null    | null | null             | null              | en               | I'm Mohammed Amrani | http://mohammed.freepages.com | Male | 0              | 127.0.0.1 | Africa/Algiers | true        | true           | true              | true             |
+      | group_id      | latest_login_at     | latest_activity_at  | temp_user | registered_at       | latest_profile_sync_at | login_id  | login    | email                | first_name | last_name | student_id | country_code | birth_date | graduation_year | grade | address | zipcode | city | land_line_number | cell_phone_number | default_language | free_text           | web_site                      | sex  | email_verified | last_ip   | time_zone      | notify_news | photo_autoload | public_first_name | public_last_name |
+      | @UserMohammed | 2019-07-16 22:02:28 | 2019-07-16 22:02:28 | 0         | 2019-07-16 22:02:28 | 2019-07-16 22:02:28    | 100000001 | mohammed | mohammedam@gmail.com | Mohammed   | Amrani    | 123456789  | dz           | 2000-07-02 | 2020            | 0     | null    | null    | null | null             | null              | en               | I'm Mohammed Amrani | http://mohammed.freepages.com | Male | 0              | 127.0.0.1 | Africa/Algiers | true        | true           | true              | true             |
     And the table "groups" should be:
-      | id                  | name                                 | type  | description | ABS(TIMESTAMPDIFF(SECOND, NOW(), created_at)) < 3 | is_open | send_emails | text_id                                 |
-      | 2                   | AllUsers                             | Base  | null        | false                                             | false   | false       | null                                    |
-      | 4                   | TempUsers                            | Base  | null        | false                                             | false   | false       | null                                    |
-      | 4037200794235010051 | Example badges                       | Other | null        | true                                              | false   | false       | https://badges.example.com/             |
-      | 5577006791947779410 | mohammed                             | User  | mohammed    | true                                              | false   | false       | null                                    |
-      | 6129484611666145821 | Example badges with multiple parents | Other | null        | true                                              | false   | false       | https://badges.example.com/parents      |
-      | 8674665223082153551 | Example #1                           | Other | null        | true                                              | false   | false       | https://badges.example.com/examples/one |
+      | id                                 | name                                 | type  | description | ABS(TIMESTAMPDIFF(SECOND, NOW(), created_at)) < 3 | is_open | send_emails | text_id                                 |
+      | 2                                  | AllUsers                             | Base  | null        | false                                             | false   | false       | null                                    |
+      | 4                                  | TempUsers                            | Base  | null        | false                                             | false   | false       | null                                    |
+      | @GroupExampleBadges                | Example badges                       | Other | null        | true                                              | false   | false       | https://badges.example.com/             |
+      | @UserMohammed                      | mohammed                             | User  | mohammed    | true                                              | false   | false       | null                                    |
+      | @GroupExampleBadgesMultipleParents | Example badges with multiple parents | Other | null        | true                                              | false   | false       | https://badges.example.com/parents      |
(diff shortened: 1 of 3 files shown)


smadbe (approved):

GeoffreyHuck:
Here's the options I could think of:

### Option 1: Rewrite the tests: rewrite the IDs with the new ones

PRO:
* Easy & fast

Con:
* It doesn't fix the issue. Next time we have a similar problem, we have to do it again.

### Option 2: Rewrite the tests: rewrite the way we query the tables

Instead of querying the tables, we specifically write the SQL query.
Ex: `the query "SELECT group_id, name FROM groups [possible JOINS] ORDER BY name ASC" should return:`

PRO:
* We can keep the ids in the first column

CON:
* The tests would be more difficult to understand
* The tests wouldn't necessary be equivalent to the old ones, we have to be careful we test the same things

### Option 3: Implement a super algorithm that solves the problem

I think there is one. What we need to find is whether a permutation of IDs correspond to what we want to check. This could be implemented using backtracking. For example:
- For each table, we match the rows of the test with the rows in the database, without ordering
- When we find an error, we go back at the moment we had the choice between selecting one row or another to set the reference
- If we arrive at the end of the tables check without an error: PASS; if it's not possible: ERROR

PRO:
* Works all the time

CON:
* Difficult to write/debug
* The complexity might be too high, although there might be ways to optimize it
* Cucumber is supposed to run the features sequentially. This method would require to "go back". It's a little weird.

### Option 4: Require the tests to be written in a way so there's no possible ambiguity

Here's an example:

image

I added `=@` when we set a reference. So we are explicit about it and can easily throw error when a non-defined reference is used.

In this case, the `groups` table appear with the `name` column first, because it allows us to unambiguously set the reference in the `id` column.

Then, for the table `groups_groups`, since all references are set (otherwise an error could be thrown because there's no `=@`, we can implement an `all rows must be in` algorithm: checking that all rows are present, without a specific order.

PRO:
* A specific error is thrown if there's a mistake, rather than just making the test fail

CON:
* The id column would not appear first in the tests anymore
* There might be cases were this method is not enough. If we want to check IDs for objects for which we have no way to set the reference by ordering with other columns.


**In addition: Why we cannot simplify the IDs using sequence numbers:**
Instead of using references, we could use `<rand(1)>`, `<rand(2)>`, `<rand(3)>` to specify that the number is the first, second, or third. But this would fail if we use the random generator for something else between generating the IDs.

It's also possible to do some mix.

smadbe:
Actually I prefer option 1 to 2, 3, 4. Changes are actually not straightforward to implement and/or make tests harder to understand/write. It is not that often that we do changes that require updating all ids.

> **In addition: Why we cannot simplify the IDs using sequence numbers:** Instead of using references, we could use `<rand(1)>`, `<rand(2)>`, `<rand(3)>` to specify that the number is the first, second, or third. But this would fail if we use the random generator for something else between generating the IDs.

That's another option... actually what I had in mind was even "simpler": mock the random seeder so that it outputs 1, 2, 3, 4... (I find `<rand(1)>`, `<rand(2)>` again yet another thing to understand). You could even set it with a guerkin rule "And I reset the random id generator to use sequence numbers" which would reset it to 1 to only use this sequence number rand generator only when it helps you. 
But of course, that would still require "rework" on all tests when you change something, but it should be easier (something like "I add '2' to all ids after that part"

GeoffreyHuck:
In the specific situation mentioned, I could get rid of the issue by generating the required random number later in the process, so the ids already specified stay the same.

I like the idea with sequence numbers. Maybe we could also specify the start number to not collide with ids that are already hard-coded, which is the case in many tests : `And I reset the random id generator to use sequence numbers starting at XXX`.

Since we don't need it now, we can close this PR. I just keep it open until I write a recap in the documentation, since the issue is likely to come back in the future, then I'll close it.

smadbe:
ok 👍

PR #1031, opened 2024-02-26
Update sessions database schema; adapt services and tests

GeoffreyHuck:
Related to https://github.com/France-ioi/AlgoreaBackend/issues/1022

This PR updates the database schema related to sessions and tokens, and update the services and tests to reflect this change. The new behaviors are not implemented yet (unless they are a direct side-effect of the new schema).

This is new schema:
```
sessions table:
session_id (PK)
user_id (FK)
refresh_token (optional)

access_tokens table:
token 
session_id (FK)
expires_at
issued_at
```

All services have been adapted to use the new schema, and the tests have been updated to reflect the changes.

The big idea is:
* We now have a `sessionID` in the request context that is set in the authentication middleware, along with the `userID`. So we can access the `sessionID` from the services.

Since it's a big PR, I think it's best to review it commit by commit. The commit messages include details.


db/migrations/2402201446_new_sessions_schema.sql
```diff
@@ -0,0 +1,135 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT='Sessions represent a logged in user, on a specific device.';
+
+CREATE TABLE `access_tokens` (
+  `token` VARBINARY(2000) NOT NULL COMMENT 'The access token.',
+  `session_id` BIGINT NOT NULL,
+  CONSTRAINT `fk_access_tokens_sessions_session_id` FOREIGN KEY (`session_id`) REFERENCES `new_sessions`(`session_id`) ON DELETE CASCADE,
+  `expires_at` DATETIME NOT NULL COMMENT 'The time the token expires and becomes invalid. It should be deleted after this time.',
+  `issued_at` DATETIME NOT NULL DEFAULT NOW() COMMENT 'The time the token was issued.'
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT='Access tokens (short lifetime) distributed to users, to access a specific session.';
+
```
smadbe, line 68: I see you lost most of the indexes that were existing before: (at least in this commit)
```
 INDEX `expires_at` (`expires_at`) USING BTREE,
  INDEX `access_token_prefix` (`access_token`(767)) USING BTREE,
  INDEX `user_id` (`user_id`) USING BTREE,
  INDEX `refresh_token_prefix` (`refresh_token`(767)) USING BTREE,
```
I assume most of them were making sense. (including this way of indexing the prefix)

smadbe (approved):

smadbe:
The first 5 commits are related with another issue!

db/migrations/2402201446_new_sessions_schema.sql
GeoffreyHuck: fixed. Added the same indexes.

From what I could see, only the index on `access_token` is really useful. The other ones I'm not sure, but it shouldn't hurt too much as those tables have few indexes anyway.

GeoffreyHuck:
I fixed the issue with the rebase. Fortunately, it didn't mess up with the comments.

PR #1032, opened 2024-02-28
Parallel session logout

GeoffreyHuck:
Related to #1022 

Add a test to make sure we only delete the current session if the user has more than one on logout.

It was the case, but it wasn't tested.

Easier to review commit by commit.


app/api/auth/logout.feature
```diff
@@ -37,6 +37,35 @@ Feature: Sign the current user out
...
+    When I send a POST request to "/auth/logout"
+    Then the response code should be 200
+    And the response body should be, in JSON:
+    """
+    {
+      "success": true,
+      "message": "success"
+    }
+    """
+    And the response header "Set-Cookie" should be "[NULL]"
```
smadbe, line 57: Just a remark, and not related with this PR but I find `should be "[NULL]"`  not clear... I suppose that's not the content of the cookie. If the cookie should not be set, it could be `should not be set`.  But here I assume that's something even different as the way to delete a cookie is to expire it I think (I may be wrong). I prefer where gherkin rules are explicit, I don't want to have to check the guerkin rule to understand it.

smadbe (approved):

app/api/auth/logout.feature
GeoffreyHuck: fixed. I made a new feature to handle this case `And the response header "Set-Cookie" should not be set`.

But we cannot use it all the time. Some tests are multi-tests that checks either for "not defined", or for some specific values:

```
And the response header "Set-Cookie" should be "<expected_cookie>"
...
Examples: 
  | expected_cookie
  | [Header not defined]
  | [Header not defined]
  | [Header not defined]
  | access_token=2!ny93zqri9a2adn4v1ut6iz...
  | access_token=; Path=/; Domain=127.0.0.1...
```

I renamed the old `[NULL]` to `[Header not defined]`, to describe what it is. I don't think we can do better than this.

PR #1033, opened 2024-03-05
Makefile: add command serve-swagger

GeoffreyHuck:
`make serve-swagger`

Generates the swagger file, and starts the "serve" command, so it can be checked in the browser.

This improves the process to check if the doc has been generated correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,8 +96,8 @@ lint: $(GOLANGCILINT)
 validate-swagger:
 	swagger generate spec --scan-models -o ./swagger.yaml && swagger validate ./swagger.yaml
 
-serve-swagger:
-	swagger generate spec --scan-models -o ./swagger.yaml && swagger validate ./swagger.yaml && swagger serve ./swagger.yaml --no-open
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -96,6 +96,9 @@ lint: $(GOLANGCILINT)
 validate-swagger:
 	swagger generate spec --scan-models -o ./swagger.yaml && swagger validate ./swagger.yaml
 
+serve-swagger:
+	swagger generate spec --scan-models -o ./swagger.yaml && swagger validate ./swagger.yaml && swagger serve ./swagger.yaml --no-open
```
smadbe, line 100: Coudn't it be:
```suggestion
serve-swagger: validate-swagger
	swagger serve ./swagger.yaml --no-open
```

smadbe (approved):

Makefile
GeoffreyHuck: fixed. I didn't now we can do this, it's a lot better!

PR #1034, opened 2024-03-05
Parallel session refresh

GeoffreyHuck:
Related to #1022 

This PR includes two changes:

1. When we call the service to refresh with a token that is **not** the most recent one for the session, we return the most recent one, with its expiration time
2. When we call the service to refresh with a token that is **the most recent** one, and if it has been issued **less than 5 minutes ago**, we just return the current token without refreshing, with the expiration time
3. (actual refresh) When we call the service to refresh with a token that is **the most recent one**, and if it as been issued **more than 5 minutes ago**, we refresh the token

- Added a new *.feature file for the new behavior that uses the new testing system
- The new testing system now handles sessions and access tokens
- Updated the older tests. The times have been updated to make it easier to read, and to make the old tests compliant with the new behaviors.
- Updated the service swagger doc

Easier to read commit by commit, with the details in commit descriptions.

## The lock `userID` have been changed to be on `sessionID`

There was a lock on `userID` to prevent multiple refresh concurrently. It was modified to be on the `sessionID` instead, because we don't want to have a lock waiting on one device because we coincidentally have a refresh on another device for the same user.

I had a bug with it (two last commits), because a test was using the lock manually without clearing it, a another test was blocked because it uses the same `sessionID`. It's weird to use the lock manually instead of using the implementation function `withLock`, which would have cleaned it automatically. Anyway, the test was updated to clear the lock, to fix the bug.

### DIFF
--- a/app/api/auth/create_access_token.go
+++ b/app/api/auth/create_access_token.go
@@ -59,17 +59,23 @@ const parsedRequestData ctxKey = iota
 //			and if the authentication is valid.
 //			This happens when the frontend app loads and the user is already logged-in.
 //			Then,
-//			if the access token used isn't the most recent access token of the user, we return the most recent access token.
-//			If the access token used is the most recent access token of the user, and it has been refreshed LESS than 5 minutes ago,
-//			we just return it.
-//			If the access token used is the most recent access token of the user, and it has been refreshed MORE than 5 minutes ago,
-//			we refresh the access token and return the new access token
-//			(locally for temporary users or via the login module for normal users) and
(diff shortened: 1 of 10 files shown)


GeoffreyHuck:
2. If we try to refresh with a token that is the most recent one, and if it has been issued less than 5 minutes ago, we just return the current token without refreshing, with the expiration time
3. (actual refresh) If we try to refresh with a token that is the most recent one, and if it as been issued more than 5 minutes ago, we refresh the token

smadbe (approved):

PR #1036, opened 2024-03-07
createAccessTokenToken: Limit number of session at 10 / users to avoid session spamming

GeoffreyHuck:
Related to #1022 

Limits the number of sessions to 10 for a user.

When a new session is created (user logs in), we count the number of sessions. If it's more than 10, we remove the oldest one(s).

We first retrieve the list of sessions along with the most recent `issued_at` access token, sorted by `issued_at DESC`. Then we delete the oldest.

Adds to new Gherkin features:
- `there are X sessions for user @Session`, checks the number of sessions for a user
- `there is no session @Session`, checks that a session doesn't exists (has been deleted)

### Do we need an index?

The query is executed every time a user logs in.

Here's an explain of the request :
image

`sessions` is filtered by `user_id`, which is a foreign key.
`access_tokens` needs to sort the matching `session_id` by `issued_at`. We will never have many access tokens for one session in the database because:
1. We'll remove the expired tokens (coming soon)
2. We can only refresh the token every 5 min

So, no index required.

### An unnecessary global variable was removed in the test package

Some lint tests were failing, which shown that we had a global variable `db` for the database in the test package. A refactoring was done to remove the global variable. Details are in the last commit message.

> Easier to review commit by commit.

### DIFF
--- a/app/database/session_store.go
+++ b/app/database/session_store.go
@@ -1,5 +1,7 @@
 package database
 
+import "github.com/France-ioi/AlgoreaBackend/app/database/mysqldb"
+
 // SessionStore implements database operations on `sessions`.
 type SessionStore struct {
 	*DataStore
@@ -41,8 +43,8 @@ type sessionWithMostRecentIssuedAt struct {
 	IssuedAt  Time
 }
 
-// GetUserSessionsSortedByMostRecentIssuedAt returns the user's sessions sorted by most recent issued_at.
-func (s *SessionStore) GetUserSessionsSortedByMostRecentIssuedAt(userID int64) []sessionWithMostRecentIssuedAt {
+// GetUserOldSessionsToDelete returns the user's old sessions to delete because we keep only the `nbKeep` most recent ones.
+func (s *SessionStore) GetUserOldSessionsToDelete(userID int64, nbKeep int) []sessionWithMostRecentIssuedAt {
 	var sessions []sessionWithMostRecentIssuedAt
 
 	// Sessions without access tokens are treated as the oldest ones.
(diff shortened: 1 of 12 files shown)


smadbe:
> We first retrieve the list of sessions along with the most recent `issued_at` access token, sorted by `issued_at DESC`. Then we delete the oldest.

I suppose/hope you do sorted by `issued_at ASC` and take the first one instead :-D

I think for being future-safe (and conceptually correct), you must consider that a session might have no token at all, in which case either `issued_at = 1000-01-01` or the session is always destroyed.

app/database/session_store.go
```diff
@@ -32,3 +32,47 @@ func (s *SessionStore) GetUserAndSessionIDByValidAccessToken(token string) (user
...
+		oldestSessions := sessions[max:]
+
+		oldestSessionIDs := make([]int64, len(oldestSessions))
+		for i, session := range oldestSessions {
+			oldestSessionIDs[i] = session.SessionID
+		}
+
+		err := s.Delete("session_id IN (?)", oldestSessionIDs).
+			Error()
+		mustNotBeError(err)
```
smadbe, line 76: - Is there a reason to do a select (with `max` unused rows) and N delete queries ? If no, do the whole thing in one single SQL DELETE query!
- as discussed, make sure you delete sessions with no access token
- use a tie-breaker so that the rows you delete is predictable

smadbe (changes requested):

app/database/session_store.go
GeoffreyHuck: - The sessions without access tokens are now deleted in priority. Used `UNION ALL`, not sure if it's possible in a different way: on one end we have those with no join (left join), and on the other end we have is an aggregation.
- Added `session_id` as a tie-breaker

For the first item I'm not sure I understand. There's only one `DELETE` query with the `IN` operator. Do you mean merging the `SELECT` and the `DELETE` together? This might end with a query that's very difficult to understand.

smadbe: Discussion continued in Slack.

smadbe (changes requested):
I am still waiting for the latest comments to be answered / addressed for restarting a review.

app/database/session_store.go
GeoffreyHuck: fixed, the queries have been merged. There are some implementations details I've put in comments and in the commit messages (the transformation was done step-by-step in the last 3 commits).

smadbe (approved):
👍

PR #1043, opened 2024-03-27
Update group: handle approval requirement changes

### LINKED ISSUE #1039: Update group: handle approval requirement changes
## Motivations

The current update group service lets the user change the approval as he wants, but in practice, you should not be able to change the rules while some member has approved them already.

## Subtasks
- [ ] `require_lock_membership_approval_until` should be nullable (it should be able to re-set this value to null when not null)
- [ ] if there is at least one participant (user ou team) in the group and one of the approval rule is strengthened:
  - [ ] if `approval_change_action` is not given, fail (403?)
  - [ ] if `approval_change_action` is "empty", remove all the participant from group 
    - [ ] update `membership_changes` table as well !
  - [ ] if  `approval_change_action` is "reinvite", remove all the participant from group and invite them to join the group
- [ ] if no approval rule is strengthened, `approval_change_action` **must** not be given
- [ ] `require_lock_membership_approval_until` is strengthened if the new value is >NOW and either:
  - [ ] the previous value was null, or
  - [ ] the previous value < the new value
- [ ] `require_watch_approval` is strengthened if was `false` and becomes `true`
- [ ] `require_personal_info_access_approval`  is strengthened if was `none` and become s`read` or `edit`, or if was `read` and becomes `edit`

GeoffreyHuck:
fixes #1039 

Implement rules when strengthening a `require_* field as detailed in #1039 

- Update swagger doc with the description of the new rules
- A few refactors had to be done because the lint didn't pass (cyclomatic complexity too high)
- Tests were put in a new feature files because the functionality is quite big, and so we can use the new test features
- New features add to be added to the test system to handle pending requests, group membership changes
- Updated database schema to add the `action` enum `removed_due_to_approval_change` in table `group_membership_changes`
- The test system has been slightly refactored because it couldn't handle the way we define default values correctly (see below)

## Refactor of the test system

There are two limitations with the test system.

### 1. Everything is in one file, and it keeps growing

For now, the idea is to split it by modules:
- everything related to the Gherkin features related to `groups` will be defined in `app_language_groups.go`.
- everything related to the Gherkin features related to `group_pending_requests` will be defined in `app_language_group_pending_requests.go`.
- ...
- Additionally, we can have files with common functions, for example, `app_language_db.go`.

It's a first step, it might be refactored into objects in the future because there are common functionalities between modules.

### 2. The way we handle default values

To provide a higher degree of abstraction, in many definitions, we define rows in tables that are a requirement of the definition (for example: when we had a member to a group, we also add a row in the user table).

The problem was that the system would put in database the values defined the last time if was defined, directly or indirectly. For example: if we define a group with name `my name`, but a later feature defines the same group with the name `default name`, the database would contain `default name`.

The solution retained is to explicitly declare in the testing system when we set a field:

```
  ctx.setGroupFieldInDatabase(
    ctx.getGroupPrimaryKey(groupID),
    "require_personal_info_access_approval",
    group["require_personal_info_access_approval"],
  )
```

In this way, values are not redefined at default when there are set.

This also makes it clearer what is defined in the current feature.

### Test refactor

Documentation will be added in the devdoc to precise the plans. For now, it has only been made for groups, group pending requests and group membership changes, because those are the modules that were impacted with this PR.

### Notes

The changes are large and since the commits are incremental and many refactoring were needed, it might be easier to review everything at once rather than commit by commit. But the commit messages contain all the details if needed.


GeoffreyHuck:
There apparently a test coverage issue. I'll fix that tomorrow. But it's already reviewable.

app/api/groups/update_group.go
```diff
@@ -46,12 +53,47 @@ type groupUpdateInput struct {
...
+	RequireLockMembershipApprovalUntil *database.Time `json:"require_lock_membership_approval_until" validate:"changing_requires_can_manage_at_least=memberships_and_group,strengthening_requires_approval_change_action"` //nolint:lll
+	// Strengthened if the new value is `true` and the old value is `false`.
+	//
+	// Not considered strengthened if the group doesn't have any participants.
+	//
+	// If it is strengthened, `approval_change_action` must be set.
+	RequireWatchApproval       bool `json:"require_watch_approval" validate:"changing_requires_can_manage_at_least=memberships_and_group,strengthening_requires_approval_change_action"` //nolint:lll
+	RequireMembersToJoinParent bool `json:"require_members_to_join_parent" validate:"changing_requires_can_manage_at_least=memberships_and_group"`                                       //nolint:lll
+
+	// Nullable
```
smadbe, line 81: nullable? What is the semantic of giving `null` ?

smadbe (approved):
👍  (I haven't checked every details)

PR #1054, opened 2024-04-03
Add a stack trace in the logs when the binary crashes. 

GeoffreyHuck:
Because we know it crashes but we don't know where.


smadbe (approved):
👍

PR #1060, opened 2024-04-22
Fix slow query: switch the two where conditions.

GeoffreyHuck:
First the condition on `parent_item_id`, and then the `NOT IN` makes the query a lot faster!


smadbe (approved):
🤷‍♂️

PR #1066, opened 2024-05-07
Fix slow trigger before_delete_items_items

GeoffreyHuck:
The trigger `before_delete_items_items` can be very slow. It insert an entry in table `results_propagate` for each entry in `results` of a given `item_id`. This means the insertion of 50k rows when we remove a child of `694914435881177216`.

This PR extract the inserts in table `results_propagate` out of the `before_delete_items_items`, and put it at the beginning of the process of `results propagation`.

Instead of doing the insert in the trigger, we add the `item_id` in question in a new table (`results_propagate_items`). Then, when we do the `results propagation`, we first retrieve the `item_id` in the new table, and do the inserts at this moment.

We do one transaction for each `item_id`, because we know it can take a long time, and we want to keep transactions as small as possible.

### DIFF
--- a/app/database/result_store_propagate.go
+++ b/app/database/result_store_propagate.go
@@ -36,6 +36,8 @@ func (s *ResultStore) propagate() (err error) {
 	defer recoverPanics(&err)
 	// Use a lock so that we don't execute the listener multiple times in parallel
 	mustNotBeError(s.WithNamedLock(propagateLockName, propagateLockTimeout, func(s *DataStore) error {
+		s.setResultsPropagationFromResultsPropagateItems()
+
 		mustNotBeError(s.InTransaction(func(s *DataStore) error {
 			initTransactionTime := time.Now()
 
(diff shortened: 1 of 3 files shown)


smadbe:
> We do one transaction for each item_id, because we know it can take a long time, and we want to keep transactions as small as possible.

The "problem" is that it can be very small for many... and not small enough for others.

Does the insert locks anything (even just `results_propagate`) ?

If no, there is no reason to do one transaction per item, I suppose it should even be possible to run it in a single query (which would be faster than n times a transaction+a query)
If yes, and always takes <0.5s, that's ok.
If yes and it may take >0.5s, it should also split in such a way that each transaction, only adds ..let's say 5k entries... so that other sync real-time (web) queries are not slowed down.

smadbe:
And so could, you, instead of what you currently do in `setResultsPropagationFromResultsPropagateItems`, run the old 
```INSERT INTO `results_propagate`
  SELECT `participant_id`, `attempt_id`, `item_id`, 'to_be_propagated' AS `state`
  FROM `results`
  WHERE `item_id` = OLD.`parent_item_id`
  ON DUPLICATE KEY UPDATE `state` = 'to_be_recomputed';``` request but adding a `LIMIT 5000` on the select and track the number of row changed which is returned by the sql engine?

**Comment by GeoffreyHuck** [comment]
Added a `LIMIT...OFFSET...`. The `rowsAffected` returned is sometimes > offset, I don't know why. To be safe, we execute the query until we get 0. This means the query will be execute one time for nothing, if that's a problem I can investigate further what happens with the number of rows affected.

**Comment by smadbe** [comment]
as discussed:
- maybe test with a ` AND ... NOT IN (SELECT .. FROM results_propagate WHERE ...)` instead of the `OFFSET` 
- add a test with only `INSERT`s and a test with only `UPDATE`s to check that the `RowsAffected` work as expected.

File: app/database/result_store_propagate.go [hunk 1]
```diff
@@ -359,3 +361,68 @@ func (s *ResultStore) propagate() (err error) {
...
 	return nil
 }
+
+// setResultsPropagationFromResultsPropagateItems inserts results_propagate rows from results_propagate_items.
+func (s *DataStore) setResultsPropagationFromResultsPropagateItems() {
+	const maxInserts = 5000
+
+	const fromResultsWithItemIDToPropagateQueryPart = `
+		FROM results
+		WHERE item_id IN (SELECT DISTINCT item_id FROM results_propagate_items ORDER BY item_id)
```
smadbe, line 371: why `DISTINCT` ? `item_id` is already a PK on results_propagate_items

smadbe (changes requested):

app/database/result_store_propagate.go
GeoffreyHuck: fixed, indeed

GeoffreyHuck:
Also comment a test the was failing 2 every 3 runs on CI. With 11min for failing test, it makes me lose 20min every time I push.

I can look into this when the current emergency is done. I checked a little bit and didn't find anything wrong with it. My guess: it's just too slow on CI and it times out (it tests a lock that has 10s timeout, executing the results propagation 30 times).

smadbe (approved):

PR #1070, opened 2024-05-10
Adapt invitation service to current needs

### LINKED ISSUE #1056: Adapt invitation service to current needs
## Motivations

The group invitation service is one of the oldest service and many things have changed around it. In addition, to get "approvals" that the user has to approval, he needs to get them (and he is not allowed to get the group info to get them).

## Subtasks

Update the `GET /current-user/group-invitations` service so that:
- [x] only return actual invitation the user has received (that he has not approved yet, which has not been removed, ...), no more join request things
- [x] so no need for the "action" in the output anymore
- [x] the `group` in output should include the 3 approval field values
- [x] `inviting_user` should not be `nullable` anymore I assume
- [x] I think `within_weeks` is not useful anymore... so if it is easier to remove it, remove it (otherwise, you can leave it as it is)

GeoffreyHuck:
fixes #1056 

Modifications of service `invitationsView`:

- only return actual invitation the user has received (that he has not approved yet, which has not been removed, ...), no more join request things
- remove the `action`" in the output, not needed anymore
- include the 3  `require approval` fields for the group
- `inviting_user` in response is not `nullable` anymore
- remove `within_weeks` in response, not useful anymore

### Review

Easier to review commit by commit. The details are in commit messages.


smadbe (approved):
looks good

smadbe:
Let me merge it. As it breaks the current API, I'll deploy when I am sure we can fix the frontend immediately.

PR #1074, opened 2024-05-14
Speed up slow query `start-result-path`.

GeoffreyHuck:
We check the permissions before instead of inside the big query.

The use of `WITH...` to check the permissions made the whole query slow. This is weird because the query to check the permissions itself is very fast, but it looks it messes with the optimizer in some way.

Another problem appeared as the query got very big (8 items in the path), it would spend a lot of time in `state=statistics`. Removing the `WITH...` altogether before fixes it (probably the algorithm to determine the order to do the JOINs gets very slow when there are many JOINs + WITH).


smadbe (approved):
Thanks, ok we'll see. 👍

PR #1076, opened 2024-05-15
Fix lock timeout when the `results propagation` leads to an unlocked groups

GeoffreyHuck:
The named lock was asked again without releasing it first, leading to a lock timeout.

It's now out of the scope of the lock, so it'll be released before asking it again.


smadbe (approved):
looks good

PR #1077, opened 2024-05-15
`startResultPath` and `startResult`: the `results propagation` is now async

GeoffreyHuck:
Needed to add a boolean in `MarkAsToBePropagated` to make it compatible with the other places it's used, but it's temporary: the propagation will be async everywhere in a further commit.

### DIFF
--- a/app/service/propagation.go
+++ b/app/service/propagation.go
@@ -44,13 +44,19 @@ func SchedulePropagation(store *database.DataStore, endpoint string, types []str
 
 	if endpoint == "" || endpointFailed {
 		// Sync.
-		err := store.InTransaction(func(store *database.DataStore) error {
+		if store.IsInTransaction() {
 			store.ScheduleItemsAncestorsPropagation()
 			store.SchedulePermissionsPropagation()
 			store.ScheduleResultsPropagation()
+		} else {
+			err := store.InTransaction(func(store *database.DataStore) error {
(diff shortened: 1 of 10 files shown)


app/service/propagation.go
```diff
@@ -44,13 +44,19 @@ func SchedulePropagation(store *database.DataStore, endpoint string, types []str
 
 	if endpoint == "" || endpointFailed {
 		// Sync.
-		err := store.InTransaction(func(store *database.DataStore) error {
+		if store.IsInTransaction() {
```
smadbe, line 47: I don't get the "// Sync." comment, this case corresponds to sync propagation?

Does it make sense to always run these in a global transaction?

GeoffreyHuck: Yes, sync propagation in this case. We schedule them inside a transaction, but the propagation are called when the transaction ends, because that's how it was implemented for now.

I think a better rule for scheduling would be:
- if we are in a transaction, we call it after the current transaction
- if we are not, we call it now

This would remove the need for the weird check.

Also there is something else I just figured out: it's a bad idea to call the endpoint while being inside a transaction, it just makes it longer. So the same kind of rule should be implemented for the call to the endpoint: now outside a transaction, after the transaction if we are in it.

I'm implementing those in a new commit.

smadbe (approved):

PR #1082, opened 2024-05-27
Allow not providing the access token when the task token is provided

### LINKED ISSUE #1057: Allow not providing the access token when the task token is provided
## Motivations

As shown in the [token interaction with tasks](https://france-ioi.github.io/algorea-devdoc/task_platforms/token/), task interactions starts with getting a task token, and then may ask for an answer token or a hint request token, both given a task token.

For these 2 requests, both the usual access token and the task token are provided. However, other agents than the frontend may need to use these services. A CLI is planned as well. It is better if these external tools do not need the access token.

## Subtasks
For the `POST /answers`  (generate an answer token) and `POST /items/ask-hint` (Register a hint request)  services:
- [x] do not require nor use the access token
- [x] make sure you fully validate the task token, i.e. the signature and that the date is either today or yesterday)
- [x] Update service spec
- [x] Make sure you test both the cases where UserID = and != ParticipantId

GeoffreyHuck:
fixes #1057 

Update `itemGetAnswerToken` and `itemGetHintToken` services to be called without an access token.

Those services were removed from the auth middleware.

## Tests to check we don't accept falsified tokens

To simulate the falsification of a token, a single byte was changed in its payload. This way, the signature of the token isn't valid anymore. Details are in the specific commit.

## Tests for teams

Two tests were added for a team but I'm not really sure how teams are working. When a hint is asked, it's asked for the team? If so, the test is probably correct.

## Do we need more robustness tests for the teams?

There are no tests that specifically check that:
- The user is really in the participant team when a team is used
- The team has the right to submit / ask a hint

I'm not sure whether this should be added. Should we consider that since the token has been generated by the backend, we can trust everything in it? And what if some permissions changed since?

## Review

Easier to review commit by commit with the details in commit messages.

### DIFF
--- a/app/api/items/ask_hint.feature
+++ b/app/api/items/ask_hint.feature
@@ -26,8 +26,7 @@ Feature: Ask for a hint
     And time is frozen
 
   Scenario: User is able to ask for a hint
-    Given I am the user with id "101"
-    And the database has the following table 'attempts':
+    Given the database has the following table 'attempts':
       | id | participant_id |
       | 0  | 101            |
     And the database has the following table 'results':
@@ -90,8 +89,7 @@ Feature: Ask for a hint
     And the table "results_propagate" should be empty
 
   Scenario: User is able to ask for a hint with a minimal hint token
-    Given I am the user with id "101"
-    And the database has the following table 'attempts':
+    Given the database has the following table 'attempts':
       | id | participant_id |
       | 0  | 101            |
     And the database has the following table 'results':
@@ -154,8 +152,7 @@ Feature: Ask for a hint
     And the table "results_propagate" should be empty
 
   Scenario: User is able to ask for an already given hint
-    Given I am the user with id "101"
-    And the database has the following table 'attempts':
+    Given the database has the following table 'attempts':
       | id | participant_id |
       | 0  | 101            |
     And the database has the following table 'results':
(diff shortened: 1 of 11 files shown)


app/api/items/ask_hint.feature
```diff
@@ -89,9 +93,71 @@ Feature: Ask for a hint
...
 
+  Scenario: User is able to ask for a hint for a team (participant_id is the first integer in idAttempt in the task token)
+    Given the database has the following table 'attempts':
+      | id | participant_id |
+      | 0  | 201            |
+    And the database has the following table 'results':
+      | attempt_id | participant_id | item_id | hints_requested        | hints_cached | started_at          |
+      | 0          | 201            | 50      | [0,  1, "hint" , null] | 4            | 2020-01-01 00:00:00 |
+      | 0          | 201            | 10      | null                   | 0            | 2020-01-01 00:00:00 |
+  And "priorUserTaskToken" is a token signed by the app with the following payload:
```
smadbe, line 104: isn't there a misalignment here?

smadbe (approved):
👍 

For the remarks about teams, that should be good enough for now.

app/api/items/ask_hint.feature
GeoffreyHuck: fixed

PR #1084, opened 2024-05-29
Save grade: use score/answer as auth

### LINKED ISSUE #1079: Save grade: use score/answer as auth
## Motivations

As shown in the [token interaction with tasks](https://france-ioi.github.io/algorea-devdoc/task_platforms/token/), task interactions starts with getting a task token, and then may ask for an answer token or a hint request token, and then the task sends its score to the save grade service, either via a score token (signed by the task grader) or an answer token+score if the task has no server (so basically the score is given by the user himself).

As the task grader may send the score much later, he cannot use the API access token to call the service. Also it does necessarily have the task token. 

## Subtasks
For the  POST /items/save-grade service:
- [x] do not require nor use the access token
- [x] remove the task token from input
- [x] so the input is either the answer token+score, or the score token... these have all the required information to run the service
- [x] make sure you fully validate the tokens, i.e. the signature (from either the task grader for score token, or from the backend if (and only if) the task platform has no keys) and that the date is either today or yesterday)
- [x] do not return the "task token" anymore as we do not use it in practice in the frontend (and the task grader should not get it)
- [x] Update service spec
- [x] Make sure you test both the cases where UserID = and != ParticipantId

GeoffreyHuck:
fixes #1079 

Remove authentication with `access token` for the service `saveGrade`. Use either the `answer token` or the `score token`.

Also added tests to check:
- The tokens cannot be falsified
- The tokens cannot be expired
- The service works with `participant_id` != `user_id`

## Representation of inputs in the service

The method is the same as the one that was already in use. If the input is `answer token`+`score`, we construct a `score token` from it. Then, in the rest of the service, we only use the `score token` to access the inputs. This avoid avoid `if/else` depending of whether the input consists of `answer token`+`score` or `score token`.

## Unmarshalling of `score token` need to use the `LocalItemID`

The current system checks the signature at the same time of unmarshalling tokens. Problem: we need to use the `LocalItemID` inside the `score token` to know which platform key to use to unmarshal the `score token` itself. So a method `token.GetUnsafeFromToken` has been added, to be able to retrieve a parameter inside the token without checking the signature. `Unsafe` was added in the name to make sure this method isn't use in wrong situations.

## Parsing and validation of tokens

The same method that was use to parse and validate the fields `idUser`, `idItemLocal` and `idAttempt` of `task tokens` and put them in a `Converted` sub structure during unmarshalling has been reused for `answer tokens`. This way is more consistent, and we are sure we are dealing with valid information when we access the content of the token.

## Little uncertainty about score tokens

Some tests are using a `score token` with an undefined `AttemptID` and `LocalItemID`. So for now the parsing and validation of `score token` works if there is no `AttemptID` or `LocalItemID`. But maybe it should fail? Can it happen in reality?

## Review

Easier to review all modifications at once as some refactoring were made. But the details of the changes can be seen in commit messages.


smadbe:
> Some tests are using a score token with an undefined AttemptID and LocalItemID. So for now the parsing and validation of score token works if there is no AttemptID or LocalItemID. But maybe it should fail? Can it happen in reality?

But I don't get how you set the score to the `result` entry if the `item_id` or `attempt_id` is missing, so I cannot see how it may "work". If it is only on the token parsing, that's an internal decision you have to make, but I don't see cases where no  having `item_id` or `attempt_id` makes sense.

smadbe:
the PR closing was a mistake

smadbe (approved):
👍

GeoffreyHuck:
The unmarshalling now fails if there is no `idAttempt` or `idLocalItem`, because `score tokens` are used only in this service, and they're required for the service to work.

zenovich:
A note for the sake of history: https://github.com/France-ioi/AlgoreaBackend/commit/995754de5b0eec1140dd35145ac64efe9cb7cea3#r157170748

PR #1085, opened 2024-06-04
Make sure the tests cannot be run on live database as it empties the database

GeoffreyHuck:
An error is thrown if the test environment uses a database on AWS.

So the issue that happened last time wont happen anymore.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -136,6 +136,11 @@ func DBConfig(globalConfig *viper.Viper) (config *mysql.Config, err error) {
 	vConfig.SetEnvPrefix(fmt.Sprintf("%s_%s_", envPrefix, databaseConfigKey))
 	vConfig.AutomaticEnv()
 	err = vConfig.Unmarshal(&config)
+
+	if appenv.IsEnvTest() && strings.HasSuffix(config.Addr, "amazonaws.com") {
+		err = fmt.Errorf("cannot connect to AWS RDS in tests as it empties the database")
+	}
+
 	return
 }
 


smadbe (changes requested):
> Note: in the test TestDBConfig_FailsOnAWSForTestEnv, appenv.IsEnvTest has to be patch to return true for the test to succeed on CI. Which means the env is not called "test" on CI. Not sure if that's expected.

If the test are not run in "test" env, isn't all that PR useless? That does not look like a detail.

I would have prefer the code not to bound a specific vendor (`amazonaws.com`) but I assume white listing (only allow `127.0.0.1|localhost`) would not work sometimes.

But more basically, why did you have that DB in your config? To call a service? But so, with that change, will you call the service then? If you change the env to call the service, don't we fallback to the same problem if you run `restore` after changing that env?

GeoffreyHuck:
I investigated a little bit further:

1. The `env` is not set to `test` when `make test` is used to execute the tests, like it is done on CI. I think it defaults to `dev`. This is most likely not intended.
2. The `env` set to `test` when we execute the tests in the IDE (e.g. Jetbrains).

That's why it works on my machine but not on the CI.

## The process that caused the issue

1. Set the config DB to the live DB, to test a service
2. Compile & Run the server
3. Call the service I want by starting a `curl` to the server, which uses the live DB
4. ... (working on something else and forgot to remove the live DB from the config)
5. Run a test (in the IDE, so the `test env` is properly set
6. ... (the first step of running a test with database is to clean the database)

This is the issue that this PR solves.

It would probably be nice to make sure the `env` is set to `test` when `make test` is called too.

Does this make sense to you?

smadbe:
Wouldn't it make sense to force the env **to be** "test" (not force the to "test"!) when running test, and force the env not to be "test" when "serving". This would force to have 2 different config files, and force the choice (you could still keep one for default depending on what is your default operation).

GeoffreyHuck:
Just found something interesting.

For the record, I've put all the config in `config.yml` and not `config.test.yaml`. I remembered it didn't work. I just found what happened: there was a syntax error in `config.test.yaml`, and so it wasn't used at all.

In the logs I had `Cannot merge "test" config file, ignoring it: While parsing config: yaml: line 10: found unexpected end of stream`.

So ended up setting everything is `config.yml`.

But even then, I don't think it's just a problem of `env`. I could have just though "let's start the server in test mode because I'm testing something" (`AlgoreaBackend serve test`), and woudl have updated the `config.test.yaml` with the live database, and then I would have faced the same problem. So even if we fix the `env`, we might run into this problem again in the future.

Maybe the kill switch should be in the test module itself when the tables are purged? Right before they are purged, we check if we are on aws? Or do you see another way of doing?

smadbe:
> For the record, I've put all the config in config.yml and not config.test.yaml. I remembered it didn't work. I just found what happened: there was a syntax error in config.test.yaml, and so it wasn't used at all.
> 
> In the logs I had Cannot merge "test" config file, ignoring it: While parsing config: yaml: line 10: found unexpected end of stream.
>
> So ended up setting everything is config.yml.
Have you fixed the problem now?

> But even then, I don't think it's just a problem of env. I could have just though "let's start the server in test mode because I'm testing something" (AlgoreaBackend serve test), and would have updated the config.test.yaml with the live database, and then I would have faced the same problem. So even if we fix the env, we might run into this problem again in the future.

That's why my suggestion was to forbid running `serve` in `test` env... that would not prevent you from copying the `test` config in the prod one. But at least the prod one could not be used to run test by mistake.

> Maybe the kill switch should be in the test module itself when the tables are purged? Right before they are purged, we check if we are on aws? Or do you see another way of doing?

My initial point was that checking `aws` in the url is very vendor specific, if a partner wants to run it on Google cloud, the issue would come back.

GeoffreyHuck:
Yes now it is fixed.

Okay I missed your point yesterday. That would work indeed. Forcing "test" for tests, and forbidding "test" for `serve`. Let's implement this!

GeoffreyHuck:
It's done.

But I'm still not sure it's enough.

Nothing actually forbids to just use one `config.yaml`, and never set up a `config.test.yaml`.

smadbe:
> t's done.
> 
> But I'm still not sure it's enough.
> 
> Nothing actually forbids to just use one `config.yaml`, and never set up a `config.test.yaml`.

An option would be not to use the `config.yaml` file for tests.

GeoffreyHuck:
Yeah, I'll also add a big warning comment in the `config.test.sample.yml`. I think it's acceptable.

So:
- [ ] Don't use `config.yaml` for the `test` env, use only `config.test.yaml`. Return an error if it doesn't exist.
- [ ] Add a warning comment to never use a live database in `config.test.sample.yaml`
- [ ] Add a line in `README.md` to say to configure the `test` environment for tests
- [ ] Update the CI I think
- [ ] Remove the kill switch with the reference to `aws`

GeoffreyHuck:
Done. The code is simple but the tests were a bit tricky to implement. I added comments to make it a bit clearer.

smadbe (approved):
👍

PR #1086, opened 2024-06-06
Service for getting a token for editing another user's profile

### LINKED ISSUE #1044: Service for getting a token for editing another user's profile
## Overview

When joining some groups, user may grant to group manager to "edit" their profile. The intend is to let the group manager (typically the teacher, instructor, ...) fix some personal information and change the password of a user who barely knows about his account on the website.

As the authentication is handled by the login-module, we need to provide the login-module with a proof that the current user is allowed to modify another user.

## Acceptance criteria

- [x] create a new service for requesting such a token for a given **target** user id
- [x] the current user should only be allowed to request such a token if he is a manager (implicit or explicit) of a group of which the target user is a member and which (the group) has `require_personal_info_access_approval === 'edit'`
- [x] the token should be encoded in aes-128-ecb with the usual client secret and contain, as json:
  -  `requester_id`: the `loginId` of the current user
  - `target_id` : the `loginId` of the target user
  - `expires` : a unix timestamp of NOW + 30min
- [x] the service should also return: 
  - alg: the algorithm which has been used for creating the token.  by decrease order of preference: AES-128-ECB,  AES-256-CBC, AES-256-GCM - implement the one which requires reasonable time of implementation (otherwise, it can be move to another issue)
  - tag (optional) the public tag generated by the encoding (for instance in GCM)

GeoffreyHuck:
fixes #1044

## Encryption Algorithm

There are a few warnings on StackOverflow about using `AES-128-ECB` and `AES-256-CBC`, so the one implemented is `AES-256-GCM` (it includes a signature, which seems like a good idea but I don't know much about cryptography).

The most important thing for security is to make sure the `nonce` is always different.

**Notes on the implementation:**
- The result is encoded in hex
- The first 12 bytes is the `nounce`
- Only the first `32 bytes` of the secret are used as the key, because the key must be `32 bytes` for `AES-256`.

I'm not sure what the `tag (optional) the public tag generated by the encoding (for instance in GCM)` means (in the issue).

## Review

Easier to review all at once. But as usual the individual commit messages contain more info.


app/api/users/generate_profile_edit_token.feature
smadbe, line 1: The tests do not really insure that the "check for the management relation" is well computed  as complex cases (`@Manager` could be manager of an ancestor of `@Class`, the user may be descendant of `Class`) are not considered.

app/encrypt/aes.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"io"
+)
+
+// AES256GCM encrypts the plaintext using AES-256-GCM with the provided key.
+// It returns the ciphertext with the nonce prepended.
```
smadbe, line 12: is prepending the nonce completely "standard"? It should probably be explain in the service doc

smadbe (approved):
looks good. I hope GCM will be easy to implement by Michel

app/encrypt/aes.go
GeoffreyHuck: Added the details in swagger doc. It's standard for implementations in `go` I think, but maybe not for everything. BTW if Michel has trouble with decrypting I can help, I made a few mistakes during the implementation so I know some potential mistakes.

app/api/users/generate_profile_edit_token.feature
GeoffreyHuck: fixed. To be sure, if `user` is in `Class`, and `Class` is in `School`, it's possible that `require_personal_info_access_approval`=`edit` is defined on `School` and not `Class`? If yes, then the query wasn't correct but now is.

GeoffreyHuck: fixed, added in doc.

PR #1088, opened 2024-06-10
Refactor existing and temp user session processes

### LINKED ISSUE #1022: Allow multiple session in parallel

GeoffreyHuck:
fixes #1022 

We now have the same behavior of `refreshToken` for temp user and normal user:
- For temp users, refresh used to create a new session, we now keep the same session and only create a new access token
- For temp users, refresh used to delete the old access token, we now delete only the expired tokens of the user

This is the last PR for the parallel sessions.

Also refactors what's related to testing users into a new file to simplify `testhelpers/steps_app_language.go`.

Easier to review everything at once. Details of the steps in the commit messages.

### DIFF
--- a/app/api/auth/refresh_access_token.feature
+++ b/app/api/auth/refresh_access_token.feature
@@ -49,21 +49,23 @@ Feature: Create a new access token
     And the response header "Set-Cookie" should be "<expected_cookie>"
     And logs should contain:
       """
-      Generated a session token expiring in 7200 seconds for a temporary user with group_id = 12
+      Refreshed a session token expiring in 7200 seconds for a temporary user with group_id = 12
       """
     And the table "sessions" should be:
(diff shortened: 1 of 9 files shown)


smadbe (approved):
👍

PR #1094, opened 2024-06-16
`refreshToken`: put the lock before the most recent token is retrieved

GeoffreyHuck:
If another refresh is already happening, the most recent token will have changed when the lock is released, and with the current code, it will be refreshed again.

The following process had an issue:
1. We start two refresh requests
2. The first one takes the lock and refreshes the token, while the second gets the most recent token of the session, then waits for the lock
3. The first request finishes, put the new token in the database, and release the lock
4. The second request gets the lock, refreshes the token again. But it shouldn't, because now there is a valid most recent token for the session!

With this change, 4. becomes: The second request acquires the lock, gets the most recent valid token for the session, notices it is too new to refresh, and simply returns it.

### DIFF
--- a/app/api/auth/refresh_access_token.go
+++ b/app/api/auth/refresh_access_token.go
@@ -50,37 +50,39 @@ func (srv *Service) refreshAccessToken(w http.ResponseWriter, r *http.Request) s
 	var expiresIn int32
 	apiError := service.NoError
 
-	sessionMostRecentToken := store.
-		AccessTokens().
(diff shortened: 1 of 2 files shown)


smadbe (approved):
> Note: maybe we need a test for that.

Yes, that seems very useful to prevent regression.

smadbe:
@zenovich could you take over this PR (i.e. check it and add the required tests)?

PR #1096, opened 2024-06-20
Update golangci-lint in Makefile when the installed version is wrong

zenovich:


smadbe (approved):
:+1: 

Try to always provide in the PR description what was the motivation behind a change, even minor.

PR #1098, opened 2024-07-02
`updateGroup` with strengthened approvals: fix "reinvite" doesn't properly create invitations

GeoffreyHuck:
`reinvite` didn't work, because we didn't add the entries in table `group_membership_changes`.

I also realized a "transition" system was developed to handle those kind of changes. So the service was updated to use this transition system. For consistence and to make sure we don't end up with inconsistent states that messes up with the transition system.

Now the invitations should work as expected because the transition system takes care of it.

Two slight changes from before when the approval are strengthened:
- with `empty`:  `leave_request` doesn't stay `leave_request` but leaves the group
- with `reinvite`: `leave_request` leaves the group and doesn't get re-invited

I think those are the expected behaviors.


smadbe (approved):
:+1:

PR #1099, opened 2024-07-08
Make sure the HTTP call is never made inside a database transaction.

### LINKED ISSUE #1081: Make sure the call of the propagation endpoint is not inside a database transaction
Make sure the HTTP call is never made inside a database transaction.

GeoffreyHuck:
fixes #1081 

Because we want to have fast transaction.

Now, we call the endpoint at the end of the current transaction.

### Tricky part: How to pass the config endpoint to the `DataStore`

That was the trickiest part.

The first question is: Do `DataStore` need to know about how to schedule the propagation? I choose that the answer was yes because ~~it was easier~~ the fact we do it after a transaction means it makes sense to be aware of transactions.

Also, `InTransaction` is called in many places, and it doesn't make sense at all to pass the endpoint as a parameter to this function, because most transactions are not about propagations at all.

What was done: the `request Context` was already passed to the `DataStore`, but it wasn't used. So the endpoint is now stored in the `request Context`, in a middleware, and retrieved from the `DataStore`. It's the easiest way I found.

#### Alternatives

1. Make the configuration available globally. It's always tempting, but if we start like this we might end up with more and more globals.
2. Have another method called `InTransactionWithPropagation` with the endpoint as parameter, that does the same as `InTransaction` but with a propagation. The downside is that we would have to know which one to use at the start of the transaction. It seems to me that the propagation is a notion that appears at a lower level than the service.

### Make sure the propagation types are unique when we call the endpoint

We also make the types of propagation, sent as a parameter when we call the propagation endpoint, unique. We don't have to, it would work without it, but it seems cleaner.

### Review

Easier to review commit by commit. Details in the commit messages.

### DIFF
--- a/app/database/data_store.go
+++ b/app/database/data_store.go
@@ -279,10 +279,6 @@ func (s *DataStore) WithForeignKeyChecksDisabled(blockFunc func(*DataStore) erro
 	})
 }
 
-func (s *DataStore) IsInTransaction() bool {
-	return s.DB.isInTransaction()
-}
-
 // WithNamedLock wraps the given function in GET_LOCK/RELEASE_LOCK.
 func (s *DataStore) WithNamedLock(lockName string, timeout time.Duration, txFunc func(*DataStore) error) error {
 	return s.withNamedLock(lockName, timeout, func(db *DB) error {
--- a/app/database/data_store.go
+++ b/app/database/data_store.go
@@ -176,10 +176,11 @@ func (s *DataStore) NewID() int64 {
 }
 
 type awaitingTriggers struct {
-	ItemAncestors  bool
-	GroupAncestors bool
-	Permissions    bool
-	Results        bool
+	ItemAncestors            bool
+	GroupAncestors           bool
+	Permissions              bool
+	Results                  bool
+	SchedulePropagationTypes []string
 }
 type dbContextKey string
 
@@ -202,6 +203,12 @@ func (s *DataStore) InTransaction(txFunc func(*DataStore) error) error {
 
 	triggersToRun := s.ctx.Value(triggersContextKey).(*awaitingTriggers)
 
+	if len(triggersToRun.SchedulePropagationTypes) > 0 {
+		types := triggersToRun.SchedulePropagationTypes
+		triggersToRun.SchedulePropagationTypes = []string{}
+
+		StartAsyncPropagation(s, s.Context().Value("propagation_endpoint").(string), types)
+	}
 	if triggersToRun.GroupAncestors {
 		triggersToRun.GroupAncestors = false
 		s.createNewAncestors("groups", "group")
@@ -220,6 +227,15 @@ func (s *DataStore) InTransaction(txFunc func(*DataStore) error) error {
 	}
 
 	return err
+}
+
+// SchedulePropagation schedules a run of the propagation for the given types after the transaction commit.
+func (s *DataStore) SchedulePropagation(types []string) {
+	s.mustBeInTransaction()
+
+	triggersToRun := s.DB.ctx.Value(triggersContextKey).(*awaitingTriggers)
+	// TODO: filter type for available types, and make sure we don't have duplicates, maybe even sort them
+	triggersToRun.SchedulePropagationTypes = append(triggersToRun.SchedulePropagationTypes, types...)
 }
 
 // ScheduleResultsPropagation schedules a run of ResultStore::propagate() after the transaction commit.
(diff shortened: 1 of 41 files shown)


smadbe:
I would prefer that @zenovich reviews this one.

zenovich:
> The first question is: Do DataStore need to know about how to schedule the propagation? I choose that the answer was yes because it was easier the fact we do it after a transaction means it makes sense to be aware of transactions.

I still believe that the answer is "no" as it was discussed days before on Slack. The reason is that DataStore is just a clever DB interface decorator. But probably we could rethink the meaning of DataStore.

And actually I need to study this a bit more from scratch, as, from my perspective (from what it looked like two years ago), mutating operations on the DB resulted in setting the DB into inconsistent state (because of the operations themselves and also because of MySQL triggers), the propagations fixed that by bringing the DB into the consistent state again. That was the reason of calling the propagations in the same transaction with the mutating operations.

So, first I would like to study the current situation.

@smadbe, is it an urgent task?

zenovich:
**The main thing:**
If we schedule a propagation outside of a transaction, we cannot guarantee that data modifications already committed to the DB will be followed a corresponding propagation scheduled. Indeed, let's consider a situation where, for instance, the AWS Lamba function gets killed because of a timeout right after it has committed the data modifications to the DB, but before it called another AWS Lambda function to trigger the propagation.

Such situations are really dangerous and hard to debug. They will definitely lead to production failures and DB inconsistency issues.

I would think 100500 times before doing this and finally would find another approach.

app/database/data_store.go
```diff
@@ -222,6 +231,14 @@ func (s *DataStore) InTransaction(txFunc func(*DataStore) error) error {
 	return err
 }
 
+// SchedulePropagation schedules a run of the propagation for the given types after the transaction commit.
+func (s *DataStore) SchedulePropagation(types []string) {
+	s.mustBeInTransaction()
+
+	triggersToRun := s.DB.ctx.Value(triggersContextKey).(*awaitingTriggers)
+	triggersToRun.SchedulePropagationTypes = utils.UniqueStrings(append(triggersToRun.SchedulePropagationTypes, types...))
```
zenovich, line 239: If the order of types doesn't matter, It would be much clearer to store map[string]struct{} in triggersToRun.SchedulePropagationTypes instead of creating utils.UniqueStrings() and tests for it.

If the order of types matters, UniqueString() breaks it anyway since map doesn't preserve the order of keys.

app/database/data_store.go
```diff
@@ -262,10 +279,6 @@ func (s *DataStore) WithForeignKeyChecksDisabled(blockFunc func(*DataStore) erro
 	})
 }
 
-func (s *DataStore) IsInTransaction() bool {
```
zenovich, line 265: It was a nice useful method.

app/database/propagation.go
```diff
@@ -43,20 +45,13 @@ func SchedulePropagation(store *database.DataStore, endpoint string, types []str
...
-				store.ScheduleResultsPropagation()
-
-				return nil
-			})
-			MustNotBeError(err)
-		}
+
+			return nil
+		})
+		mustNotBeError(err)
```
zenovich, line 55: Public methods cannot return errors via panic (see https://go.dev/doc/effective_go#recover)

app/database/propagation.go
```diff
@@ -1,24 +1,26 @@
...
 )
 
-const PropagationEndpointTimeout = 3 * time.Second
+const endpointTimeout = 3 * time.Second
 
-// SchedulePropagation schedules asynchronous propagation of the given types.
+// StartAsyncPropagation schedules asynchronous propagation of the given types.
 // If endpoint is an empty string, it will be done synchronously.
-func SchedulePropagation(store *database.DataStore, endpoint string, types []string) {
+func StartAsyncPropagation(store *DataStore, endpoint string, types []string) {
```
zenovich, line 15: I would move this method into app/service. It can be called right from the endpoint handler in case when the async propagation is chosen. The database package is for database-related things only.

app/utils/string_test.go
```diff
@@ -1,6 +1,8 @@
 package utils
```
zenovich, line 1: A package cannot be called 'utils' as all the packages should have meaningful names. 'utils' is a classic example of bad package names. We should do something with it (see https://go.dev/blog/package-names)

zenovich (changes requested):
All the above

zenovich:
Duplicate of #1298

Issue #1047, opened 2024-04-03
Make sure async propagation does not block other requests


smadbe:
## Motivations

We moved some propagation to async because there were too long. But of course, we have to make sure these propagation may not block other sync requests (which would wait for a 15min propagation to complete).

As propagation works by iteration, it would convenient if:
- propagation would commit from time to time: maybe commit at every iteration would make it slow... a rule may be "every iteration if >0.1sec has passed since the last one
- start the lock related to the iteration with low priority: the locking should be in low priority, not the queries in the lock

Note: that would also solve the problem of a propagation which would run for >15min and so which would not progress when killed after 15min by lambda.

## Subtasks 
* check what would be relevant and possible

GeoffreyHuck:
As discussed the other day, the propagation should schedule the run of another propagation if it's getting killed at 15min.

GeoffreyHuck:
# Notes on the problem

## Two processes that can cause the problem

1. A service is called that schedules a big propagation
2. Another service is called that wants to run a normally small propagation, but since the big propagation is in the prolongations tables, it ends up calling a big propagation, and timeout

or

1. A service schedule a big `async results` propagation (this one sets a `lock`)
2. The propagation is run is `async,` and the `lock` is taken
3. Another service wants to run the results propagation but cannot get the `lock`, and timeout

## Solutions

We can either:
- Make all prolongations `async`, or
- Allow to execute a propagation on a subset, but in this case we would also have to handle the the `lock` for the `results` propagation

We already talked about it on the phone and we decided to make all propagation `async`.

GeoffreyHuck:
# Tasks

- [ ] make all the propagation `async`, except the `groups ancestors` that is fast and which have services who needs the result right away
- [ ]  propagation would commit at the end of the step if >0.1sec has passed since the last one
- [ ]  start the lock related to the iteration with low priority: the locking should be in low priority, not the queries in the lock (need clarification)

PR #1100, opened 2024-07-08
Make all propagations async

GeoffreyHuck:
Related to #1047 (First step of: Make sure async propagation does not block other requests).

The `async` propagation, when executed `async` or `sync` when there's no endpoint or an error during its call, always execute the following propagations:
- `items_ancestors`
- `permissions`
- `results`

There were tests that checked things like "verify that the `results` propagation has been run but not the `permissions` propagation. Those tests have been updated because it cannot happen anymore.

Note: The `groups ancestor`propagation is an exception. It is always executed in sync because there are cases where its results are used in intermediate computations. Since it's very fast compared to the others, it shouldn't be a problem.

### Review

Easier to review commit by commit as there are a few steps. Details in commit messages.


smadbe (approved):
:+1:

Issue #1105, opened 2024-07-14
Make the test system easier to use and faster


zenovich:
# Our test system has two issues:
1) It produces too much output in the console. In the case when some test (especially some BDD scenario) fails, it's hard to find logs related to that particular test.
2) It's a bit slow. It takes much time to get all the tests pass before pushing.

# What I suggest to do in order to eliminate/reduce the mentioned issues:

## 1. Output.
I suggest suppressing output (stdout/stderr) of passed tests and BDD scenarios. Also, I suggest suppressing output of Godog related to passed scenarios / suites completely. For Godog, it can be done globally if we run each scenario as a separate Go test which is possible with contemporary Godog, while for other tests it can be done by adding one line of code at the top of each test producing the output. 

### Subtasks:
1) Upgrade Godog to the latest version (0.14.1)
2) Suppress output of passed Godog scenarios/suites.
3) Suppress output of passed Go tests (those of them that write to the console).

### Enable test results caching
Currently, in order to make it possible to run the needed tests run fast during working on a task, a developer should figure out (guess) which tests are affected by current code changes and run them manually one by one (filter them by specifying '-run REGEXP' for 'go test' or feature filename:line for Godog (or even by setting Godog tags in feature files)). All those things are time-consuming and require a lot of developer's attention.

On the other hand, as Golang is able to cache test results for packages when all the tests pass, we can employ this Golang feature in our project. The only thing why Golang doesn't cache test results for us now (when we run 'make test') is that we use '-race', '-coverpkg', '-coverprofile', and '-covermode' flags when running tests. These flags are needed to check for data races and collect the code coverage. While needed to test the final code before pushing to GitHub, it's unlikely all those flags are needed on each test run while working on a task. 

So, first, I suggest adding a new Makefile target for running tests while developing:
```
make test-dev
```

It will work in the same way as 'make test', but it will not collect code coverage or check for data races. This will enable per-package test results caching when all tests in a package pass.

### Split BDD features into sub-packages
As a second step, for the often situation where only one/several BDD scenarios fail leading to disabling caching for its whole package, especially when there are many BDD scenarios in that package taking much time to run, we can go even further and split our BDD features into separate sub-packages. This would enable more granular caching.

PR #1107, opened 2024-07-14
Upgrade Godog to v0.14.1, run BDD scenarios as sub-tests

zenovich:
As a subtask of #1105, in order to be able to suppress output of passed BDD scenarios in the future, here we upgrade Godog to the latest version (v0.14.1) and modify testhelpers.RunGodogTest() accordingly to run each BDD scenario as a separate Go test.


zenovich:
@GeoffreyHuck, note: this doesn't affect the test code coverage

smadbe (approved):
FYI @GeoffreyHuck

PR #1108, opened 2024-07-14
Suppress output of passed BDD scenarios and suites

zenovich:
As a subtask of #1105, we suppress output of passed BDD scenarios and suites.
Depends on #1107


smadbe (approved):
FYI @GeoffreyHuck

PR #1112, opened 2024-07-16
Use a named volume for the 'db' service in docker-complose.yml

zenovich:
Currently the DB storage (/var/lib/mysql) of the 'db' service (the database container) in docker-compose.yml is stored inside the container. Because of this, the database (and the data inside it) gets removed each time we recreate the container (to upgrade MySQL or by accident).

Here I suggest using a named volume for /var/lib/mysql which will be persistently stored outside of the container.

Note, like on any update before, running `docker compose down` and `docker compose up -d` to upgrade the configuration will delete the previously configured container (with all the db data) as we haven't used the named volume before.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,8 @@ services:
     ports:
       - "3306:3306"
       - "33060:33060"
+    volumes:
+      - mysql-volume:/var/lib/mysql
     environment:
       MYSQL_ROOT_PASSWORD: a_root_db_password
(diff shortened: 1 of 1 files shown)


smadbe (approved):
No need to add a sentence in the README to explain that?

zenovich:
> No need to add a sentence in the README to explain that?

Probably not, as it works out of the box.
May I merge?

smadbe:
yes

PR #1113, opened 2024-07-16
Upgrade mysql in docker-compose.yml to 8.0.28

zenovich:
As we have MySQL 8.0.28 in prod, it's time to upgrade it in docker-compose.yml

Note: Possibly it's better to apply this change in development environment after configuring the named volume for the DB storage introduced in #1112 . It's good to be able to upgrade the local MySQL fast, without recreating all the databases.


zenovich:
Depends on #1114

zenovich:
The issue is related to user_batches table. It contains a check constraint

CONSTRAINT `ck_user_batches_custom_prefix` CHECK (regexp_like(cast(`custom_prefix` as char charset binary),_utf8mb4'^[a-z0-9-]+$'))
which works fine on MySQL < 8.0.22, but is completely incompatible with MySQL 8.0.22+. After upgrading MySQL to 8.0.22+ (8.0.28 in our case), this table became inaccessible: it's impossible to select data from it, alter it, rename it, drop it:

mysql> select * from `user_batches`;
ERROR 3995 (HY000): Character set 'binary' cannot be used in conjunction with 'utf8mb4_0900_ai_ci' in call to regexp_like.
mysql> ALTER TABLE `user_batches` DROP CHECK `ck_user_batches_custom_prefix`;
ERROR 3995 (HY000): Character set 'binary' cannot be used in conjunction with 'utf8mb4_0900_ai_ci' in call to regexp_like.
mysql> show create table `user_batches`;
ERROR 3995 (HY000): Character set 'binary' cannot be used in conjunction with 'utf8mb4_0900_ai_ci' in call to regexp_like.
mysql> drop table `user_batches`;
ERROR 3995 (HY000): Character set 'binary' cannot be used in conjunction with 'utf8mb4_0900_ai_ci' in call to regexp_like.
The only way to deal with this is to create another table and use it instead of user_batches.

So, here, besides changing the version of MySQL in docker_compose.yml, we:
1. create table user_batches_new with the same structure as of user_batches,
2. fix old migrations related to user_batches to make them compatible with MySQL 8.0.28,
3. replace user_batches with user_batches_new everywhere in the code

smadbe (approved):
I think the 2nd commit is part of another PR and should not be there. But whatever. :-)

PR #1115, opened 2024-07-16
Add the major version number (v2) into the module path as it is not v1 anymore

zenovich:
Maybe it sounds strange, weird, and even wild, but the modules system of Go requires packages which major versions have  reached 2.0.0 to contain '/vN' at the end of their module import path (where N is the major version number). Otherwise, external modules cannot import our packages. For the same reason, https://pkg.go.dev/github.com/France-ioi/AlgoreaBackend stuck at v1.13.12 (our latest v1.x release). See https://go.dev/wiki/Modules#why-must-major-version-numbers-appear-in-import-paths for more information.

Here, as we are at version 2.21.0 already, we finalize bumping version to 2.x in terms of Go Semantic Import Versioning.


smadbe (approved):
😓
(cc @GeoffreyHuck )

PR #1116, opened 2024-07-16
invitationsView: optimize query

### LINKED ISSUE #1095: Slow request: get group invitation
Investigate why this request is slow on dev db: (2.5-4sec)
```
time curl 'https://dev.algorea.org/api/current-user/group-invitations' -H 'Authorization: Bearer <demousertoken>' 
```

Under some condition(s) running (non-parallel) UI tests (for instance on branch https://github.com/France-ioi/AlgoreaFrontend/tree/feature/join-by-code-approvals), the service sometimes takes up to **40sec** responding !!! (reproducible)

GeoffreyHuck:
fixes https://github.com/France-ioi/AlgoreaBackend/issues/1095

The query was slow because for each entry we did a `SELECT MAX()...` to make sure the record is the most recent one for `member_id, group_id,action`. Instead we now put this logic in the `FROM` (.Table()), so we know all records we check are the most recent ones.

No need for an additional index because we don't change the fields we filter on, we just change the order we do the filter.

### DIFF
--- a/app/api/currentuser/get_group_invitations.go
+++ b/app/api/currentuser/get_group_invitations.go
@@ -110,10 +110,16 @@ type groupWithApprovals struct {
 func (srv *Service) getGroupInvitations(w http.ResponseWriter, r *http.Request) service.APIError {
 	user := srv.GetUser(r)
 
-	query := srv.GetStore(r).GroupMembershipChanges().
+	query := srv.GetStore(r).
+		Table(`
+			(SELECT group_id, member_id, MAX(at) AS at, action, ANY_VALUE(initiator_id) AS initiator_id
+			   FROM group_membership_changes
+		    GROUP BY group_id, member_id, action
+			) AS latest_group_membership_changes
+		`).
 		Select(`
-			group_membership_changes.group_id,
-			group_membership_changes.at,
+			latest_group_membership_changes.group_id AS group_id,
+			latest_group_membership_changes.at AS at,
 			users.group_id AS inviting_user__id,
 			users.login AS inviting_user__login,
 			users.first_name AS inviting_user__first_name,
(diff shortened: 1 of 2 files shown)


app/api/currentuser/get_group_invitations.go
```diff
@@ -110,10 +110,16 @@ type groupWithApprovals struct {
 func (srv *Service) getGroupInvitations(w http.ResponseWriter, r *http.Request) service.APIError {
 	user := srv.GetUser(r)
 
-	query := srv.GetStore(r).GroupMembershipChanges().
+	query := srv.GetStore(r).
+		Table(`
+			(SELECT group_id, member_id, MAX(at) AS at, action, ANY_VALUE(initiator_id) AS initiator_id
```
zenovich, line 115: Do we really don't care which of possible `initiator_id` is chosen within rows having same (group_id, member_id, action)?

app/api/currentuser/get_group_invitations.go
```diff
@@ -110,10 +110,16 @@ type groupWithApprovals struct {
...
+			   FROM group_membership_changes
+		    GROUP BY group_id, member_id, action
+			) AS latest_group_membership_changes
```
zenovich, line 118: Why is it `latest_group_membership_changes`? Yes, it has MAX(at) instead of `at`, but `initator_id` may come from any row.

zenovich:
Unfortunately, I don't know how to reproduce the issue we are trying to eliminate here

app/api/currentuser/get_group_invitations.go
smadbe: I had the same remark.
Possibly there could be in practice a single invitation for a same user-group, but if that's the case, the condition is still a bit weirdy

zenovich:
The improved test passes on the master branch

zenovich:
This PR probably can be closed because of #1122

PR #1119, opened 2024-07-19
Introduce an in-memory MySQL in docker-compose.yml for running tests

zenovich:
Introduce an in-memory MySQL in docker-compose.yml for running tests and use it by default in conf/config.test.sample.yaml.

On my computer it noticeably speeds up tests (11m42s vs 14m08s).

### DIFF
--- a/conf/config.test.sample.yaml
+++ b/conf/config.test.sample.yaml
@@ -19,7 +19,7 @@ token:
 database:
   user: algorea
   passwd: a_db_password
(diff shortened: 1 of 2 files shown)


zenovich:
@GeoffreyHuck, fyi

zenovich:
As we use MySQL 8.0.28 here, this PR depends on #1113 and #1114

smadbe (approved):

PR #1130, opened 2024-07-25
Make ids of references in cucumber tests definite with the same sorting order as the references

zenovich:
Currently IDs for the references in cucumber tests are generated by rand.Int63() which has some disadvantages:
1. It's very hard to use `the table "*" should be` step when a column contains references, because the order of random ids is unpredictable, it requires trial and error method, such tests would be hard to support.
2. As the sorting order of random IDs changes randomly every time we add a new reference, we cannot rely on ordering table in `the table "*" should be` step by a column with references, such tests become fragile.

Here we introduce definite (not random) IDs for references. These definite IDs have the same sorting order as the references, so it becomes easy to use test steps depending on DB tables ordering with references like here:
```
    And the table "group_membership_changes" should be:
      | group_id | member_id | action                         | at                | initiator_id |
      | @Class   | @Student1 | removed_due_to_approval_change | {{currentTime()}} | @Teacher     |
      | @Class   | @Student2 | invitation_created             | {{currentTime()}} | @Teacher     |
      | @Class   | @Student3 | join_request_refused           | {{currentTime()}} | @Teacher     |
      | @Class   | @Student4 | join_request_refused           | {{currentTime()}} | @Teacher     |
```

Also, here we disallow implicit creation of new references which have been used in UserIsAManagerOfTheGroupWith() and theGroupIsADescendantOfTheGroup(). The reasons are that
1) Such implicitly created references are inaccessible from tests (there names contain spaces and other special characters).
2) Such implicitly created references would have ids out of the correct sorting order and there objects appear in the order DB tables in unexpected places hardening to check the results in the DB.
3) Implicit things make tests harder to understand. For instance, steps calling UserIsAManagerOfTheGroupWith() doesn't mention creation of an intermediate group at all.
Instead of creating something implicitly, it's always possible to create it explicitly, not hard at all.

After all, we add a new test step `^(@\w+) is a child of the group (@\w+)$` for convenient creation of group_group relations.

### DIFF
--- a/app/api/groups/get_permissions_can_request_help_to.feature
+++ b/app/api/groups/get_permissions_can_request_help_to.feature
@@ -19,7 +19,8 @@ Feature: Get permissions can_request_help_to for a group
       | @ClassParent             | @ClassParentParent       |          |
       | @Class                   | @ClassParent             |          |
       | @OtherSourceGroup        |                          |          |
-    And @Teacher is a manager of the group @Class and can grant group access
+    And @Teacher is a manager of the group @ClassAnotherParent and can grant group access
+    And @Class is a child of the group @ClassAnotherParent
     And there are the following items:
       | item                        | type    |
       | @ChapterParent              | Chapter |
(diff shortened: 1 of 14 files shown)


zenovich:
fyi, @GeoffreyHuck

smadbe (approved):

PR #1131, opened 2024-07-25
Refactor time-related template functions used in cucumber tests

zenovich:
1. Rename template functions used in cucumber tests for clarity:
- relativeTime -> relativeTimeDB (returns time relative to the current time in the DB format),
- timeToRFC -> timeDBToJS (converts time from DB format to JS format).
2. Add support for nanoseconds in timeDBToJS (will be used in next PRs).
3. Add a new function relativeTimeDBMs (similar to relativeTimeDB, but with milliseconds; will be used in next PRs).
4. Add new functions currentTimeDB & currentTimeDBMs returning the current time in the DB format (with and without milliseconds; will be used in next PRs).

### DIFF
--- a/app/api/currentuser/get_group_invitations.feature
+++ b/app/api/currentuser/get_group_invitations.feature
@@ -2,7 +2,7 @@ Feature: Get group invitations for the current user
   Background:
     Given the database has the following table 'groups':
       | id | type    | name                          | description                   | require_personal_info_access_approval | require_lock_membership_approval_until | require_watch_approval |
-      | 1  | Class   | Our Class                     | Our class group               | none                                  | {{relativeTime("+96h")}}               | true                   |
+      | 1  | Class   | Our Class                     | Our class group               | none                                  | {{relativeTimeDB("+96h")}}             | true                   |
       | 2  | Team    | Our Team                      | Our team group                | none                                  | null                                   | false                  |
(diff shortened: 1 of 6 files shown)


zenovich:
fyi, @GeoffreyHuck

smadbe (approved):

smadbe (changes requested):
Hum actually

> timeToRFC -> timeDBToJS (converts time from DB format to JS format).

It is not really a JS format. It converts the DB format to the RFCxyc format.

smadbe:
or iso8601 or just string format, but "JS format" makes no sense to me.

zenovich:
> Hum actually
> 
> > timeToRFC -> timeDBToJS (converts time from DB format to JS format).
> 
> It is not really a JS format. It converts the DB format to the RFCxyc format.

I know, but in the context of our cucumber tests it is the format we in our JSON response for time values. It's not clear which function to use and where. It's pain. I think the timeDBToJS is clearer, but maybe there are other options. There are lots of RFCs anyway.

zenovich:
It will be timeDBToRFC()

smadbe:
That's correct that RFC alone does not mean anything, it is  RFC 3339 for the date time format.

zenovich:
Can I rename it into timeDBToRFC3339 later in separate PR?

smadbe (approved):
> Can I rename it into timeDBToRFC3339 later in a separate PR?

I don't see the point, but yes

PR #1132, opened 2024-07-25
Improve cucumber error message when there are more rows than expected in a DB table

zenovich:
1. Print the row data,
2. Substitute IDs with corresponding references (if any),
3. Substitute current time with {{currentTimeDB()}} or {{currentTimeDBMs()}}

It looks like:
```
there are more rows in the SQL results than expected. expected: 3, the next row:
| @Class | @Student4 | join_request_refused | {{currentTime()}} | @Teacher |
```

(Note: the commit from #1130 is included here just to make the PR work)

Depends on #1131, #1130


zenovich:
fyi, @GeoffreyHuck

smadbe (approved):
I don't get why you do not have the other PR as base instead of including it in it, but anyway...

PR #1134, opened 2024-07-25
Store timestamps with milliseconds in group_membership_changes.at and group_pending_requests.at

zenovich:
For the case when there are multiple group_membership changes in the same second we still want to be able to order them by `at`. So

1. Add support for timestamps with fraction parts in query get parameters and pagination
2. Store milliseconds in group_membership_changes.at and group_pending_requests.at


app/api/groups/update_group_require_strengthened.feature
```diff
@@ -103,18 +103,20 @@ Feature:
...
-      | @Class | @Student4 | join_request  |
-      | @Class | @Student5 | invitation    |
-      | @Other | @Student5 | join_request  |
+    And the database has the following table 'group_pending_requests':
+      | group_id | member_id | type          | at                      |
+      | @Class   | @Student1 | leave_request | 2020-01-01 00:00:01.000 |
+      | @Class   | @Student3 | join_request  | 2020-01-01 00:00:03.000 |
+      | @Class   | @Student4 | join_request  | 2020-01-01 00:00:04.000 |
+      | @Class   | @Student5 | invitation    | 2020-01-01 00:00:05.000 |
+      | @Other   | @Student5 | join_request  | 2020-01-01 00:00:15.000 |
```
smadbe, line 119: having milliseconds everywhere pollutes the readability of tests.
As it is `.000`. And then if you get 6 decimals, you'll put '.000000' everywhere?
Seriously, when it is about setting values ("the database has the following table"), it should be as when you fill the db, if you do not mention the milliseconds, it assumes it is all 0s. Of course, the JSON output, you need to use the full returned value.

app/api/currentuser/create_group_leave_request.robustness.feature
```diff
@@ -18,8 +18,8 @@ Feature: User creates a request to leave a group - robustness
       | 15              | 21             | null                        |
     And the groups ancestors are computed
     And the database has the following table 'group_pending_requests':
-      | group_id | member_id | type          | at                  |
-      | 14       | 21        | leave_request | 2019-05-30 11:00:00 |
+      | group_id | member_id | type          | at                      |
+      | 14       | 21        | leave_request | 2019-05-30 11:00:00.001 |
```
smadbe, line 22: I have the impression that a single case (in robustness?) would be enough to demonstrate milliseconds... otherwise that pollutes the test. (larger column make it more difficult to review on GH for instance)

smadbe (commented):
It makes the collisions less likely but that may still happen. You still have to handle the case two cases equal.

app/api/groups/update_group_require_strengthened.feature
zenovich: We the opposite, in the JS we skip trailing zeros. Here, in tests I enter trailing zeros just because it will be used for comparing with rows returned by the DB which doesn't skip trailing zeros.

zenovich:
> It makes the collisions less likely but that may still happen. You still have to handle the case two cases equal.

Right, the collisions are still possible. There is no need to test it: it will not work, the test will be flaky as the order is not guaranteed.

We can handle this situation lately: we can join group_membership_changes and group_pending_request by (group_id, member_id, at). Here is the final preparation step to make it possible.

app/api/currentuser/create_group_leave_request.robustness.feature
zenovich: In robustness it's just to check the table stays unchanged, but the DB returns it with milliseconds, so it should be set with milliseconds to be equal

app/api/groups/update_group_require_strengthened.feature
zenovich: If we skip the trailing zeros, 'the table "*" should stay unchanged' step will not work as currently we compare values as strings.

PR #1138, opened 2024-07-29
Speed up itemActivityLogForItem service

### LINKED ISSUE #987: Analyze log service perf

zenovich:
1. Revert most of #1020 while keeping useful commits.
2. Create a correct index for the covering index range scan on `answers` for the case when there are more than 10,000 matching answers.
3. Make the query use this index.

Fixes #987


zenovich:
> Why reverting the test modifications done in #1020?

The test has been designed to check the ordering of the activities including those having the same timestamp, same timestamps and item_id, same timestamp, item_id, and participant_id, and so on. This way we test that the sorting by each column works correctly. Also, with the same data, the test checks the tie-breaking on pagination for each column.

At the same time, the change https://github.com/France-ioi/AlgoreaBackend/pull/1020/commits/50aa3a8a8bfb925eb57a9308b1e5043bacfa5c86#diff-af13c261947b5355b31a883266ec781272ac7780f39c9d3b97fd15e98c5fc505L645 sets different timestamp for each activity making the test ignoring sorting by columns other than the timestamp.

As the sorting/pagination functionality is important and prone to frequent breakdowns during service modifications (I know that because I modified this service several times), we have to restore the test.

Unfortunately, all other test-related commits are strongly dependent on the above change and cannot be kept for this reason. Sorry for that.

zenovich:
See some other comments in the commit messages

zenovich:
This PR doesn't affect the performance (checked)

smadbe (approved):
Ok, we'll try on the actual db

PR #1139, opened 2024-08-01
Fix item activity log services (display can_watch_answer correctly, show items visible to the team when 'as_team_id' is given).

### LINKED ISSUE #1025: Add watch item permission in the itemActivityLogForAllItems service

zenovich:
1. "can_watch_item_answer" added in #1027 didn't respect implicit permissions propagation from ancestor groups (see https://france-ioi.github.io/algorea-devdoc/items/access-rights/#propagation). So, when a user is a descendant of a group having can_watch_generated='answer' on an item, "can_watch_item_answer" was still false, it only respected direct permissions of the user. Here we fix this issue.
2. "can_watch_item_answer" added in #1027 was displayed even if no 'watch_group_id' was given which doesn't make sense. This PR removes "can_watch_item_answer" from the responses if "watch_group_id" is not given.
3. Here we rename "can_watch_item_answer" into "can_watch_answer" as there is no need to be so specific.
4. The service used to return items visible to the user (not to the team) when "as_team_id" was given. This PR eliminates this issue as well.

Also, some useful functions making code more expressive and concise are being introduced here (see the new package "golang").

Fixes #1025 (really)

### DIFF
--- a/app/api/items/get_activity_log.feature
+++ b/app/api/items/get_activity_log.feature
@@ -11,6 +11,8 @@ Feature: Get activity log
       | 13 | Class | Our Class  |
       | 20 | Other | Some Group |
       | 30 | Team  | Our Team   |
+      | 40 | Club  | Our Club   |
+      | 50 | Club  | Team Club  |
     And the database has the following table 'group_managers':
       | group_id | manager_id | can_watch_members |
       | 11       | 31         | true              |
@@ -22,6 +24,8 @@ Feature: Get activity log
       | 13              | 41             | 2019-05-30 11:00:00            |
       | 20              | 21             | null                           |
       | 30              | 21             | null                           |
+      | 40              | 21             | null                           |
+      | 50              | 30             | null                           |
     And the groups ancestors are computed
     And the database has the following table 'attempts':
       | id | participant_id |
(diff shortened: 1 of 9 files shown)


zenovich:
fyi @GeoffreyHuck

smadbe (approved):
> Here we rename "can_watch_item_answer" into "can_watch_answer" as there is no need to be so specific.

That breaks the API. Bit annoying. We'll have to pay attention for the deployment.

PR #1143, opened 2024-08-05
Restore mistakenly disabled "revive" linter

zenovich:
Here I restore `revive` linter and fix the issues it found.

As a part of this task I restore mistakenly renamed parameters and variables in testhelpers/steps_db.go (see https://github.com/France-ioi/AlgoreaBackend/pull/893/commits/41e7673e2f937c3f89c90da80b74a110eecba331). The problem is that when a function operates *godog.Table, a database table, and a list column names of Godog table at once, it should use very concise and specific variable names. We cannot have `func (ctx *TestContext) DBHasTable(table string, data *messages.PickleStepArgument_PickleTable)` because it looks like we have two tables as arguments. Instead, it should be `func (ctx *TestContext) DBHasTable(tableName string, data *messages.PickleStepArgument_PickleTable)`. A similar reason for having `columnNames` instead of `columns` because column!=columnName, table!=tableName and so on. And also all parameter names must be consistent within the file.

Also, here I rename functions and methods having weird/meaningless/vague/inverse names. The winners in this category are:
https://github.com/France-ioi/AlgoreaBackend/compare/linter_revive?expand=1#diff-f96817a3c49523e7fd931051e26100e752e659384495143f7635e0127386a7f7L152
and https://github.com/France-ioi/AlgoreaBackend/compare/linter_revive?expand=1#diff-330e9ce6d05b4d2288d8630f3207d9c93d62f603482bd49746f0256cd5c45efc


smadbe (approved):
:+1:

Issue #1148, opened 2024-08-05
Bug: /items/{item_id}/parents may return `null` results


smadbe:
# Bug

The `/items/{item_id}/parents` may return `null` value for the `results` key while the doc does not mention this option.  Either fix the doc or fix the behavior.

From what I see it typically happen when a new user (e.g. temp user) access a content he hasn't visited before. Calling the service a second time does not return `null` anymore.

zenovich:
@smadbe Can we close this one?

PR #1155, opened 2024-08-15
Rework nullable fields and descriptions of embedded fields in go-swagger spec + make items.type not null + refactoring

zenovich:
I. go-swagger:
1. Use our patched version of go-swagger to generate specs. This patched version preserves descriptions of fields with embedded structures and marks all pointers (except for those having JSON 'omitempty' option set) as nullable (x-nullable: true).
2. Convert the generated go-swagger specs to OpenAPI 3.0.0 preserving tags set on data fields with references and rename 'x-nullable' to 'nullable' to make it work with `redocly` spec viewer.
3. Employ 'redoc.standalone.js' of 'redocly' to browse the generated specs in `swagger-deploy`.
4. Update 'Makefile' and 'README.md'.
5. Get rid of unneeded 'Nullable' comments.

II. Make `items`.`type` NOT NULL.

III. Refactoring
1. Get rid of unnecessary utils.DateTime.
2. Move utils.Ptr() to golang.Ptr().
3. Refactor app/api/auth/create_access_token.go and auth/middleware.go to handle errors correctly and not provide additional information on internal server error to the user.
4. Allow panicking with APIError in API service handlers
5. Introduce a new method service.MustBeNoError()
6. Get rid of unnecessary utils.Capitalize()
7. Remove unnecessary rule exclusion from .golangci.yml

**IMPORTANT CHANGES (FIXED BUGS):**
1. In responses of `invitationsView`, `inviting_user.first_name` & `inviting_user.last_name` become nullable (they should be as corresponding columns are nullable in the DB).
2. In responses of `listThreads`, `participant.first_name` & `participant.last_name` are only shown if the current user has rights to view them (similarly to other services).


zenovich:
Fixes #1148

smadbe (approved):

PR #1158, opened 2024-08-19
update cors exposed headers

smadbe:
Update the exposed headers defined in our cors file:
- remove the "Link" header which is not used
- add the "Backend-Version" so that the frontend can read it if needed
- add the "Date" so that the client can easily compute a time offset with the server, useful for time related things (contest countdown, ...)


zenovich (approved):
Looks good

PR #1159, opened 2024-08-19
Return the modified result data (together with the linked attempt data) in the format of attemptsList response row on success in resultStart

### LINKED ISSUE #1156: Return result information when creating it
## Motivations

Currently we have no ways to guess some information about the `result` when creating it such as `allows_submissions_until`. So, each time we visit a content we haven't seen, we end up to do 'get-results' -> no result > 'start result' -> 'get-results', so with a 3x request delay before being able to fetch the content. If we can get the information with the start-result, we can "guess" what would 'get-results' return.

## Subtasks
* in the `/items/{ids}/start-result` service, on success, return as data the created result, in the format returned by `/items/{item_id}/attempts`[i].

zenovich:
1. service.Response is being converted into a generic to make the type of Data field specific. Otherwise, json.Marshal() ignores custom marshallers defined with pointer receivers for its fields as it is non-addressable resulting in marshalling database.Time as '{}'.
2. A new template function timeDBMsToRFC() for godog tests is being introduced to format RFC3339 time with milliseconds explicitly.
3. The existing template function timeDBToRFC() is being modified to return timestamps without the fraction part.
4. The resultStart service is being changed to return attemptsList response row on success.

Fixes #1156

### DIFF
--- a/app/api/answers/create_answer.go
+++ b/app/api/answers/create_answer.go
@@ -123,9 +123,9 @@ func (srv *Service) saveAnswerWithType(rw http.ResponseWriter, httpReq *http.Req
 
 	var result render.Renderer
 	if isCurrent {
-		result = service.UpdateSuccess(nil)
+		result = service.UpdateSuccess[*struct{}](nil)
 	} else {
-		result = service.CreationSuccess(nil)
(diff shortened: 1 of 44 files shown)


smadbe (approved):
:+1:

Issue #1164, opened 2024-08-29
Temp user deletion: add a delay before deleting


smadbe:
## Overview
Current the `delete-temp-users` delete all temp user whose sessions has expired (I assume). It is currently run once a week.
We would like to keep a bit longer these expired temp user (for instance 2 months) to be able to extract some stats about content usage and possibly analyze some abuses.

## Acceptance criteria
- [ ] add an argument to the `delete-temp-users` command so that only temp users with a session expired for more than this argument are actually deleted

PR #1166, opened 2024-08-30
Introduce --delay parameter of "delete-temp-users" command allowing to delay deletions of temporary users with expired sessions

zenovich:
1. Introduce --delay parameter of "delete-temp-users" command allowing to delay deletions of temporary users with expired sessions
2. Rework related tests.

### DIFF
--- a/app/database/user_store.go
+++ b/app/database/user_store.go
@@ -2,6 +2,7 @@ package database
 
 import (
 	"fmt"
+	"time"
 )
 
 // UserStore implements database operations on `users`.
@@ -16,7 +17,7 @@ func (s *UserStore) ByID(id int64) *DB {
 
 const deleteWithTrapsBatchSize = 1000
 
-// DeleteTemporaryWithTraps deletes temporary users who don't have active sessions.
+// DeleteTemporaryWithTraps deletes temporary users who don't have sessions or whose sessions expired more than `delay` ago.
 // It also removes linked rows in the tables:
 //  1. [`filters`, `sessions`, `access_tokens`]
 //     having `user_id` = `users.group_id`;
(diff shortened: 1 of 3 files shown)


zenovich:
Fixes #1164

app/database/user_store.go
```diff
@@ -16,7 +17,7 @@ func (s *UserStore) ByID(id int64) *DB {
 
 const deleteWithTrapsBatchSize = 1000
 
-// DeleteTemporaryWithTraps deletes temporary users who don't have active sessions.
+// DeleteTemporaryWithTraps deletes temporary users who don't have sessions or whose sessions expired more than `delay` ago.
```
smadbe, line 20: It means if we apply session cleanup (for whatever reason), we may delete temp users much earlier than expected.  Same if they logout (even if technically there is no interface for logout)
Hum I think that will not work, we need to find another criteria...
I would say that if it has no session, we use the `latest_login_at` value (or `latest_activity_at` but at the condition it is updated when we ask for token extension (probably not the case for now)).

smadbe (changes requested):

cmd/delete_temp_users.go
```diff
@@ -44,5 +58,8 @@ func init() { //nolint:gochecknoinits,gocyclo
 		},
 	}
 
+	deleteTempUsersCmd.Flags().DurationVar(&delay, "delay", 60*24*time.Hour,
+		"delay between expiration and deletion of temporary users")
```
smadbe, line 62: it should mention the unit!

zenovich: > it should mention the unit!

The unit is specified in the command line. It looks like "1440h" or "30m".

app/database/user_store.go
zenovich: 1. As you said, we don't have the "log out" feature which could delete sessions
2. When we delete old sessions (in createAccessToken()) we still keep 10 most recent sessions, so this will not affect deletion of temporary users.
So, I think the current solution will work.

zenovich: Also, we do not update `latest_login_at` on access token refreshing (as that's not considered as new logging in).
For `latest_activity_at`, it is only updated in itemGetHintToken, itemGetAnswerToken, accessTokenCreate (but not for temporary users), userDataRefresh, itemTaskTokenGenerate, so not enough to rely on this column.

smadbe: > When we delete old sessions (in createAccessToken()) we still keep 10 most recent sessions, so this will not affect deletion of temporary users.

But that's a temporary assumption. We could choose to completely delete expired session... and if the new session is created afterwards and fails, we may have no sessions.

smadbe (approved):

PR #1171, opened 2024-09-15
Introduce a new config setting "server.disable_results_propagation" allowing to disable the results propagation everywhere

zenovich:
1. Introduce a new config setting "server.disable_results_propagation" allowing to disable the results propagation everywhere for all commands of the app
2. Remove unneeded command line flag --disable-results-propagation of the "propagation" command.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -5,6 +5,7 @@ import (
 	crand "crypto/rand"
 	"encoding/binary"
 	"fmt"
+	"net/http"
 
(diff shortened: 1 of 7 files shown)


zenovich:
It breaks many endpoints when set for the server, but should work fine for the 'propagation' command.
It can be enabled via env var: ALGOREA_SERVER__DISABLERESULTSPROPAGATION=true

smadbe (approved):

PR #1177, opened 2024-09-20
Rework processing of results_propagate_items table

zenovich:
1) As the only place we insert new items into results_propagate_items is the `before_delete_items_items` trigger which inserts `OLD.parent_item_id` there in order to have its results recomputed, it doesn't make sense to mark such results as 'to_be_propagated' as we have been doing until now. We should always mark such results as 'to_be_recomputed'.

2) As the items in the table are supposed to be marked as 'to_be_recomputed', the table name seems to be misleading. Instead of 'results_propagate_items', it should be 'results_recompute_for_items'. Here we rename it.

3) The currently deployed code inserts results by chunks, but it can work incorrectly when a concurrent DB session is inserting new results for marked items into the results table. In this case, some results can be mistakenly silently skipped.

4) Added the missing test.

As marking the results for parent items of deleted items_items to be recomputed has not been working correctly for VERY long time, **upon deployment of this PR**, we should recompute all the results linked to chapters and skills. It can be done by executing:
```
insert into results_recompute_for_items (item_id) select id from items where type IN ('Skill','Chapter');
```
and running the 'propagation'/'db-recompute' command after that.

**NOTE:** Recomputing all the results linked to chapters and skills will lock the DB for several minutes (because of the UPDATE query in recomputeResultsMarkedAsToBeRecomputedAndMarkThemAsToBePropagated()), as there will be an unsupported situation where there will be too many results marked as to 'to_be_recomputed' at once. This situation cannot occur normally.

### DIFF
--- a/app/database/result_store_propagate.go
+++ b/app/database/result_store_propagate.go
@@ -48,7 +48,7 @@ func (s *ResultStore) propagate() (err error) {
 	// Use a lock so that we don't execute the listener multiple times in parallel
 	mustNotBeError(s.WithNamedLock(resultsPropagationLockName, resultsPropagationLockWaitTimeout, func(s *DataStore) error {
 		CallBeforePropagationStepHook(PropagationStepResultsInsideNamedLockInsertIntoResultsPropagate)
-		setResultsPropagationFromResultsPropagateItems(s)
+		setResultsPropagationFromTableResultsRecomputeForItems(s)
 
 		// Initially there can be results of any kind
 		for {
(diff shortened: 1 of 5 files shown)


smadbe:
> As marking the results for parent items of deleted items_items to be recomputed has not been working correctly for VERY long time, **upon deployment of this PR**, we should recompute all the results linked to chapters and skills. It can be done by executing:
> 
> ```
> insert into results_recompute_for_items (item_id) select id from items where type IN ('Skill','Chapter');
> ```
> 
> and running the 'propagation'/'db-recompute' command after that.
> 
> **NOTE:** Recomputing all the results linked to chapters and skills will lock the DB for several minutes (because of the UPDATE query in recomputeResultsMarkedAsToBeRecomputedAndMarkThemAsToBePropagated()), as there will be an unsupported situation where there will be too many results marked as to 'to_be_recomputed' at once. This situation cannot occur normally.

For that part, I cannot take the risk of marking all of them at once... probably I can do my chunks with LIMIT-OFFSET.

smadbe (approved):

zenovich:
LIMIT-OFFSET is risky. It can skip some rows again

PR #1180, opened 2024-09-30
Speed up the results propagation and make it less locking + Introduce a command recomputing all the results of chapters/skills

zenovich:
1. Speed up the results propagation by propagating only results modified during recomputation instead of propagating all the recomputed results. It makes the results propagation an order of magnitude faster.
2. Split the results recomputation into atomic chunks of 1000 results instead of recomputing all results marked as 'to_be_recomputed' at once. This way, the results propagation will not lock the DB no matter how many results marked as 'to_be_recomputed' will be there.
3. Introduce a new command 'recompute-results' allowing to recompute results of all the chapters/skills one by one without locking the DB.

After this task is deployed, it will finally be possible to finish the deployment of #1177 by simply running 'AlgoreaBackend db-recompute'.

### DIFF
--- a/app/database/result_store_propagate.go
+++ b/app/database/result_store_propagate.go
@@ -7,9 +7,10 @@ import (
 )
 
 const (
-	resultsPropagationLockName        = "listener_propagate"
-	resultsPropagationLockWaitTimeout = 10 * time.Second
-	resultsPropagationChunkSize       = 200
+	resultsPropagationLockName               = "listener_propagate"
+	resultsPropagationLockWaitTimeout        = 10 * time.Second
+	resultsPropagationPropagationChunkSize   = 200
+	resultsPropagationRecomputationChunkSize = 1000
 )
 
 // propagate recomputes fields of results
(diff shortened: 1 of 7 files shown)


app/database/result_store_propagate.go
```diff
@@ -28,7 +29,8 @@ const (
 //  5. We process all objects that are marked as 'to_be_recomputed' and that have no children marked as 'to_be_recomputed'.
 //     Then, if an object has children, we update
 //     latest_activity_at, tasks_tried, tasks_with_help, validated_at.
-//  6. We mark all results marked as 'to_be_recomputed' as 'to_be_propagated'.
+//  6. We mark all modified results marked as 'to_be_recomputed' as 'to_be_propagated' and
+//     unmark all unchanged results marked as 'to_be_recomputed'.
```
smadbe, line 33: shouldn't you "text workflow" mention this new chunk size ?

smadbe (approved):
:+1:  I haven't spent enough time on it to validate 100% that it always works.

app/database/result_store_propagate.go
zenovich: fixed

PR #1183, opened 2024-10-08
answerGet: extend visibility to allow thread viewers and users having validated results on the task to view answers

### LINKED ISSUE #1042: Get-answer: extend visibility to thread viewer and user who validated the task
## Motivations
Currently, only owners and observers (with answer watch perm on the item and watch perm on the participant) could get an answer (by its id).

We have two new use cases where we would like users to be able to view an answer:
- a user who has `can_watch: result` on the item, can watch the participant and who has validated the activity
- with the forum, some other users may know about the answer (to help the user), and so should be able to get an answer by id

## Subtasks

- [ ] allow a user who has `can_watch: result` on the item, who can watch the participant and who has validated the activity, to view an answer
- [ ] Make sure a user who sees a thread, can see answers it may mention, so in practice apply the permission as the thread reading permissions [here](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/#listing--reading-a-thread) 

## Notes

- The forum part should be similar to  [the service](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v2.16.0.html#tag/threads/operation/threadGet). However, the perm rule has just changed and the issue fixing it is not started yet #1163 
- The 4 rules are not independent, possibly it is possible to "optimize" a lot by removing overlapping rules

zenovich:
Fixes #1042 
Depends on #1182

Also, the get_answer.feature has been completely rewritten here to make it maintainable.

### DIFF
--- a/app/api/answers/get_answer.go
+++ b/app/api/answers/get_answer.go
@@ -20,17 +20,16 @@ import (
 //				- be a member of the `answers.participant_id` team or
 //				  `answers.participant_id` should be equal to the user's self group.
 //
-//			- If the user is an observer/helper
+//			- If the user is an observer (a manager with `can_watch_members` of an ancestor of `answers.participant_id` group)
 //				- (s)he should have `can_watch` >= 'answer' permission on the `answers.item_id` OR
 //				- `can_watch` >= 'result' permission on the `answers.item_id` together with a validated result
 //	        (personally or as a team) on the `answers.item_id`.
 //
-//				The user is considered as an observer if (s)he is a manager with `can_watch_members`
-//		 		of an ancestor of `answers.participant_id` group.
-//
-//			  The user is considered as a helper if (s)he is a descendent of the thread's `helper_group_id`
-//	      for the `answers.participant_id`-`answers.item_id` pair while the thread is active or
-//	      closed less than 2 weeks ago.
+//			- If the user is a thread reader (when the thread for the `answers.participant_id`-`answers.item_id` pair exists)
+//				- (s)he should have `can_watch` >= 'answer' permission on the `answers.item_id` OR
+//				- (s)he should be a descendent of the thread's `helper_group_id` and have `can_watch` >= 'result' permission
+//				  on the `answers.item_id` together with a validated result (personally or as a team) on the `answers.item_id`
+//				  while the thread should be active or closed less than 2 weeks ago.
 //
 //			If any of the preconditions fails, the 'forbidden' error is returned.
 //		parameters:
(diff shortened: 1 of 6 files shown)


app/api/answers/get_answer.go
```diff
@@ -10,38 +10,46 @@ import (
...
-//		"401":
-//			"$ref": "#/responses/unauthorizedResponse"
-//		"403":
-//			"$ref": "#/responses/forbiddenResponse"
-//		"500":
-//			"$ref": "#/responses/internalErrorResponse"
+//				The user is considered as an observer if (s)he is a manager with `can_watch_members`
+//		 		of an ancestor of `answers.participant_id` group.
+//
+//			  The user is considered as a helper if (s)he is an ancestor of the thread's `helper_group_id`
```
smadbe, line 31: ancestor? a descendant!

app/api/answers/get_answer.go
```diff
@@ -10,38 +10,46 @@ import (
...
-//				`answers.participant_id` should be equal to the user's self group.
+//			- If the user is a participant
+//				- (s)he (or one of his/her teams) should have at least 'content' access rights to the `answers.item_id` and
+//				- be a member of the `answers.participant_id` team or
+//				  `answers.participant_id` should be equal to the user's self group.
 //
-//		- If the user is an observer
-//			- (s)he should have `can_watch` >= 'answer' permission on the `answers.item_id` and
-//			- be a manager with `can_watch_members` of an ancestor of `answers.participant_id` group.
+//			- If the user is an observer/helper
```
smadbe, line 23: I think there is still a helper case missing: the user have can_watch>=answer permission on the item but no observation on the user, no validation. It does not match your criteria for being an observed or a helper.

smadbe (changes requested):
First remarks on the specs (code may be wrong, tests may be missing)

app/api/answers/get_answer.go
zenovich: The user is a manager of an ancestor of `answers.participant_id`. So he manages `answers.participant_id` as well. Sounds correct, doesn't it?

smadbe: From the doc "be part of the group the participant has requested help to".

`helper_group_id` is the group of person who can help... so the current user must be part of it.. so descendant

zenovich: fixed

app/api/answers/get_answer.go
zenovich: Fixed:
Now we allow viewing the answer if the thread exists AND the user has `can_watch>=answer` on the item. We do not check if either the user is a descendant of the helper_group_id or if the thread is expired in this case.

The purpose of checking if the thread exists in this case is not clear, but anyway...

app/api/answers/get_answer.go
```diff
@@ -10,38 +10,45 @@ import (
...
-//			"$ref": "#/responses/badRequestResponse"
-//		"401":
-//			"$ref": "#/responses/unauthorizedResponse"
-//		"403":
-//			"$ref": "#/responses/forbiddenResponse"
-//		"500":
-//			"$ref": "#/responses/internalErrorResponse"
+//			- If the user is a thread reader (when the thread for the `answers.participant_id`-`answers.item_id` pair exists)
+//				- (s)he should have `can_watch` >= 'answer' permission on the `answers.item_id` OR
+//				- (s)he should be a descendent of the thread's `helper_group_id` and have `can_watch` >= 'result' permission
```
smadbe, line 30:
```suggestion
//				- (s)he should be a descendant of the thread's `helper_group_id` and have `can_watch` >= 'result' permission
```

smadbe (approved):
:+1: Ok, we'll mainly see when we'll use if it fully works as expected.

PR #1187, opened 2024-10-11
Retry DB transactions on lock wait timeout errors similarly to retrying on deadlocks

zenovich:
As stated in https://dev.mysql.com/doc/refman/8.0/en/innodb-error-handling.html,

> Both deadlocks and lock wait timeouts are normal on busy servers and it is necessary for applications to be aware that they may happen and handle them by retrying. You can make them less likely by doing as little work as possible between the first change to data during a transaction and the commit, so the locks are held for the shortest possible time and for the smallest possible number of rows. Sometimes splitting work between different transactions may be practical and helpful.


zenovich:
> It would be a good idea to set innodb_lock_wait_timeout=1 (or similar small number) on prod before deploying this change. Otherwise, the clients may wait for too long and we may have lots of waiting MySQL sessions on the DB server.

(but I'm not sure it's required)

smadbe (approved):
Didn't I left a review already ? it wrote that it may be useful to print to logs in such a case so that we may detect strange behavior.

zenovich:
> Didn't I left a review already ? it wrote that it may be useful to print to logs in such a case so that we may detect strange behavior.

fixed

PR #1190, opened 2024-10-12
Store the DB time instead of the server time in threads.latest_update_at and compare its value with the DB time instead of the server time as well

zenovich:
For the sake of consistency with other services, the thread-related services should store the DB time in threads.latest_update_at (not the server time). Also they should compare threads.latest_update_at with the DB time (not the server time).


smadbe (approved):
:+1:

PR #1191, opened 2024-10-13
Introduce a new method database.DB.With(), rework database.DB.Union() & database.DB.UnionAll()

zenovich:
1. Introduce a new method database.DB.With() allowing to add a CTE (common table expression) to the query.
2. Fix a bug in database.DB.Raw() causing "FOR UPDATE" to be duplicated sometimes in queries.
3. Modify database.DB.Union() and database.DB.UnionAll() methods to accept the "query" parameter as *database.DB instead of interface{}.
4. Rework the places where we use DB.Raw() to add CTEs into queries by employing the new DB.With() method there.


smadbe (approved):
:+1:

PR #1193, opened 2024-10-13
Speed up groupRootsView

zenovich:
This speeds up the service groupRootsView five times.

1. As preparation, get rid of selectGroupsDataForMenu() function. It's really inconvenient and hard-to-maintain.
2. Speed up the service groupRootsView:
- Consider only ancestors of managed groups excluding users (as we exclude them on the top level anyway), i.e. do not use GroupStore.ManagedUsersAndAncestorsOfManagedGroupsForGroup() for that.
- Do not use "SELECT ... FROM `groups` ... WHERE groups.id IN(...) OR groups.id IN(...)" construct for getting the combined list of ancestors of managed and joined groups as this leads to the full scan on the `groups` table. Instead, use a CTE made up of UNION of two subqueries.
- Remove unnecessary WHERE conditions.

### DIFF
--- a/app/api/groups/get_group.go
+++ b/app/api/groups/get_group.go
@@ -141,6 +141,34 @@ func (srv *Service) getGroup(w http.ResponseWriter, r *http.Request) service.API
 	store := srv.GetStore(r)
 
 	query := store.Groups().PickVisibleGroups(store.Groups().DB, user).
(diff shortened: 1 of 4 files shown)


smadbe (approved):
:+1: :+1:

PR #1204, opened 2024-10-23
Optimize the SQL query in groupChildrenView: do not count all the children of a group to determine if it is empty or not

zenovich:
Just a small improvement


smadbe (approved):
:+1:

PR #1206, opened 2024-10-25
Make all the services compatible with transactions retrying and cover this feature with tests for all services + make error messages related to JSON parsing clearer

### LINKED ISSUE #1196: Error 400 "EOF" on item PUT
As reported and discussed on 16/10/24 on Slack, the `PUT /item/:id` may be return a 400 error with "EOF" on error message while the input JSON is valid. 

More details on slack...

zenovich:
1. Make cucumber tests implicitly verify that all the services work fine when their transactions are retried.
2. Fix all the services failing when their transactions are retried.
3. Sleep 50ms instead of 1s on average before retrying a transaction.
4. Make error responses a bit more clear when JSON input is broken.
5. Add some obvious improvements related to row locking.

Fixes #1196


smadbe (approved):
:+1:

PR #1210, opened 2024-11-07
Make groups&items propagations twice faster, do not join groups table to determine group memberships, do not mark new items for propagation, run/schedule propagations in GroupGroupStore.Transition() only when needed + testing-related fixes

zenovich:
1. Make groups&items propagations twice faster:
- introduce a new column groups_groups.is_team_membership which equals to 1 when a parent group is a team, set its value automatically on groups_groups creation,
- move the 'processing' state into a new temporary table,
- always filter *_propagate tables by exact values of ancestors_propagation_state,
- perform covering index lookups on groups_groups instead of index lookups (create an index for this).
2. Do not join the groups table to determine if group memberships (use the new groups_groups.is_team_membership column instead).
3. Do not mark new items for propagation, only call the items propagation in itemCreate if at least one items_items row is created.
4. Disallow modifying groups.type and groups_groups.is_team_membership.
5. Change the DB username for migration-tests to match the definer of triggers in the dump file.
6. Sort all the data returned by currentUserDataExport & currentUserFullDataExport to make tests stable.
7. Do not run/schedule propagations in GroupGroupStore.Transition() when the parent group is a team, schedule the results propagation in GroupGroupStore.Transition() only if a new relation is inserted and the parent group is not a team.

### DIFF
--- a/db/migrations/2410282334_add_trigger_before_update_groups_to_disallow_modifying_type.sql
+++ b/db/migrations/2410282334_add_trigger_before_update_groups_to_disallow_modifying_type.sql
@@ -1,0 +1,11 @@
+-- +migrate Up
+-- +migrate StatementBegin
+CREATE TRIGGER `before_update_groups` BEFORE UPDATE ON `groups` FOR EACH ROW BEGIN
+  IF OLD.`type` != NEW.`type` THEN
+    SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = 'Unable to change immutable groups.type';
+  END IF;
+END
+-- +migrate StatementEnd
+
+-- +migrate Down
+DROP TRIGGER `before_update_groups`;
--- a/db/migrations/2410282334_add_trigger_before_update_groups_to_disallow_modifying_type.sql
+++ b/db/migrations/2410282334_add_trigger_before_update_groups_to_disallow_modifying_type.sql
@@ -1,8 +1,8 @@
 -- +migrate Up
 -- +migrate StatementBegin
 CREATE TRIGGER `before_update_groups` BEFORE UPDATE ON `groups` FOR EACH ROW BEGIN
-  IF OLD.`type` != NEW.`type` THEN
-    SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = 'Unable to change immutable groups.type';
+  IF OLD.`type` != NEW.`type` AND (OLD.`type` IN ('User', 'Team') OR NEW.`type` IN ('User', 'Team')) THEN
+    SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = 'Unable to change groups.type from/to User/Team';
   END IF;
 END
 -- +migrate StatementEnd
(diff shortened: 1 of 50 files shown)


db/migrations/2410292333_remove_unnecessary_enum_values_from_groups_propagate_ancestors_computation_state.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+ALTER TABLE `groups_propagate` MODIFY COLUMN `ancestors_computation_state` ENUM('todo', 'done') NOT NULL;
```
smadbe, line 2: As I already mentioned/asked in the past: Is it really better to have a todo/done and possibly not an entry for any existing group, versus no "state" column, just considering an entry to mean "todo". So basically deleting the rows instead of marking them "done"? It would allow the table to stay very small instead of converging to a size similar to the `groups` table. Maybe updating all is easier than deleting some and updating some ?

db/migrations/2410292340_remove_unnecessary_enum_values_from_items_propagate_ancestors_computation_state.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+ALTER TABLE `items_propagate` MODIFY COLUMN `ancestors_computation_state` ENUM('todo', 'done') NOT NULL;
```
smadbe, line 2: same here

smadbe (approved):

db/migrations/2410292333_remove_unnecessary_enum_values_from_groups_propagate_ancestors_computation_state.sql
zenovich: Although it's very easy to implement now, I believe it would make the propagations slower because we would increase fragmentation of *_propagate table each time we delete rows from them. Currently we mostly update existing rows and never delete them preventing the fragmentation.

db/migrations/2410282334_add_trigger_before_update_groups_to_disallow_modifying_type.sql
```diff
@@ -0,0 +1,11 @@
+-- +migrate Up
+-- +migrate StatementBegin
+CREATE TRIGGER `before_update_groups` BEFORE UPDATE ON `groups` FOR EACH ROW BEGIN
+  IF OLD.`type` != NEW.`type` THEN
+    SIGNAL SQLSTATE '45000' SET MESSAGE_TEXT = 'Unable to change immutable groups.type';
```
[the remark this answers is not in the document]
zenovich: I'm going to fix this now. BTW, there are no endpoints allowing to change a group type currently.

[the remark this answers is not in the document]
zenovich: fixed

PR #1227, opened 2024-12-11
Return unlocked items in saveGrade + other fixes

### LINKED ISSUE #1224: Save the grade: return the unlocked content
## Motivations
We need to notify somehow the user that the score he has achieved has given him access to more content.

## Note
I assume that requires the score to be computed immediately... and of course, we should not require to compute all pending scores but just the one which has just been changed.

## Subtasks
- [ ] Add to the save grade service output, the list of item which has been newly unlocked by that submission. 
- [ ] The list should contains at least the `type` of item, its `id` and its `title` (as usual, in the user's default language or if not available, in the default item lang), and the `language_tag` of this title

zenovich:
1. Return unlocked items in saveGrade
2. Upgrade go-sql-driver/sql to v1.8.1
3. Fix redocly/cli at v1.25.15 (v1.26 doesn't work with our generated schema)
4. Call PingContext() instead of Ping() to handle timeouts on pinging the DB
5. Fix setting can_enter_until on items unlocking: set can_enter_until to +inf when the unlocked item requires explicit entry and the current value of can_enter_until is not +inf.

Fixes #1224

### DIFF
--- a/app/api/items/save_grade.feature
+++ b/app/api/items/save_grade.feature
@@ -640,10 +640,10 @@ Feature: Save grading result
       | 0  | 101            |
       | 1  | 101            |
     And the database table "items" also has the following row:
-      | id  | platform_id | url                                                                     | validation_type | default_language_tag |
-      | 80  | 20          | http://taskplatform1.mblockelet.info/task.html?taskId=4034495436721840  | All             | fr                   |
-      | 90  | 20          | http://taskplatform1.mblockelet.info/task.html?taskId=4034495436721841  | All             | fr                   |
-      | 100 | 20          | http://taskplatform1.mblockelet.info/task.html?taskId=4034495436721842  | All             | fr                   |
+      | id  | type    | validation_type | default_language_tag |
+      | 80  | Chapter | All             | fr                   |
+      | 90  | Chapter | All             | fr                   |
+      | 100 | Chapter | All             | de                   |
     And the database table "item_dependencies" also has the following row:
       | item_id | dependent_item_id | score |
       | 60      | 80                | 0     |
@@ -706,8 +706,8 @@ Feature: Save grading result
             },
             {
               "item_id": 100,
-              "language_tag": null,
-              "title": null,
+              "language_tag": "de",
+              "title": "Kapitel C",
               "type": "Chapter"
             }
           ]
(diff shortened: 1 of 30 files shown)


app/api/items/save_grade.feature
```diff
@@ -593,9 +625,107 @@ Feature: Save grading result
...
+
+  Scenario: Unlocks multiple items recursively
+    Given I am the user with id "101"
+    And the database has the following table "attempts":
+      | id | participant_id |
+      | 0  | 101            |
+      | 1  | 101            |
+    And the database table "items" also has the following row:
+      | id  | platform_id | url                                                                     | validation_type | default_language_tag |
+      | 80  | 20          | http://taskplatform1.mblockelet.info/task.html?taskId=4034495436721840  | All             | fr                   |
```
smadbe, line 644: if 80 is a chapter, it is not supposed to have a url and a platform id. shouldn't it be the "type" column in `items` table definition as it changes things for the propagation?

app/database/permission_granted_store.go
```diff
@@ -65,3 +69,7 @@ func (s *PermissionGrantedStore) EditIndexByName(name string) int {
 func (s *PermissionGrantedStore) EditNameByIndex(index int) string {
 	return s.PermissionNameByKindAndIndex("edit", index)
 }
+
+func (s *PermissionGrantedStore) permissionsPropagateTableName() string {
+	return golang.IfElse(s.arePropagationsSync(), "permissions_propagate_sync", "permissions_propagate")
```
smadbe, line 74: So 2 propagations cannot run in parallel when "sync"? Except for this table name, was it an actual requirement?

smadbe (approved):

app/database/permission_granted_store.go
zenovich: Nothing stops sync propagations from being run concurrently. The *_sync tables are processed inside transactions, concurrent sessions don’t see rows added by other sessions.

app/api/items/save_grade.feature
zenovich: The propagation never looks at items.type

zenovich: Fixed the test

PR #1231, opened 2024-12-14
Add user_id into "request complete" logs

### LINKED ISSUE #1226: Include the user id in the log
## Motivations

- When we debug a problem (one request without body payload), it is often not possible to reproduce a problem because we do not know what user initiated the request

- We would like to extract some usage data based on logs, but we need to be able to aggregate by user somehow 

## Subtasks
* Add the authenticated user'id in the logs written by the backend (only in "request complete" as "request started" does not have the info I assume)

zenovich:
Add user_id into "request complete" logs for all the services, even for those that doesn't use the auth.UserMiddleware.
Also, make cucumber tests checking logs able to see the whole logs (not only messages) and make such tests independent of log formatter chosen in the config.

Fixes #1226

### DIFF
--- a/app/api/answers/submit.feature
+++ b/app/api/answers/submit.feature
@@ -73,6 +73,10 @@ Feature: Submit a new answer
         "success": true
       }
       """
+    And logs should contain:
+      """
+      user_id=101
+      """
     And the table "answers" should be:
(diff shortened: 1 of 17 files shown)


smadbe (approved):
:+1:

PR #1232, opened 2024-12-26
Adapt threads-related services to the recent forum permission rules change + fix some bugs related to permissions checking there

### LINKED ISSUE #1163: Adapt to forum permission change
## Overview
The permission rules for writing / listing a forum have slightly change. Adapt the code.

## Acceptance criteria

Changes are these one: https://github.com/France-ioi/algorea-devdoc/commit/dd96b1935c275c01df842ba61a60573c58a7202a
- [ ] adapt listThreads service
- [ ] adapt threadUpdate service
- [ ] adapt threadGet service

zenovich:
+ Apply changes described in #1163.
+ Fix thread visibility checking in listThreads (see https://github.com/France-ioi/AlgoreaBackend/issues/1163#issuecomment-2497715563).
+ Require the current user to have can_view>=content permission on item_id if item_id is given in listThreads (return 403 otherwise).
+ Require the current user to have can_view>=content permission on the item in threadGet (return 403 otherwise).

Fixes #1163


zenovich:
@smadbe, threadUpdate doesn't require the current user to have can_view>=content permission on the thread's item. As the result, a thread's participant can update the thread even if he is not allowed to view content of the thread's item. I'm not sure, but it looks like a bug. Should we fix it?

smadbe (approved):

smadbe:
> @smadbe, threadUpdate doesn't require the current user to have can_view>=content permission on the thread's item. As the result, a thread's participant can update the thread even if he is not allowed to view content of the thread's item. I'm not sure, but it looks like a bug. Should we fix it?

Yes probably it should not be allowed. In particular, he should not be allowed to open a thread without being allowed to view the item... for closing, one may argue you could be allowed to close without being able to view the content... but yes that's safer to just disallow it.

zenovich:
> > @smadbe, threadUpdate doesn't require the current user to have can_view>=content permission on the thread's item. As the result, a thread's participant can update the thread even if he is not allowed to view content of the thread's item. I'm not sure, but it looks like a bug. Should we fix it?
> 
> Yes probably it should not be allowed. In particular, he should not be allowed to open a thread without being allowed to view the item... for closing, one may argue you could be allowed to close without being able to view the content... but yes that's safer to just disallow it.

This will be done in a separate PR.

Issue #1219, opened 2024-11-18
Golang lock in accessTokenCreate (when we refresh a token) doesn't work


zenovich:
In accessTokenCreate, when we refresh a token of a non-temporary users (refreshAccessToken), we use locking implemented in Golang (sessionIDsInProgress.WithLock()) which is just a sync.Map storing Golang channels.
When we send a request to the login module to refresh a token,  we acquire a lock with sessionIDsInProgress.WithLock() to prevent  concurrent requests from refreshing the same token. This is important to do because the login module generates not only a new access token, but also a new refresh token and revokes the old one. So, we should not allow race conditions here, otherwise we can store a revoked refresh token into our DB making the user session broken.

The idea behind implementing this with sync.Map and Golang channels in 2019 was based on the suggestion that our server app will run as a single-instance application handling all the request. I wasn't aware that we will run multiple concurrent instances of AlgoreaBackend at the same time, one instance per user request, like we currently do on Amazon. As each instance has its own sessionIDsInProgress map, the locking doesn't work at all with the current method of use of AlgoreaBackend.

As we've implemented a reliable MySQL named locking that can be used outside of transactions (as part of #1215), we can use it for preventing concurrent token refreshing.

PR #1242, opened 2025-01-28
Return 404 and delete the session in refreshAccessToken when the login module does not recognize the refresh token

### LINKED ISSUE #1208: Handle revocated refresh token
## Motivations
Currently, if the refresh token is revocated, we get a 5xx error on authentication. It is not really "unexpected", we should handle it properly.

## Subtasks
 Handle that properly : the only solution may be just to inform the frontend that it has to fully reauthenticate, so for instance with a 403.

zenovich:
Also, some cucumber test steps are renamed here.

Fixes #1208


app/api/auth/refresh_access_token.go
```diff
@@ -106,6 +106,18 @@ func (srv *Service) refreshTokens(
 	oldToken := &oauth2.Token{RefreshToken: refreshToken}
 	oauthConfig := auth.GetOAuthConfig(srv.AuthConfig)
 	token, err := oauthConfig.TokenSource(ctx, oldToken).Token()
+
+	var retrieveError *oauth2.RetrieveError
+	if errors.As(err, &retrieveError) &&
+		retrieveError.Response.StatusCode == http.StatusUnauthorized {
+		// The refresh token is invalid
+		logging.SharedLogger.WithContext(ctx).
+			Warnf("The refresh token is invalid for user %d", user.GroupID)
```
smadbe, line 115: Actually, is it expected or not? Can you check (in login-module if needed) if the refresh token has an expiration?
Several options
* it has an expiration and we receive the info: we should check this value before trying the use it
* it has an expiration and we don't know it: maybe it should be just a "info" log... there is nothing to "warn"
* it is not supposed to expire: we should investigate why that happens (with Michel) first

zenovich: There are several possible errors returned by the login module on refreshing a token (all with 401 error code, but with different hints):
- Cannot decrypt the refresh token
- Token is not linked to client
- Token has expired
- Token has been revoked

In #1208 the hint is "Token has been revoked".

So we could handle different situations in different ways if we want.

zenovich: The actual ttl of refresh tokens in the login module is 1 year.

smadbe: I checked again at an error we got this morning, it was again "Token has been revoked". Could you check in the php code what it corresponds to? A realistic reason I see for this to happen so often would be that the user disconnects from the authenticator in another tab; from the authenticator directly or logging out from another app using the same authenticator. Could you check if that's possible ? 
I am pretty sure it is a "normal" case here but we have to check. If it is a valid case, the message should become an "info" and not a warning.

zenovich: I think the cause is that the user was sending concurrent requests to the login module for new access tokens with the same refresh token. After handling the first request, the login module marked the refresh token as revoked and generated a new one (it always returns a pair of a new access token and a new refresh token). Then it tried to handle the second request, but discovered that the given refresh token has been revoked.
So, I think, we should consider this situation as weird (a user tries to use a revoked refresh token): return 404 and log a warning.
The main problem here is, as such concurrent requests for new access tokens can lead to a situation where we save a revoked refresh token, we should not allow such requests to be concurrent. The idea was to use a lock (see https://github.com/France-ioi/AlgoreaBackend/blob/master/app/api/auth/refresh_access_token.go#L68). The issue is that the current implementation of this lock doesn't work (see https://github.com/France-ioi/AlgoreaBackend/issues/1219) which can be a real cause of #1208)

smadbe: Thanks for the analysis :+1:

zenovich: I have reproduced the issue locally:
```
curl -X POST http://127.0.0.1:8000/oauth/token -d 'client_id=9&client_secret=...&grant_type=refresh_token&refresh_token=def5020086270f4d2c982636058dd8010ea6665e007c0ded38b4e6e4cf05777182aa036944bc782fbecc052cd0839330b2a6043a6c5f5db50766a999092274880ed6bce4e39f03d85b945f32f6a5c0e3c24bba266b4fc3966544220886f26c34c09664b64ba69e20abd2a64c2dd896e76c558fbb6139fda7ed6badd1f4980f9f0da72efddf085845a0b9d7f65ff8b8f6c04a2c75b3d529d091e104ce7ca2291248660e9c7398fb1faacd52ac308a36d38ae83b35c6986c3f27199bd6c034e994fd544ebf072d0df34f9162ef7b35462839a9861ac7a0f4b5734a918ce6835380c908a191261aab0ee3917b074a8b99911354ee134180f1420b5d3fb18e5e3c7d289d6f7252d7f2e0fe68b72681e56f9a63beabc5f65121bc1e439b91d473e8996371dcc02ff4f8f8fce67a54e5f58131b60f91413909676f9a10ed3422c35d7dda9e7d2d6719e0ead3b2c2e492fb665e036f1076f713f675044091dcf9afcb0806e9792867c43b2bd0e4e0ab67af87f667fe'
{"token_type":"Bearer","expires_in":31536000,"access_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImp0aSI6ImM3NjMxODQ3ZGEwYjdjYThiNTI1MTI1Nzk3YzlhYTBjMGEwZTU4NjI5Y2Q1YTVmMTA4MWExYzNlYmNjNTRiM2NlMjRkYjhmNWQwYTJhNTc3In0.eyJhdWQiOiI5IiwianRpIjoiYzc2MzE4NDdkYTBiN2NhOGI1MjUxMjU3OTdjOWFhMGMwYTBlNTg2MjljZDVhNWYxMDgxYTFjM2ViY2M1NGIzY2UyNGRiOGY1ZDBhMmE1NzciLCJpYXQiOjE3NTA5Mjk1NjYsIm5iZiI6MTc1MDkyOTU2NiwiZXhwIjoxNzgyNDY1NTY2LCJzdWIiOiIxMDAwMDAwMDQiLCJzY29wZXMiOlsiYWNjb3VudCJdfQ.JGORkA3X2GXxwG1uJzcxj6UKL_ko1F53qsXUpNRbg9eE9RI0oGsy8Cho9rnSZEjFC_AXFycS2-xxE77g4fu0gh5ogfn5S370QWgao6trm_vRP01_kP_v3b_bnbIS33tk1Qp_j_07_WyQ5ROhpmOTxmbKHgl6wmtywxjRixhle3q9zbv7nDOpAui5kNlWNHwuxVPb3qhj7MN0gxFvCHQ3j0N8EVVcrApudUVXiZVHIaGgKsqp6gJlKF75Yl8YcQUvt9LVheuvfLqY-x3lVIadKeJTuWEE7y1Jab4VFa456KdJkLdaNBra2dFTDiTcnFXoDBM-_2C1IjJYRfj_sOCzdGymmFlBtnGotw8exR2wAPv0mQ5dE2I85bYajxkDPx9yT84t12Y7aNEOHAWOumOTgQLrQGEW44P-x7uVI88eBrGwgETJVyplU9y_SfxDk2e9SsM6P9QCC1Mt0oXmeZdTbqwzP-yseRQx3n4AMLFkXYaKz3SZRX0yxGSV-MpdyX499gmTPluxMmY9SbKrObEQRF7i-5o1s2boaX3kR3Fpmb1DpJuMY2T_9S6oJvkKfdcrLdtegJAKjpO_i6QBF6d0fYtaw7CG_eang6PbSPmYZrXa0i-wmeM-8L5waImzX91n15MGr6FYjTvsegqHKlIEQr1H0vXkqBEUNEXV0YSdm5g","refresh_token":"def50200b56ab3133ae23fb41bdf4299e6166950a523cf8c43b2dfd21c32e1734846fab138b8a4e510310cc3f9a1cb92290fc0a9ade88b2f9da47234a91b2e55f8bf9de22d42b8361cecf0d119a23d8860fe0748170c80748d6b3f829eba982baad3faf595764f13221e164d0c485a6464a71f029338690e71884dad6ea28cd66d5125c83f7c240b96c692e3974d4361c7109198bc449447537a8707688c0ed78e5a7205e4bcb2754590be563a8107400d7d8be626ab230a094044df1754cc9b62838ba33805e6fc825be1244e89190855d01e96877d95b982cce6d9f80531f3c9427eeee0bd99ef9f641f7c859aea546cb52d9492025c228171c682240c18141d4c3abb739e4c3fda8db8294ee1b12031a543971bb5951b50476af27a8a6c2b9cbfa2d5b5cfdc6d23e3d84ac14a09b714e5559def32bed3d65938b60172ff6bb30678f67bcca26862021e42274369422dffacfed6984c9c4922418a4909f0192a3478f7bce08efcac66740c8f221e67dc6e"}

% curl -X POST http://127.0.0.1:8000/oauth/token -d 'client_id=9&client_secret=...&grant_type=refresh_token&refresh_token=def5020086270f4d2c982636058dd8010ea6665e007c0ded38b4e6e4cf05777182aa036944bc782fbecc052cd0839330b2a6043a6c5f5db50766a999092274880ed6bce4e39f03d85b945f32f6a5c0e3c24bba266b4fc3966544220886f26c34c09664b64ba69e20abd2a64c2dd896e76c558fbb6139fda7ed6badd1f4980f9f0da72efddf085845a0b9d7f65ff8b8f6c04a2c75b3d529d091e104ce7ca2291248660e9c7398fb1faacd52ac308a36d38ae83b35c6986c3f27199bd6c034e994fd544ebf072d0df34f9162ef7b35462839a9861ac7a0f4b5734a918ce6835380c908a191261aab0ee3917b074a8b99911354ee134180f1420b5d3fb18e5e3c7d289d6f7252d7f2e0fe68b72681e56f9a63beabc5f65121bc1e439b91d473e8996371dcc02ff4f8f8fce67a54e5f58131b60f91413909676f9a10ed3422c35d7dda9e7d2d6719e0ead3b2c2e492fb665e036f1076f713f675044091dcf9afcb0806e9792867c43b2bd0e4e0ab67af87f667fe'
{"error":"invalid_request","message":"The refresh token is invalid.","hint":"Token has been revoked"}
```

smadbe: I didn't know the refresh token was changed each time it was used.

> So, I think, we should consider this situation as weird (a user tries to use a revoked refresh token): return 404 and log a warning.

You mean on the login module? On AlgoreaBackend, the user does not know anything about the refresh token... Anyway we need to allow the user (the frontend) to request an access token refresh in parallel, and that should not create a 404, nor a warning, nor an invalid token.

zenovich: > You mean on the login module?
Yes

> Anyway we need to allow the user (the frontend) to request an access token refresh in parallel, and that should not create a 404, nor a warning, nor an invalid token.
This means we should fix #1219 which is about this issue.

But still, we need to merge this task in order to remove stale refresh tokens from the DB, return 404 instead of 500, and log the weird situation as a warning.

PR #1246, opened 2025-02-10
Acquire shared row locks instead of exclusive row locks in DB queries causing request timeouts in production

zenovich:
Acquire shared row locks instead of exclusive row locks in startResult(), ItemStore.CheckSubmissionRights(), ResultStore.GetHintsInfoForActiveAttempt()

### DIFF
--- a/app/database/result_store.go
+++ b/app/database/result_store.go
@@ -29,6 +29,7 @@ func (s *ResultStore) GetHintsInfoForActiveAttempt(participantID, attemptID, ite
 	mustNotBeError(s.Results().
 		ByID(participantID, attemptID, itemID).
 		WithCustomWriteLocks(golang.NewSet("attempts"), golang.NewSet("results")).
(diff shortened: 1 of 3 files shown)


app/database/result_store.go
```diff
@@ -28,7 +28,7 @@ func (s *ResultStore) GetHintsInfoForActiveAttempt(participantID, attemptID, ite
 	var hintsInfo HintsInfo
 	mustNotBeError(s.Results().
 		ByID(participantID, attemptID, itemID).
-		WithExclusiveWriteLock().Select("hints_requested, hints_cached").
```
smadbe, line 31: So the `.Select("hints_requested, hints_cached")` has disappeared?

smadbe (approved):

app/database/result_store.go
zenovich: Wow! Nice catch! Fixed.

PR #1247, opened 2025-02-11
Do not analyze tables or recompute db caches in db-migrate command + always close the DB connection in commands

### LINKED ISSUE #1184: db-migrate: just migrate
Db-migrate currently migrate then analyze then propagate. The migration is only taking 5% of the time and failure. (Anyway propagation is running all the time in prod.)  So we loose control on the migration process which is quite critical. 

So just do db migration on db-migrate... and move the rest to other commands if needed.

zenovich:
1. Do not analyze tables or recompute db caches in db-migrate command
2. Close the db connection in commands even when errors occur

Fixes #1184 

Note: We need to guarantee somehow that there are no concurrent requests/propagations running while the db-migrate command is being executed. Otherwise, any unpredictable data corruptions are possible as old code (code from a previous release) is not compatible with the DB after migrations are applied and the new code (code from a release to be deployed) is not compatible with the DB before migrations are applied. So, neither the old code, nor the new code can be running concurrently while we are executing DB migrations.

### DIFF
--- a/cmd/db_gen_load.go
+++ b/cmd/db_gen_load.go
@@ -54,6 +54,7 @@ func init() { //nolint:gochecknoinits,gocyclo,gocognit
 			tablesToTruncate := []string{"groups", "groups_groups", "groups_ancestors", "groups_attempts"}
 			for _, tableName := range tablesToTruncate {
 				if _, err := rawdb.Exec(fmt.Sprintf("TRUNCATE %s", tableName)); err != nil {
+					_ = rawdb.Close()
 					panic(err)
 				}
 			}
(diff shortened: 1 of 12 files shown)


smadbe:
> Note: We need to guarantee somehow that there are no concurrent requests/propagations running while the db-migrate command is being executed. Otherwise, any unpredictable data corruptions are possible as old code (code from a previous release) is not compatible with the DB after migrations are applied and the new code (code from a release to be deployed) is not compatible with the DB before migrations are applied. So, neither the old code, nor the new code can be running concurrently while we are executing DB migrations.

I cannot guarantee that and I don't see how I could guarantee that... The only way I see would be that you use the propagation lock in the migration.

smadbe (approved):

PR #1251, opened 2025-02-13
Remove special characters from search strings in database.DB.WhereSearchStringMatches()

### LINKED ISSUE #1209: Escape item search
## Motivations
We may get 5xx on the search service. 

For instance:
`unexpected error: Error 1064: syntax error, unexpected $end, expecting FTS_TERM or FTS_NUMB or '*', stack trace: goroutine 1 [running]:...`  on `/items/search?limit=11&search=Motif%2520art%2520-&types_include=Chapter%2CTask`

## Subtasks
* Handle escaping properly.

zenovich:
Preprocess search strings in database.DB.WhereSearchStringMatches() keeping only letters/digits/apostrophes.

Fixes #1209

Note: Unfortunately, there is no way to escape special characters acting like operators of boolean full-text search in MySQL inside search words and use such words outside of the exact phrase operator (double quotes) at the same time. So, we should filter out all the operator characters [+-()~<>*"@]. At the same time, for security reason, here we remove all the characters except for alpha-numeric characters (for all languages of the world) and apostrophes.

### DIFF
--- a/app/database/search.go
+++ b/app/database/search.go
@@ -1,6 +1,9 @@
 package database
 
-import "strings"
+import (
+	"strings"
+	"unicode"
+)
 
 // WhereSearchStringMatches returns a composable query where {field} matches the search string {searchString}
 // All the words in the search string are matched with "AND".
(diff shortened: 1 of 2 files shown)


smadbe (approved):
No feature test that the type of string which was triggering 5xx is ok now?

PR #1276, opened 2025-05-10
Introduce a new field is_direct_manager is output of groupManagersView in the case when include_managers_of_ancestor_groups=1

zenovich:
The value of the field is true for output rows corresponding to direct (explicit) managers of the `group_id` group.


app/api/groups/get_managers.go
```diff
@@ -27,6 +27,8 @@ type GroupManagersViewResponseRowThroughAncestorGroups struct {
 	CanGrantGroupAccessThroughAncestorGroups bool `json:"can_grant_group_access_through_ancestor_groups"`
 	// displayed only when include_managers_of_ancestor_groups=1, note that the group is an ancestor of itself
 	CanWatchMembersThroughAncestorGroups bool `json:"can_watch_members_through_ancestor_groups"`
+	// displayed only when include_managers_of_ancestor_groups=1
```
smadbe, line 30: May be worth a more precise explanation so that we do not wonder what "direct" means exactly later.
I would have put the field in every case (always true if `include_managers_of_ancestor_groups`=0) but that's ok like that as well.

smadbe (changes requested):
The problem still remain on the meaning of the `can_manage` field... Now we now that if `is_direct_manager` is `false`, we know that  `can_manage` value is wrong... but does it make sense to include a wrong value in the response?

app/api/groups/get_managers.go
zenovich: reworked everything

zenovich: removed the field is_direct_manager

smadbe (approved):

PR #1277, opened 2025-05-11
Rework row locking in groupManagerDelete

zenovich:
1. Add locking for the case where managerID == user.GroupID
2. Use more granular locking in the case where managerID != user.GroupID

### DIFF
--- a/app/api/groups/remove_manager.go
+++ b/app/api/groups/remove_manager.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/France-ioi/AlgoreaBackend/v2/app/database"
 	"github.com/France-ioi/AlgoreaBackend/v2/app/service"
+	"github.com/France-ioi/AlgoreaBackend/v2/golang"
 )
 
 // swagger:operation DELETE /groups/{group_id}/managers/{manager_id} groups groupManagerDelete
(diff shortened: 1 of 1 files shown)


app/api/groups/remove_manager.go
```diff
@@ -66,11 +67,14 @@ func (srv *Service) removeGroupManager(w http.ResponseWriter, r *http.Request) s
 		//    or have can_manage:memberships_and_group permission on the groupID
 		// 2) there should be a row in group_managers for the given groupID-managerID pair
 		if managerID == user.GroupID {
-			found, err = store.GroupManagers().
+			found, err = store.GroupManagers().WithExclusiveWriteLock().
```
smadbe, line 70: What's the risk without this lock?

smadbe (approved):

app/api/groups/remove_manager.go
zenovich: Without this lock, the query would not see the rows committed concurrently after the start of the transaction.

zenovich: It was weird to have locking only for one of two cases

smadbe: I think I don't see more reasons for the other lock 😄

PR #1299, opened 2025-06-23
Make the resultStart service less locking

zenovich:
We don't need to lock the attempts table when we retrieve the attempts list, also, we can use a shared lock on the results table instead of an exclusive lock.

### DIFF
--- a/app/api/items/start_result.go
+++ b/app/api/items/start_result.go
@@ -135,7 +135,7 @@ func (srv *Service) startResult(w http.ResponseWriter, r *http.Request) service.
 
 		service.MustNotBeError(constructQueryForGettingAttemptsList(store, participantID, itemID, srv.GetUser(r)).
 			Where("attempts.id = ?", attemptID).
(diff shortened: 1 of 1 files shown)


app/api/items/start_result.go
```diff
@@ -135,7 +135,7 @@ func (srv *Service) startResult(w http.ResponseWriter, r *http.Request) service.
 
 		service.MustNotBeError(constructQueryForGettingAttemptsList(store, participantID, itemID, srv.GetUser(r)).
 			Where("attempts.id = ?", attemptID).
-			WithCustomWriteLocks(golang.NewSet("attempts"), golang.NewSet("results")).
+			WithCustomWriteLocks(golang.NewSet("results"), golang.NewSet[string]()).
```
smadbe, line 138: >  also, we can use a shared lock on the results table instead of an exclusive lock.

What makes it a shared lock ?

zenovich:
```
func (conn *DB) WithCustomWriteLocks(shared, exclusive *golang.Set[string]) *DB {
```

PR #1302, opened 2025-07-03
Optimize permissions marking in after_update_items_items even more

zenovich:
Instead of marking generated permissions related to the child item as 'propagate_to=self', we will, for each generated permissions related to the parent item, mark the (group_id, child item's id) pair as 'propagate_to=self' (this will work better if we decide to remove permissions_generated having none/zero in all the permission-related columns).


smadbe (approved):
:+1:
That makes sense.

PR #1303, opened 2025-07-04
Optimize permissions marking in before_delete_items_items even more

zenovich:
Instead of marking generated permissions related to the child item as 'propagate_to=self', for each generated permissions related to the parent item, mark the (group_id, child item's id) pair as 'propagate_to=self' (we only want to recompute those generated permissions of the child item that have been propagated from the parent item, we do not want to recompute all the generated permission of the child item)

### DIFF
--- a/app/api/items/update_item.feature
+++ b/app/api/items/update_item.feature
@@ -178,7 +178,9 @@ Background:
       | 5577006791947779410 | 50      | content  | none           | none      | none     | false    | 5577006791947779410 | 1                                                       |
     And the table "permissions_generated" should be:
       | group_id            | item_id | can_view_generated | can_grant_view_generated | can_watch_generated | can_edit_generated | is_owner_generated |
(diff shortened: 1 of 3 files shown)


smadbe (approved):
👍 
If I understand it well, if the parent has a lot of permissions but none of them are propagated to the child, we'll have more "computation" in this case than in the previous one.

zenovich:
> 👍 If I understand it well, if the parent has a lot of permissions but none of them are propagated to the child, we'll have more "computation" in this case than in the previous one.

Not exactly. If the parent has a lot of permissions, but none of them are propagated to the child, we'll still recompute them, but it's not more "computation" than when we mark permissions of the child item:
1. Currently, when none of permissions are propagated we still create permissions_generated with none/zeros for the child item (and we never remove empty generated permissions). So when we mark permissions of the child item, it's, at least, the same number of rows.
2. As there can be permissions for other groups propagated to the same child item from other parent items, there are often more generated permission for a child item than for a parent item.

PR #1320, opened 2025-08-25
Partially synchronous propagation in itemCreate

### LINKED ISSUE #1311: Make "itemUpdate" partially async
## Motivations

The chapter edition is the main place where the propagation is still "painful". In practice, quite often (because typically the chapter with big impacts are those which are modified more often), the save is quite long and ends up with a timeout... if the user refresh a few seconds later, that's not guaranteed that the propagation has completed and so that is very likely that the user does not even see the title of an item he has just create and put in the chapter.

Note that is might also apply to itemDelete and itemCreate as well (to be checked).

Here is an extract from #1181:
> These services affect the results of many (possibly all) users on possibly a high item hierarchy depth. This may cause unlocking which may trigger permission propagation as well. These services are the main (only?) source of problems. So they need to run their propagation async.
> 
> But but... these are also affecting the current user:
> 
> * When creating an item, the user gets owner permission on the item (which require propagation to be effective) and typically the first thing the user wants to do is to edit the item,  but currently, it may not be able to see the title of the element he has just created because of propagation hasn't run yet..  _Problem we will need to fix very soon_
> 
> * The score of the parent chapter may be immediately impacted, probably it is clearer for the user if that score is updated immediately in sync. (Lower priority)
> 
> 
> => **so probably the propagation impacting the current user have to be applied in sync, the other asynchronously.**

### A bit more details 

In practice, the way the frontend works is: when adding new (newly created) item, when "save" is pressed, items are created first (giving the parent id!) and then the update is called. 

I don't clearly understand why the item appears as "non visible item" (id is returned but the string is unknown as the item is invisible (as [defined in the API](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v2.28.3.html#tag/items/operation/itemChildrenView)) at first (while the propagation is not over).

## Subtasks
When changing item children:
* the propagation should be applied immediately for the current user. In practice:
  * the permission propagation to children should be applied asap (in sync) for the current user
  * the score of the current user on the modified chapte should be updated asap (in sync)
* the propagation should be async for the other users

## Subtasks (optional)
Why, after adding a new item to a parent, the name of the newly created item may not appear. As the current user is owner of the newly created item and was able to see the parent, we are guaranteed that the user can view the newly created item, right?  The only explanation I find is that this "propagation" is asynced. But as it is guaranteed that is visible, shouldn't we force the "can_view" permission on the current user already (doing the UPSERT on `permissions_generated` (edited) without running any propagation yet) so that it is guaranteed that any subsequent GET call of the current user of the item already return the "strings" of the item ?

zenovich:
This PR modifies the itemCreate service propagate/recompute some things synchronously:
1. Synchronously propagate newly created permissions of the current user on the new item,
2. Synchronously recompute and propagate results of the current user on the parent item (if specified),
3. Synchronously propagate results of the current user on descendants of the new item (if any).
(As discussed in https://github.com/France-ioi/AlgoreaBackend/issues/1311#issuecomment-3201987399, https://github.com/France-ioi/AlgoreaBackend/issues/1311#issuecomment-3202032726, https://github.com/France-ioi/AlgoreaBackend/issues/1311#issuecomment-3202106754)

Fixes #1311 

Also, some things related to cucumber tests were reworked/improved/added here. The most remarkable thing is sorting of TestContext.dbTableData tables which should allow to compose new cucumber tests easier.

### DIFF
--- a/app/api/currentuser/get_group_invitations.feature
+++ b/app/api/currentuser/get_group_invitations.feature
@@ -19,9 +19,9 @@ Feature: Get group invitations for the current user
       | 36 | Class   | Group without inviting user   | Group without inviting user   | edit                                  | null                                   | false                  |
     And the database has the following users:
       | group_id | login       | first_name  | last_name | grade |
-      | 21       | owner       | Jean-Michel | Blanquer  | 3     |
       | 12       | user        | John        | Doe       | 1     |
       | 13       | anotheruser | Another     | User      | 1     |
+      | 21       | owner       | Jean-Michel | Blanquer  | 3     |
     And the database has the following table "groups_groups":
       | parent_group_id | child_group_id |
       | 5               | 21             |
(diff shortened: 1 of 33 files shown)


smadbe (approved):
👍

Issue #1356, opened 2025-11-05
Get log for item: extend visibility to thread viewer


smadbe:
## Motivations

(same as [we did it for answers](https://github.com/France-ioi/AlgoreaBackend/issues/1042) )

Adapt the "activity log on item" service (`/items/{ancestor_item_id}/log`) so that any user who can view a forum thread of a (participant,item) can also get the activity of this participant on this item.

## Subtasks

* Make sure a user who sees a thread, can view the activity it corresponds to, so in practice apply the thread reading permissions permission [from here](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/#listing--reading-a-thread)

## Update

As discussed on Slack: better to do it is its specific endpoint.

This new service has 2 input:
- `item_id`: we will NOT look for the activities on the descendants of that content
- `participant_id`: we will NOT look for the activities by the participant's descendants

Basically the call on that service is allowed if the current user matches the [thread reading permissions](https://france-ioi.github.io/algorea-devdoc/forum/forum-perm/#listing--reading-a-thread)

PR #1357, opened 2025-11-10
Item activity log for threads

zenovich:
1. Introduce a new service itemActivityLogForThread, add tests
2. Make sorting in services related to activity log stable.
3. Fix tiebreaking in pagination in situations where all the fields except for the activity type have the same values.

Implements #1356


smadbe (approved):
👍

